Issue #1, opened 2012-09-14
Install as app whilst b2g running


andymckay:
If you do this, nothing happens. Would be nice to have a "go restart b2g" message.

mykmelez:
Indeed! At the very least, we should let you know to restart B2G. (Ideally, we should restart it ourselves or figure out how to install an app without requiring a restart.)

mykmelez:
As of commit 6eb2f6afb484d678711fa5ea98188062816fbeb3, the addon restarts B2G Desktop automagically!

PR #4, opened 2012-09-19
Adding some features to the launch and install process

digitarald:
- Launch installed apps
- Launch with JS console by default (can be disabled with setting).
- Kill previous process when launching new B2G desktop

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -101,7 +101,7 @@ function run(app) {
     },
 
     done: function(result) {
-      dump(executable + " terminated with " + result.exitCode);
+      console.log(executables[Runtime.OS] + " terminated with " + result.exitCode);
       currentProcess = null;
     }
 
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -81,7 +81,7 @@ function run(app) {
   }
 
   if (app != null) {
-    args.push("-runapp", app);
+    args.push("--runapp", app);
   }
 
   if (currentProcess != null) {
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -177,7 +177,7 @@ function installActiveTab() {
   File.open(webappsFile, "w").writeAsync(JSON.stringify(webapps, null, 2) + "\n",
     function(error) {
       console.log(JSON.stringify(webapps[key], null, 2));
-      run(name.match(/[\w\s]+/)[0]); // Clean up name for CLI
+      run(name);
     }
   );
 }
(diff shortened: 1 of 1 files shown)


addon/lib/main.js
```diff
@@ -85,6 +100,11 @@ function run() {
       dump(data);
     },
 
+    done: function(result) {
+      dump(executable + " terminated with " + result.exitCode);
```
mykmelez: `dump` doesn't append a newline, so you need to append one yourself (the `stdout` and `stderr` functions are exceptional cases, as they dump `data` values that embed their own newlines).

However, the good news is that you don't actually need to use `dump` in this case, as the Add-on SDK platform (with which this addon is built) has a built-in `console` object that does append a newline. So this can be:

```
console.log(executable + " terminated with " + result.exitCode);
```

However, `executable` is an _nsIFile_ that stringifies to `[xpconnect wrapped nsIFile]`; to print the actual executable path, use `executable.path`, or for just the filename use `executable.leafName`, per [nsIFile](http://mxr.mozilla.org/mozilla-central/source/xpcom/io/nsIFile.idl).

addon/lib/main.js
```diff
@@ -73,7 +76,19 @@ function run() {
   let profile = URL.toFilename(Self.data.url("profile"));
   let args = ["-profile", profile];
 
-  subprocess.call({
+  if (Prefs.get("extensions.r2d2b2g.jsconsole", true)) {
+    args.push("-jsconsole");
+  }
+
+  if (app != null) {
+    args.push("-runapp", app);
```
mykmelez: `-runapp` works, but [Using the B2G desktop client](https://developer.mozilla.org/en-US/docs/Mozilla/Boot_to_Gecko/Using_the_B2G_desktop_client) documents `--runapp` (two hyphens), which also works and seems like the Unix convention these days for multi-character options, so I would use that instead.

addon/lib/main.js
```diff
@@ -155,6 +177,7 @@ function installActiveTab() {
   File.open(webappsFile, "w").writeAsync(JSON.stringify(webapps, null, 2) + "\n",
     function(error) {
       console.log(JSON.stringify(webapps[key], null, 2));
+      run(name.match(/[\w\s]+/)[0]); // Clean up name for CLI
```
mykmelez: It isn't clear what this cleanup is intended to do; can you elaborate?

digitarald: I thought it fixed some problems during launch that I had, but after testing I couldn't actually reproduce them. I'll remove it.

mykmelez:
This addresses issue #1.

PR #8, opened 2012-09-26
Install app via webapp manifest

digitarald:
Currently only via rightclicking on *.webapp links.

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -159,6 +159,22 @@ Menuitems.Menuitem({
  */
 function installActiveTab() {
   let url = URL.URL(Tabs.activeTab.url);
+  let origin = url.toString().substring(0, url.lastIndexOf(url.path));
+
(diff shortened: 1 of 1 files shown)


addon/lib/main.js
```diff
@@ -159,6 +162,61 @@ Menuitems.Menuitem({
...
+        console.warn("Unexpected Content-Type " + contentType + ", but not a biggie");
+      }
+
+      installManifest(manifestUrl, response.json);
+    }
+  }).get();
+}
+
+function installManifest(manifestUrl, webapp, installOrigin) {
+  let origin = manifestUrl.toString().replace(/([^\/])\/[^\/].*/, "$1");
```
mykmelez, line 215: Elsewhere, we extract an origin from a URL via:

```
url.toString().substring(0, url.lastIndexOf(url.path));
```

Shouldn't we do the same here?

PR #43, opened 2012-10-30
Fixing tools menu

### LINKED ISSUE #32: Firefox OS Simulator item in Tools menu should be under Web Developer menu
The Firefox OS Simulator item in the Tools menu currently appears underneath the Page Info item. Per the [v1 spec](https://docs.google.com/document/d/1S538182p1zwCAFHvIRTYCu0LkdHLVh3KZCVhOdKbh1Q/edit), it should appear just below the Web Developer menu.

digitarald:
- Moving menu item from tools in web developer (fixes #32)
- Removing Install Page menu

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -558,7 +558,7 @@ ContextMenu.Item({
 });
 
 Menuitems.Menuitem({
-  id: "webdevB2GSimulatorHelper",
+  id: "webdevFxOSSimulatorHelper",
   menuid: "menuWebDeveloperPopup",
   insertbefore: "devToolsEndSeparator",
   label: "Firefox OS Simulator",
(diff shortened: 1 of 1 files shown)


addon/lib/main.js
```diff
@@ -558,9 +558,19 @@ ContextMenu.Item({
...
+  insertbefore: "devToolsEndSeparator",
+  label: "Firefox OS Simulator",
+  onCommand: function() {
+    openHelperTab();
+  },
+});
+
+Menuitems.Menuitem({
+  id: "appmenu_webDeveloper_popup",
+  menuid: "appmenu_webDeveloper_popup",
```
mykmelez, line 572: `id` and `menuid` should have different values. It looks like this `menuid` is correct, while `id` should be something like `appMenuSimulatorItem`.

mykmelez, line 572: (Or, if we want to be consistent with the appmenu code in Firefox: `appmenu_simulator`.)

PR #48, opened 2012-10-31
Added "JS console" and "Run by default" preferences

digitarald:
Two preferences and some clean up

### DIFF
--- a/addon/data/content/css/main.css
+++ b/addon/data/content/css/main.css
@@ -413,6 +413,17 @@ figure {
     text-decoration: none;
 }
 
+.commands-options {
+    display: block;
+    margin-top: 10px;
+    line-height: 1.4em;
+    font-size: 0.8em;
+    color: #888;
+}
+.commands-options input {
+    padding-right: 5px;
+}
+
 .toggle {
     position: relative;
     border: 1px solid rgba(0, 0, 0, 0.4);
@@ -475,16 +486,13 @@ figure {
     opacity: 0.4;
     cursor: progress;
 }
-.list-actions ul {
+.list-actions {
     margin: 0;
-    padding: 0;
-}
-.list-actions ul {
+    margin-bottom: 10px;
+    padding: 10px 0 9px;
     list-style: none;
     display: block;
     border-bottom: 1px solid #eee;
-    padding: 10px 0 9px;
-    margin-bottom: 10px;
     text-align: right;
 }
 .list-actions li {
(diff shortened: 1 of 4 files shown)


mykmelez:
When I try to open the simulator with the patch applied, I get the following exception:

```
    File "resource://r2d2b2g-at-mozilla-dot-org/menuitems/lib/menuitems.js", line 95, in 
      emit(self, 'command');
    File "resource://r2d2b2g-at-mozilla-dot-org/api-utils/lib/event/core.js", line 83, in emit
    File "resource://r2d2b2g-at-mozilla-dot-org/api-utils/lib/event/core.js", line 101, in lazy
      yield listeners.shift().apply(target, args);
    File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/main.js", line 640, in 
      simulator.openHelperTab();
    File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/main.js", line 397, in 
      if (tab.url.startsWith(url)) {
  TypeError: tab.url.startsWith is not a function
```

That's probably because I'm using the simulator in Firefox 16, which doesn't support `String.startsWith()`. We should support the stable release of Firefox and thus work around this problem by using `tab.url.indexOf(url) == 0`.

Otherwise, I haven't fully reviewed this yet, but we have a product manager who is actively engaged in specifying the feature set of this product, so let's make sure we have him review any potential new features before we implement them.  He's already specified the console pref, but cc:ing him (@dangoor) here for his thoughts on the additional "run by default" pref.

(In general, I think we should avoid preferences like the plague, as each presents the user with a choice that slows him down and makes him think; so I would instead automatically run the most recent app on startup or run an app automatically when you press its Update button; but Kevin is the ultimate authority here.)

digitarald:
Thanks for the detailed feedback, Myk!

I hoped to get feedback on the pull request, as the feature itself was a minor code change.

My _Run by Default_ user story: It originated from my need to debug a single app. I also thought about _Run on update_; while great for packaged apps; for my hosted app case I didn't need to "Update" the app, restarting the simulator and launching the app directly worked as expected.

dangoor:
I agree with Myk that preferences are nearly always a bad thing and there is often a better way to get at what the user intends. I also like Myk's suggestion that we run the most recently run app by default on startup, though I wonder if there could be unintended side effects (what if that app is actually crashing the simulator? generally shouldn't happen, but I wouldn't want someone to be stuck if it did)

Repeatedly working on a single app seems like it would be the common case, which makes run by default convenient.

Sometimes (as with the error console) the user literally wants to do one thing sometimes and another thing other times, and there's no way for the software to guess. prefs and toggles are fine there.

Also, for hosted apps, Update can still have value if it makes it easy to clear the appcache without requiring a manifest change (and which wouldn't be part of this pull request :)

digitarald:
The `startsWith` fix slipped into the other, less controversial pull request #50.

PR #50, opened 2012-10-31
Feature: Reveal App

digitarald:
p.s. I am hoping that my branch merges don't confuse git, as they are listed in commits. Maybe they'll vanish when the other pull request got merged.

### DIFF
--- a/addon/data/content/css/main.css
+++ b/addon/data/content/css/main.css
@@ -413,6 +413,17 @@ figure {
     text-decoration: none;
 }
 
+.commands-options {
+    display: block;
+    margin-top: 10px;
+    line-height: 1.4em;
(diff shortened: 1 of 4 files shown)


mykmelez:
Looks good!

Per the v1 spec, I was expecting an `Open Location` button instead of a clickable filepath/URL link, but the clickable link will get folks testing the feature, and it's easy to move it to a button in a followup, so I'm going to merge this, and then we can iterate from there.

Since this incorporates the commits from [pull 48](https://github.com/mozilla/r2d2b2g/pull/48), I expect to get both sets of changes in the merge. It's less clear whether GitHub will close the other pull automatically. If not, I'll close it manually.

PR #57, opened 2012-11-01
make Windows work

mykmelez:
This makes Windows work. I haven't tested to see if it breaks other OSes, which I'll want to do before integrating it.


mykmelez:
Tested and confirmed that it works.

PR #68, opened 2012-11-06
Added "Remove" link for apps, removes app from storage and FxOS

### LINKED ISSUE #45: "Remove App" for app dashboard
Removes an app from B2G profile and the add-on storage.

digitarald:
Fixes #45

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -6,6 +6,9 @@ var Simulator = {
     "hosted": "Hosted App"
   },
 
+  CONFIRM_DELETE: "This will remove the app from Firefox OS Simulator " +
+    "and the Manager. Are you sure?",
+
   init: function() {
 
     this.toggler = $('#command-toggle')[0];
@@ -132,7 +135,7 @@ var Simulator = {
                       .text("Remove")
                       .click(function(evt) {
                         evt.preventDefault();
-                        if (confirm("This will remove the app from Firefox OS Simulator and the Manager. Are you sure?")) {
+                        if (confirm(Simulator.CONFIRM_DELETE)) {
                           window.postMessage({name: "removeApp", id: id}, "*");
                         }
                       }),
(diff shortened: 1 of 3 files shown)


addon/lib/main.js
```diff
@@ -217,6 +222,46 @@ let simulator = {
...
+    delete webapps[config.xkey];
+    File.open(webappsFile, "w").writeAsync(
+      JSON.stringify(webapps, null, 2) + "\n",
+      function(error) {
+        if (error) {
+          console.error("Error writing webapp record to registry: " + error);
+          return;
+        }
+
+        // Create target folder
```
mykmelez: Nit: this comment should describe _removing_ a folder.

addon/data/content/js/main.js
```diff
@@ -121,12 +127,21 @@ var Simulator = {
               container.append(
                 $("<div class='app'>").append(
                   $("<div class='options'>").append(
+                    $("<a href='#'>")
+                      .addClass("button")
+                      .text("Remove")
+                      .click(function(evt) {
+                        evt.preventDefault();
+                        if (confirm("This will remove the app from Firefox OS Simulator and the Manager. Are you sure?")) {
```
mykmelez: I suppose this is ok, although ideally we'd remove without prompting while giving the user a way to undo the action (as does Add-ons Manager when you remove an addon).

mykmelez: Nit: this line should be shortened to 80 characters or less in width.

digitarald: Agreed, the add-on Undo feature seems easier for the user. Storage would add a `deleted` flag and flush deleted addons from storage at unload. Should I try to get this in v1?

mykmelez: If it isn't hard to do, then yes! If it's going to take longer than the first half of today, though, then no. In that case, let's land this and do the additional work for the next cycle.

Issue #65, opened 2012-11-03
rename addon to Firefox OS Simulator


mykmelez:
Wherever "Ready to Deliver to Boot to Gecko" or "r2d2b2g" appears to users, we should change the string to "Firefox OS Simulator", or "simulator" for short. I'm thinking particularly of the addon's listing in Add-on Manager (for which we just need to update the addon's name/description in package.json).

One place we should _not_ change (right now) is the manager's URL. That is derived from the addon's ID, and it should be simpler (perhaps as simple as about:simulator), but doing that is probably complicated and doesn't matter for v1.

mykmelez:
Fixed by 971b12d3d962409e12837e5a5ce569a931a2afa5 and others.

PR #70, opened 2012-11-06
Removed welcome page r2d2b2g reference

digitarald:
As discussed in #71. Targets also #65.

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -238,7 +238,7 @@ let simulator = {
     let webappsFile = File.join(webappsDir, "webapps.json");
     let webapps = JSON.parse(File.read(webappsFile));
 
-    // Create the webapp record and write it to the registry.
+    // Delete the webapp record from the the registry.
     delete webapps[config.xkey];
     File.open(webappsFile, "w").writeAsync(
       JSON.stringify(webapps, null, 2) + "\n",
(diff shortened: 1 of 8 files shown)


addon/lib/main.js
```diff
@@ -411,7 +494,7 @@ let simulator = {
     });
   },
 
-  openHelperTab: function() {
+  openHelperTab: function(initial) {
```
mykmelez: It looks like this parameter is orphaned, as there's one call to `openHelperTab()` that passes an argument for it, but `openHelperTab()` never actually uses it. So we should remove it (and remove the argument from the call that passes it).

addon/lib/main.js
```diff
@@ -393,7 +476,7 @@ let simulator = {
     });
   },
 
-  openTab: function(url, onReady, lax) {
+  openTab: function(url, onReady, lax, append) {
```
mykmelez: This `append` parameter is similarly orphaned, now that we aren't showing the welcome page on firstrun. We should remove it too!

addon/lib/main.js
```diff
@@ -564,7 +659,7 @@ let simulator = {
 
 switch (Self.loadReason) {
   case "install":
-    simulator.openHelperTab();
+    simulator.openHelperTab(true);
```
mykmelez: You still have this one call passing an argument to a non-existent parameter!

PR #127, opened 2012-11-18
Ignoring dmg files for macs

erikvold:


mykmelez:
Hmm, we used to have this in the .gitignore file, but I removed it, as I figured it's the kind of rule that developers typically add to their global .gitignore file. The github:help article on [Ignoring Files](https://help.github.com/articles/ignoring-files) suggests a global .gitignore file that includes a rule for DMG files. But perhaps this is less typical than I think?

victorb:
Should exe files also be added? I don't think so. Platform exclusive files that normally doesn't relate to the project feels kind of unnecessary to add to the .gitignore.

digitarald:
Following git's best practices for now.

PR #136, opened 2012-11-19
Connection to debug with Firefox Remote debugger?

victorb:
I found one thing that's harder in the simulator than in other emulators of Firefox OS, debugging. There is one handy tool for it that exists in Nightly/Aurora where you can debug a remote device but it needs ip-address and port for connection to the machine.

I've added devtools.debugger.remote-enabled and devtools.debugger.remote-port to enable remote debugging on port 4242 for easier debugging but not entirely sure that it's enough.

### DIFF
--- a/prosthesis/defaults/preferences/prefs.js
+++ b/prosthesis/defaults/preferences/prefs.js
@@ -1,2 +1,4 @@
 user_pref("general.useragent.override", "Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0");
 user_pref("power.screen.timeout", 86400);
+user_pref("devtools.debugger.remote-enabled", true);
+user_pref("devtools.debugger.remote-port", 4242);


mykmelez:
I tried remote debugging on this branch, and it didn't work for me. So I looked around a bit and found [this thread](https://groups.google.com/forum/#!topic/mozilla.dev.b2g/UpDp2uTD4Mc) that suggests you also need `user_pref("marionette.defaultPrefs.enabled", false);`. Sure enough, adding that made remote debugging work!

The `devtools.debugger.remote-port` pref isn't necessary, though. And I think we're better off without it, as then B2G uses the default port of 6000, and that's the same port the remote debugger suggests you debug by default (specifically, it suggests the host:port combination localhost:6000, which is exactly the right combo for B2G Desktop).

So I'll merge your branch with those two changes. Thanks for the fix!

urbien:
Here is what worked for me after over a day of trying. 
I downloaded [b2g v25 nightly](http://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/latest-mozilla-central/b2g-25.0a1.multi.linux-x86_64.tar.bz2)
Then I changed devtools.debugger.remote-enabled to true in gaia/profile/settings.json 
Changing this in gaia/profile/prefs.js did not help, as at the start of the b2b this setting would get overwritten to false. I read here and elsewhere that I needed to change marionette setting, but it worked without it. Possibly the issue behind it got resolved by now.
Then I went to Tools-> Web Developer -> Connect
Tip: do not forget to answer OK in b2g window when debugger attempts to connect to it.
I do not know if my experience applies to the latest FF OS Simulator add-on though, as I have not found a way to run the nightly version of it (v25 at this point). Can anyone help with that?

Issue #145, opened 2012-11-21
Re-opening the Manager tab doesn't work


digitarald:
The add-on runner is not attached when the simulator manager tab is re-opened (from history for example).

mykmelez:
Fixed by the pull request.

PR #155, opened 2012-11-26
Kill process on last-tab-close and browser shutdown

### LINKED ISSUE #90: Kill Simulator when Firefox closes
Currently the process stays alive when the manager or browser closes, which causes problems.

We could also close the Simulator only when the browser closes, but attaching its life to the Manager tab seems to the better paradigm.

Thoughts? @dangoor @mykmelez

digitarald:
Asynchronous flow for all process.kill calls. Fixes #90 and #145.

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -34,7 +34,7 @@ let simulator = {
     if (this.worker) {
       this.worker.postMessage({
         name: "isRunning",
-        isRunning: !!this.process
+        isRunning: !!this.process,
       });
     }
   },
@@ -116,7 +116,7 @@ let simulator = {
         xkey: null,
         name: webapp.name,
         icon: icon,
-        manifest: webapp
+        manifest: webapp,
       }
       console.log("Stored " + JSON.stringify(apps[webappFile]));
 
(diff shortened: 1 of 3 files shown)


addon/lib/main.js
```diff
@@ -33,7 +34,7 @@ let simulator = {
     if (this.worker) {
       this.worker.postMessage({
         name: "isRunning",
-        isRunning: !!this.process,
+        isRunning: !!this.process
```
mykmelez: Nit: I generally prefer to include trailing commas in object literals these days, as they minimize code change when adding name/value pairs in subsequent changes.

mykmelez:
Overall, this looks great. I noticed two strange behaviors in testing, though:
1. If the Simulator is running when you quit Firefox, and there's a Manager tab, then starting Firefox up again causes the Simulator to launch automatically the moment you activate the Manager tab, which could be on startup (if the Manager tab was the active tab on shutdown). That's unexpected, and I imagine users would be surprised by it. Regardless of whether or not the Simulator was running when you quit Firefox, I would require users to manually start it again after starting Firefox back up.
2. If you open the Manager tab, start the Simulator, close the Manager tab, and then undo close, then the Simulator stops. This doesn't happen if you reopen the Manager tab by selecting _Tools > Web Developer > Firefox OS Simulator_; and it seems like it shouldn't happen.

digitarald:
The reason for both behaviors comes from FF resetting the form values after the page loads; which triggers a change event on the checkbox in this case. I'll fix that.

PR #156, opened 2012-11-26
Moved alpha badge to the left, fixes #61.

### LINKED ISSUE #61: manager always has horizontal scrollbar
No matter how wide I make the browser window, the manager always has a horizontal scrollbar. It should fit into the available space provided the browser window is a reasonable size.

digitarald:


mykmelez:
Ooh, making space to add a ["Fork me on GitHub" ribbon](https://github.com/blog/273-github-ribbons) on the right-hand side! ;-)

Just kidding, I bet @dangoor would not go for that. And he'd probably be right!

In any case, not only does this fix the scrolling problem, it looks pretty good on the left-hand side, right next to the name of the product!

I've been wondering lately whether "alpha" is quite the right word to put on it, but that's something to consider in another issue.

PR #174, opened 2012-12-04
set appStatus to 2 for local (i.e. packaged) apps to identify them as privileged, not certified

mykmelez:


mykmelez:
@digitarald: this is what we should be doing, yes?

digitarald:
@mykmelez Yes, it makes sense at this point. I'd recommend using the [`type`](https://developer.mozilla.org/en-US/docs/Apps/Manifest#type) provided in the manifest later, as developers need to set that anyways when they require extra permissions.

PR #186, opened 2012-12-06
Feature desktop permissions

digitarald:
## Problem

Firefox Desktop provides the best developer tools. Privileged APIs can't be accessed, so developers have to test on mobile devices at one point, where tools are limited by remote-capability.
Developers need to be able to test privileged APIs using Desktop Firefox.
## Customers

Gaia and FxOS developers
## Solution

Adding an app in Firefox OS Simulator installs the app in B2G Desktop and additionally adds the correct permissions to Firefox (P1).
The app installed in B2G Desktop should also get the requested APIs (P2).
## Implementation

**For Desktop**: Override _PermissionSettings_ [3](http://mxr.mozilla.org/mozilla-central/source/dom/permission/PermissionSettings.jsm) in FxOS Simulator, so _PermissionsInstaller.installPermissions_ [1](http://mxr.mozilla.org/mozilla-central/source/dom/apps/src/PermissionsInstaller.jsm) calls our _PermissionSettings.addPermission_, which will call _nsIPermissionManager.add_ [4](http://mxr.mozilla.org/mozilla-central/source/netwerk/base/public/nsIPermissionManager.idl) instead of _~.addFromPrincipal_.
**For Simulator**: Use Webapps API [2](http://mxr.mozilla.org/mozilla-central/source/dom/apps/src/Webapps.jsm#1196) from the _shell.js_ (command line or reading to files read on B2G Desktop launch)

References:

From https://etherpad.mozilla.org/fxoss-desktop-permissions

cc @fabricedesre, @myk

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -1039,3 +1039,69 @@ function archiveDir(zipFile, dirToArchive) {
 
   console.log("archived dir " + dirToArchive);
 }
+
+
+
+
+// FIXME: Hack
+
+let PermissionsInstaller = Cu.import("resource://gre/modules/PermissionsInstaller.jsm").PermissionsInstaller;
+
+let PermissionSettings = Cu.import("resource://gre/modules/PermissionSettings.jsm");
+let ManifestHelper = Cu.import("resource://gre/modules/AppsUtils.jsm").ManifestHelper;
+
+function CustomAddPermission(aData, aCallbacks) {
+  setTimeout(function() {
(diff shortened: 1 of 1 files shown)


mykmelez:
Looks good! I don't think this will break any native browser functionality, although it wouldn't be the first time I was wrong about something like that. In any case, it seems worth doing; we should just stay aware of any changes in the native module (or the browser's use of it) and update our code accordingly.

PR #194, opened 2012-12-12
update addon-sdk submodule to 1.12 release

mykmelez:
Add-on SDK 1.12 was released today. It isn't urgent for us to update to it, because it doesn't give us any necessary features, and AMO will continue supporting 1.11 for some time.

But the new version makes some changes to the way modules are organized and referenced that pave the way for landing the SDK's modules in Firefox, and I'd rather update to each new version, making a small number of changes each time, than jump several versions when it becomes necessary and have to figure out a large delta of changes.

So here's a branch that makes the necessary (and a few optional) changes to work with the new version of the SDK. Remember to call `git submodule update` after checking out this branch!

@digitarald, does this look good to you?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@ profile:
 	rm -rf addon/template
 	mkdir -p addon/template
 	mv gaia/profile addon/template/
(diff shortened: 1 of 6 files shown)


digitarald:
Works great for me.

Issue #177, opened 2012-12-04
jquery library differs from official release version


mykmelez:
Kris Maglione notes in his preliminary review of Simulator 1.0:

> The copy of jQuery that you include doesn't match the checksum of the official release, which we generally don't allow. I'll accept it, since the changes seem to be whitespace or encoding related, but in the future you should only include bit-identical official releases.

I'm not sure why they differ, but they shouldn't.

PR #213, opened 2012-12-13
jquery library differs from official release version

mykmelez:
Fixes issue #177: jquery library differs from official release version, by instructing Git not to mess with line endings and then replacing our copy of jQuery with the release version. I also take the opportunity to update jQuery from 1.8.0 to 1.8.3.

@digitarald, does this make sense to you?


digitarald:
Good catch, less git magic!

PR #220, opened 2012-12-19
Permissions added for each app get stored and removed on uninstall.

### LINKED ISSUE #192: addon doesn't remove permissions when user removes app
When the user removes an app, the addon doesn't remove its permissions from Firefox's permissions database. Which means the app will continue to get those permissions whenever the user loads it into a tab, even though it didn't have those permissions before the user added and then removed it.

That isn't what the user expects, and it could result in the app having permissions the user doesn't want it to have. So the addon should remove those permissions from the database when the user removes the app.

@digitarald: from our conversation in IRC this morning, it sounded like you were looking at this, so I'm assigning it to you. Please unassign it if that isn't the case!

digitarald:
Fixes #192.

@myk: Does that look sane?

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -45,7 +45,7 @@ let simulator = {
   _worker: null,
 
   get apps() {
-    return SStorage.storage.apps || {};
+    return SStorage.storage.apps || (SStorage.storage.apps = {});
   },
 
   set apps(list) {
@@ -54,7 +54,7 @@ let simulator = {
 
 
   get permissions() {
-    return SStorage.storage.permissions || {};
+    return SStorage.storage.permissions || (SStorage.storage.permissions = {});
   },
 
   set permissions(list) {
@@ -131,8 +131,6 @@ let simulator = {
         manifest: webapp,
       }
       console.log("Stored " + JSON.stringify(apps[webappFile]));
-
-      simulator.apps = apps;
 
       this.updateApp(webappFile, true);
     }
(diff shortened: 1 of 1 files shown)


addon/lib/main.js
```diff
@@ -315,6 +324,17 @@ let simulator = {
...
 
+    let permissions = simulator.permissions;
+    if (permissions[config.origin]) {
+      let host = config.host;
+      permissions[config.origin].forEach(function(type) {
+        permissionManager.remove(host, type);
+      });
+      delete permissions[config.origin];
+    }
+    simulator.permissions = permissions;
```
mykmelez: If `simulator.permissions` always returned `SStorage.storage.permissions`, then it wouldn't be necessary to reassign `simulator.permissions` after modifying `permissions`, because they would both be references to the original object, which would have been modified in place. That's not just one less line of code, it's also a simpler model, since it makes `simulator.permissions` always be a reference.

So I would make the `simulator.permissions` getter always return `SStorage.storage.permissions` via something like the following:

```
get permissions() {
  return SStorage.storage.permissions ||
         (SStorage.storage.permissions = {});
},
```

And then I would remove the `simulator.permissions` setter, so it isn't possible to set it to a different object, just to retrieve and modify it (which is sufficient for our purposes).

mykmelez:
@digitarald: for some reason GitHub doesn't notify me when you add a commit to a pull request in which I've commented. So I just assigned this pull request to myself, and hopefully that means I'll get notified when you add a commit to address that last comment. But just in case, ping me on IRC once you've done so!

mykmelez:
Argh, so I didn't get notification. How silly of GitHub. Oh well. In any case, this looks good!

PR #232, opened 2012-12-23
Control a b2g instance using debugger protocol

rpl:
- prosthesis (b2g side)
  - added a -b2gport option to choose debugger server port from cli arguments
  - added a -pbport option to choose pingback server to alert when b2g instance is loaded and ready
  - patch startDebugger to load custom actors from simulator-actors.js into the debugger server
  - added a SimulatorActor and a minimal command set (ping, getBuildID, runApp, subscribeWindowManagerEvents,
    unsubscribeWindowManagerEvents)
- r2d2b2g (firefox side)
  - refactored out b2g process spawn and control code from main.js to remote-simulator-client.js
  - added a pingback-server module, it spawns a pingback server and emit pingbackCompleted and
    pingbackTimeout events (used by remote-simulator-client to detect stalled b2g instance)
  - added a remote-simulator-client module, search a free port to allocate b2g remote debugger port,
    compute b2g executable and cli arguments, run a b2g instance, manage b2g process, route low
    level events to high level ones (ready, timeout, stdout, stderr etc.), provide helpers to
    send/receive events to/from remote simulator actor
  - added Run button on installed application (run application on a b2g running instance or
    start a new instance and run application when b2g is ready)
  - added tests for pingback-server and remote-simulator-client modules
- Makefile
  -     added optional BIN and TEST Makefile options on run/test tasks 

How to run simulator on a specified firefox binary:
     make run BIN=/path/to/firefox

How to run tests on a specified firefox binary:
     make test BIN=/path/to/firefox

How to run filtered tests on a specified firefox binary:
     make test TEST=myfilter


mykmelez:
Awesome! I'm working my way through this, and the first thing I noticed upon running `make && make test` is that two tests time out (causing other tests not to run) on Windows and Mac:

```
14 of 16 tests passed.

The following tests failed:
  test-remote-simulator-client.test:RemoteSimulatorClient isExitedWithoutReady: timed out
  test-remote-simulator-client.test:RemoteSimulatorClient run/onReady/onStdout/onExit: timed out
```

While the tests are running, two alert dialogs appear with the title "Close B2G" and a message that indicates that an instance of B2G is already running. On Mac, the message is "A copy of B2G is already open. Only one copy of B2G can be open at a time." On Windows, it's "B2G is already running, but is not responding. To open a new window, you must first close the existing B2G process, or restart your system."

I also tested on Linux (Ubuntu 12.10), where I don't see this problem (23 of 23 tests pass). So it seems specific to Windows and Mac. Perhaps there is some kind of race condition that causes tests to try to spawn additional B2G processes instead of reusing existing processes.

The second problem I encountered while running tests is the following exception on all three platforms:

```
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/timers.js", line 31, in notify
    callback.apply(null, args);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/test/harness.js", line 191, in cleanup
    loader.unload();
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/test/loader.js", line 25, in exports.Loader/<.unload
    unload(loader, reason);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/toolkit/loader.js", line 313, in unload
    notifyObservers(subject, 'sdk:loader:destroy', reason);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/events.js", line 62, in 
    data: data
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/unload.js", line 77, in onunload
    unload(reason);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/unload.js", line 56, in unload
    observers.forEach(function(observer) {
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/unload.js", line 58, in 
    observer(reason);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/unload.js", line 67, in 
    unloaders.slice().forEach(function(unloadWrapper) {
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/unload.js", line 68, in 
    unloadWrapper(reason);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/system/unload.js", line 41, in unloadWrapper
    originalDestructor.call(obj, reason);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/window-utils.js", line 107, in unload
    this._unregWindow(window);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/window-utils.js", line 98, in _unregWindow
    this._delegate.onUntrack(window);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/windows/firefox.js", line 236, in _onUntrack
    window.destroy();
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/windows/firefox.js", line 73, in BrowserWindowTrait<.destroy
    destroy: function () this._onUnload(),
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/windows/firefox.js", line 88, in BrowserWindowTrait<._onUnload
    this._destroyWindowTabTracker();
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/windows/tabs-firefox.js", line 83, in _destroyWindowTabTracker
    this._emitEvent("close", tab);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/windows/tabs-firefox.js", line 110, in _emitEvent
    tabs._emit(type, tab);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/events.js", line 123, in _emit
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/events.js", line 153, in _emitOnObject
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/main.js", line 823, in 
    simulator.kill();
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/main.js", line 542, in simulator.kill
    onKilled();
TypeError: onKilled is not a function
```

Interestingly, it does not appear to cause tests to fail. Perhaps it happens on Firefox shutdown, after tests have already completed?

rpl:
@mykmelez I've just fixed the exception when simulator.kill is called without a callback (test didn't fail because "simulator.kill()" was used in the main.js onunload  handler).

One of the two fails on Mac and Windows is probably due to different b2g-desktop behavior on the various platform,
on Linux it silently exit 0 (and I added  isExitedWithoutReady helper and test case to detect this behavior).

I'm going to reboot one of my pc... :-)

addon/lib/main.js
```diff
@@ -26,22 +26,10 @@ Cu.import("resource://gre/modules/Services.jsm");
...
-
-    if (this.worker) {
-      this.worker.postMessage({
-        name: "isRunning",
-        isRunning: !!this.process,
-      });
-    }
-  },
+const RemoteSimulatorClient = require("remote-simulator-client").RemoteSimulatorClient;
+const PingbackServer = require("pingback-server").PingbackServer;
```
mykmelez: When modules export a single value named after the module, you can reduce redundancy and complexity by setting the `exports` object to the value being exported.

In _remote-simulator-client.js_:

```
module.exports = RemoteSimulatorClient;
```

In _pingback-server.js_:

```
module.exports = PingbackServer;
```

Then these `require` statements can be simply:

```
const RemoteSimulatorClient = require("remote-simulator-client");
const PingbackServer = require("pingback-server");
```

rpl: fixed exports + removed 1 unused require

addon/lib/main.js
```diff
@@ -544,11 +531,15 @@ let simulator = {
...
 
   kill: function(onKilled) {
-    if (this.process && !this.shuttingDown) {
-      this.onKilled = onKilled;
-      this.shuttingDown = true;
-      this.process.kill();
-    } else if (onKilled) {
+    if (onKilled)
+      onKilled = onKilled.bind(this);
+    else
```
mykmelez: It seems dangerous to bind `onKilled` in `kill()`, since `onKilled` gets passed into `kill()`, which doesn't know what it does.  If `onKilled` needs to be bound, it would be better to do it wherever `kill()` gets called, even if that means doing it at multiple call sites, to avoid a situation where `kill()` gets called in the future by a caller that isn't aware of the binding and passes an `onKilled` function that misbehaves when bound.

addon/lib/main.js
```diff
@@ -544,11 +531,15 @@ let simulator = {
...
+      onKilled = function () {}; // dummy callback function
+    // WORKAROUND: currently add and update an app will be executed
```
mykmelez, line 539: It shouldn't be necessary to define `onKilled` if it isn't defined.  Just check that `typeof onKilled == "function"` before calling it.

Finally, nit: the emerging code convention in Firefox seems to be to put braces around all if/else clauses, even those comprising a single statement on a single line; so we've been doing that in r2d2b2g code too (even though I personally prefer to omit braces around single-statement/line clauses).

addon/lib/main.js
```diff
@@ -585,12 +576,82 @@ let simulator = {
...
+          });
+        }
+      },
+      onTimeout: function () {
+        if (simulator.worker) {
+          simulator.worker.postMessage({
+            name: "remoteDeveloperToolboxLog",
+            msg: "ERROR: Simulator Instance Timeout\n"
+          });
+        }
```
mykmelez: It doesn't seem like this message gets handled by anything.

addon/lib/main.js
```diff
@@ -585,12 +576,82 @@ let simulator = {
...
+    
+    this._remoteSimulator = remoteSimulator;
+    return remoteSimulator;
+  },
+
+  connectRemoteDeveloperToolbox: function() {
+    console.log("Simulator.connectRemoteDeveloperToolbox");
+    this.remoteSimulator.connectDeveloperTools();
+  },
+
```
mykmelez: It doesn't look like this code gets used either. Perhaps because you moved the caller to that separate "open devtools" branch? If so, then maybe this should get moved to that branch as well.

addon/lib/main.js
rpl: unused code (old open devtools experiment)

addon/lib/main.js
rpl: definitely!
I'm going to move in the other branch (and remove from here)

addon/lib/pingback-server.js
```diff
@@ -0,0 +1,56 @@
+const { Cc, Ci, Cr, Cu } = require("chrome");
```
mykmelez, line 8: Some of our existing files lack license headers, but we should make sure to include them in any new files (we'll also fix the existing files eventually). The license we're using for this project is MPL 2.0, so the license header for this and other new JavaScript files, per [Mozilla License Headers](http://www.mozilla.org/MPL/headers/), is:

```
/* This Source Code Form is subject to the terms of the Mozilla Public
```

addon/lib/pingback-server.js
```diff
@@ -0,0 +1,56 @@
...
+
+const { EventTarget } = require("sdk/event/target");
+const { emit } = require("sdk/event/core");
+const { Class } = require("sdk/core/heritage");
+const { setTimeout, clearTimeout } = require('sdk/timers')
```
mykmelez: Nit: trailing semicolon.

addon/lib/pingback-server.js
mykmelez, line 8: You `'use strict';` in _remote-simulator-client.js_, and it probably wouldn't hurt to do that here too (eventually we should be doing it in all our JavaScript files).

addon/lib/remote-simulator-client.js
```diff
@@ -0,0 +1,415 @@
...
+    // and kill the stalled instance
+    this.once("pingbackTimeout", (function() {
+      emit(this, "timeout", null);
+      this.kill();
+    }).bind(this));
+
+    // on pingbackCompleted, track a completed pingback and start
+    // debugger protocol connection
+    this.on("pingbackCompleted", (function() {
+      console.log("DEBUG rsc.onPingbackCompleted");
```
mykmelez: If these `console.log()` calls whose message starts with "DEBUG" are intended to be debug-level log output, you could instead log them via `console.debug()`! We don't yet use that method elsewhere in the codebase, but we might as well start doing so (along with `console.warn()` and `console.error()`, where appropriate); especially since SDK 1.13 will have support for setting a log level, with the default level for `cfx run` being _info_ (which will show messages from `console.error()`, `console.warn()`, `console.info()`, and `console.log()`), while the default level for an addon installed by a user is _error_ (which will only show messages from `console.error()`).

addon/lib/remote-simulator-client.js
```diff
@@ -0,0 +1,390 @@
...
+    }
+    return this._pingbackServer.port;
+  },
+
+  // NOTE: find a port for remoteDebuggerPort if it's null or undefined
+  get remoteDebuggerPort() {
+    var port = this._foundRemoteDebuggerPort;
+
+    if (port) {
+      return port
```
mykmelez: Nit: trailing semicolon

addon/lib/remote-simulator-client.js
```diff
@@ -0,0 +1,390 @@
...
+    this._foundRemoteDebuggerPort = port;
+  },
+
+  get jsConsoleEnabled() {
+    return Prefs.get("extensions.r2d2b2g.jsconsole", false);    
+  }
+});
+
+module.exports = RemoteSimulatorClient;
\ No newline at end of file
```
mykmelez: Nit: add a newline to the end of the file.

addon/test/test-pingback-server.js
```diff
@@ -0,0 +1,101 @@
...
+
+  pbs.start();
+  pbs.startTimeout(1000);
+};
+
+exports["test:PingbackServer stopTimeout"] = function(test, done) {
+  var pingback_timeout = false;
+  var pbs = new PingbackServer({
+    onTimeout: function () {
+      test.pass("onTimeout callback");
```
mykmelez: Since the test is never supposed to call this callback, you don't need to set and check `pingback_timeout`; just call `test.fail()` instead of `test.pass()` at this point.

addon/test/test-pingback-server.js
```diff
@@ -0,0 +1,101 @@
...
+
+  setTimeout(function () {
+    test.ok(!pingback_timeout, "onTimeout should not be called");
+    done();
+  }, 800);
+};
+
+
+require('sdk/test').run(exports);
\ No newline at end of file
```
mykmelez: Nit: add a newline to the end of this file too (and any others). Note that `git diff --check` will show you these (and other whitespace errors).

prosthesis/content/dbgport.js
```diff
@@ -0,0 +1,54 @@
...
+      dump("prosthesis:"+msg+"\n");
+  }
+
+  log("processing -dbgport command line option");
+
+  // Get the command line arguments that were passed to the b2g client
+  let args = window.arguments[0].QueryInterface(Ci.nsICommandLine);
+  let dbgport;
+ 
+   // Get the --screen argument from the command line
```
mykmelez: Nit: --screen -> --dbgport

prosthesis/content/pbport.js
```diff
@@ -0,0 +1,54 @@
...
+    if (DEBUG_LOG)
+      dump("prosthesis:"+msg+"\n");
+  }
+
+  log("processing -pbport command line option");
+  // Get the command line arguments that were passed to the b2g client
+  let args = window.arguments[0].QueryInterface(Ci.nsICommandLine);
+  let pbport;
+ 
+   // Get the --pbport argument from the command line
```
mykmelez, line 14: Nit: --screen -> --pbport

prosthesis/content/pbport.js
```diff
@@ -0,0 +1,54 @@
...
+  function usage() {
+    let msg = 'The --pbport argument specifies the desired port to pingback when ready.\n' +
+      'Use it like this:\n'+
+      '\t--pbport PORT (e.g. --pbport 5431243)\n';
+    dump(msg);
+    // exit b2g
+    Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+  }
+
+  function log(msg) dump(msg+"\n")
```
mykmelez: This `log` function appears to be redundant with the one at the top of the file.

prosthesis/content/simulator-actors.js
```diff
@@ -0,0 +1,189 @@
...
+
+    window.runAppObj = new window.AppRunner(appName);
+
+    let setReq = window.navigator.mozSettings
+      .createLock().set({'lockscreen.enabled': false});
+    setReq.onsuccess = function() {
+      // NOTE: 100ms is not enought
+      window.setTimeout(function() {
+        window.runAppObj.doRunApp();
+      }, 1000);
```
mykmelez: What happens if you try to run the app without setting _lockscreen.enabled_ to `false`? It's unfortunate that we have to wait some indeterminate amount of time after setting this setting before running the app. In my experience, it's nearly impossible to set such a timeout correctly, because of the tradeoffs between making it slow enough to work on slow machines while fast enough to seem performant on fast machines.

prosthesis/content/simulator-actors.js
```diff
@@ -0,0 +1,189 @@
...
+    var shellw = this.simulatorWindow.document.getElementById("homescreen").contentWindow;
+    return shellw;
+  },
+
+  get simulatorWindow() {
+    var window = Cc['@mozilla.org/appshell/window-mediator;1']
+      .getService(Ci.nsIWindowMediator)
+      .getMostRecentWindow("navigator:browser");
+    return window;
+  }
```
mykmelez: Nit: it seems to be an emerging Firefox convention to include a trailing comma at the end of an object literal. The archaic JavaScript strict interpreter in SpiderMonkey used to warn about it, since it wasn't legal ECMAScript, but that is changing (or has already changed?). And including a trailing comma means that the next time you add a property to the object, you don't need to change the last line of the previous property declaration.

prosthesis/content/simulator-actors.js
```diff
@@ -0,0 +1,189 @@
...
+}
```
mykmelez: Nit: trailing semicolon.

prosthesis/content/simulator-actors.js
rpl: it was a workaround used in the runapp.js packaged inside b2g-desktop,
and apparently we don't need it because we wait b2g to be ready before 
send a runApp command :-)

so I'm very happy to remove this not needed workaround :-D

rpl:
I've just finished to run test on windows and I can confirm b2g-desktop open a modal dialog on windows when another instance is already running, and this break the second remote-simulator-client test case.

I'm going to remove the isExitedWithoutReady test case so we can retry on Mac too.

rpl:
actually I was that developer :-)

I bumped it because the 100ms workaround was not enough before I moved that code into the SimulatorActor runApp command handler :-P

mykmelez:
With the latest changes, the tests all passed on one of my tests runs on Mac; but they still failed on four other test runs, each time with "test-remote-simulator-client.test:RemoteSimulatorClient run/onReady/onStdout/onExit: timed out." When they fail, a B2G process is left running after the tests finish.

rpl:
github issues doesn't support file attachment :-(
can you email me a full make test output of that single testcase ("make test TEST=remote-simulator"), 
it could help me to detect where it timeouts.
even better if I can compare output of the two cases (fail and pass).

mykmelez:
I've been looking into the test module _sdk/test_, and it looks like it's built on top of a deprecated test module called _unit-test_ that sets a 10000ms timeout by default. Direct users of _unit-test_ can change that by calling `test.waitUntilDone(timeoutInMS)`, but it isn't clear how to set a higher timeout using _sdk/test_. I suspect that default timeout is the reason I'm seeing this problem; it just isn't enough time to run the test on my (apparently slow) Mac.

In any case, I'll email you the output of a passing run and a failing run!

mykmelez:
I emailed you that output. I also tried changing _DEFAULT_PAUSE_TIMEOUT_ from 10000 to 30000 in `addon-sdk/lib/sdk/deprecated/unit-test.js`, and the tests started working consistently!

So the problem is that the default timeout is set too low for my machine, which is strange, since my machine is not that old (a MacBook Pro from summer 2010). Now we just have to figure out how to change that timeout. It's possible that _sdk/test_ doesn't provide that option, in which case we'll have to use _unit-test_.

rpl:
A possible source of delay on Mac could be related to the osascript workaround:
I would try to move it back to the first stdout event (as it was before remote-simulator-client),
because onReady could be a very busy moment :-D 

I pushed this experimental "wannabe" fix

mykmelez:
It looks like _sdk/test_ is intended to be a test harness compatible with the [CommonJS Unit Testing 1.1 spec](http://wiki.commonjs.org/wiki/Unit_Testing/1.1), which doesn't say anything about a default asynchronous test timeout or setting your own timeout.

But its code [calls `waitUntilDone()`](https://github.com/mozilla/addon-sdk/blob/1.12/lib/sdk/test.js#L63) from the deprecated _unit-test_ module, without specifying a timeout, so it gets the default timeout, and it isn't possible to override it in our test.

rpl:
... so finger crossed for the "wannabe" fix 
f871f47 please do the magic :-)

mykmelez:
That change may help, but it seems unlikely to do much. The only thing _osascript_ does is move B2G to the foreground. I suppose that may cause the process to become prioritized by the system and thus initialize faster. But the test still seems prone to timing out.

In any case, I tested it, and it did seem to make things better. The test started finishing instead of timing out. But it also started throwing an exception:

```
info: r2d2b2g: executing 'test-remote-simulator-client.test:RemoteSimulatorClient run/onReady/onStdout/onExit'
info: r2d2b2g: rsc.remoteDebuggerPort: found free port  51683
debug: r2d2b2g: WORKAROUND run osascript to show b2g-desktop window on Runtime.OS=='Darwin'
  File "resource://r2d2b2g-at-mozilla-dot-org/subprocess/lib/subprocess.js", line 1517, in readPipes/stdoutWorker</<
    options.stdout(data);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/remote-simulator-client.js", line 146, in 
    emit(this, "stdout", data);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/event/core.js", line 83, in emit
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/event/core.js", line 101, in lazy
    yield listeners.shift().apply(target, args);
  File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/event/core.js", line 61, in observer
    listener.apply(target, arguments);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/remote-simulator-client.js", line 129, in RemoteSimulatorClient<.run/<
    let path = remoteSimulator.b2gExecutable.path
ReferenceError: remoteSimulator is not defined
```

And if I change `DEFAULT_PAUSE_TIMEOUT` from 10000 to 9000 in _unit-test_, then the test starts timing out again. So the change may have made things better. But only barely. And a slightly slower machine than mine will run into the problem even with this change.

However, some good news: it turns out that there's a simple way to change the timeout! The `test` object passed to the test function is an instance of `Assert` from the _assert_ module, and it has a `_log` property that happens to be set to the `options` object from _unit-test_, which is what provides `waitUntilDone()`!

And `waitUntilDone()` can be called multiple times. It resets the timeout each time it's called. So even though it has already been called for us by _sdk/test_, we can call it again from within our test function to set a higher timeout.

So the simple fix is to call `test._log.waitUntilDone(30000)` (or some number larger than the number of milliseconds we think the test will take on any machine we care about) at the beginning of the test function!

In the long run we'll want to confer with the Add-on SDK developers about the right way to write this test, of course. But that workaround will get things working for us now. :-)

rpl:
the previous exception was due to a bug introduced moving osascript workaround into remote-simulator-client,
should be fixed in 3f078d9

33935fc set an higher timeout as suggested, to prevents random fails

mykmelez:
It works! Woohoo! So, it looks like this is ready to merge, yes?

rpl:
:+1: 

fantastic!

I can squash the history using "git rebase -i" if you prefer ;-)

mykmelez:
No worries, I don't mind the history! I'll merge now...

rpl:
I am proud to be part of this!

mykmelez:
Thanks, Luca, this is a great contribution! And it's a solid foundation on which to build the next set of great Simulator features (and improve the existing ones!).

PR #236, opened 2012-12-25
Fix prosthesis debugger server 

rpl:
fix how prosthesis enable and start debugger server
(due to changes introduced by https://bugzilla.mozilla.org/show_bug.cgi?id=815522)
and "simulatorActor" exception on debugger server restart.

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -32,17 +32,12 @@ const RemoteSimulatorClient = Class({
     this._hookInternalEvents();
   },
   // check if b2g is running and connected
-  get isConnected() {
-    return !!this.process && !this._clientConnected;
-  },
+  get isConnected() this._clientConnected,
   // check if b2g is running
-  get isRunning() {
-    return !!this.process;
-  },
+  get isRunning() !!this.process, 
   // check if b2g exited without reach a ready state
-  get isExitedWithoutReady() {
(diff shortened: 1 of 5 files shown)


rpl:
As a side effect of 5d176e8, during the first launch "devtools.debugger.remote-enabled" is undefined
so startDebugger will be called 1 time, but on any next launch "devtools.debugger.remote-enabled" 
will be already true and we set it to true anyway, then startDebugger will be called twice.

5207082 prevents race conditions due to this.

rpl:
> I would use console.warn() here, since we don't expect it to happen.

console.warn causes a "side effect fail" during tests (it's a addon-sdk unit-test feature)

startDebugger it's called twice because when that setting is already true, settings listeners registered by b2g will be called twice.

we can workaround that using an option to suppress that warning during tests (but retain the log level until we found a working solution, I'm trying to get/set the setting but it's not working correctly).

rpl:
or we can add this warning in a "known problems / test fails" list :-)

rpl:
c774b4e seems to do the magic (get/set remote-tools.enabled)

rpl:
My fault! I tested a lot of workarounds... and finally I picked the wrong one :-)
and I didn't notice that I forgot to dump the exception in the catch block :-(

thanks to your patch "patch-start-debugger.js" is much more cleaner!

mykmelez:
Merged, thanks for the fix!

PR #263, opened 2013-01-23
FIX: install/update app with permissions on b2g instance

### LINKED ISSUE #237: install/update apps using mozApps API
We currently install and update apps using a very hacky technique: we manually edit the webapp registry. That requires us to restart B2G after installing/updating an app, prevents us from registering permissions for the app, and probably has all sorts of other undesirable consequences (like #181).

We should install/update apps using the mozApps API instead. And now that @rpl has implemented an addon <-> B2G bi-directional communications channel, we can do it.

rpl:
This change hooks the new remote webappsActor to install apps into a b2g instance
using mozApps internals, so application permissions will be registered and
correctly available.

This merge should fix: #237

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -126,19 +126,21 @@ let simulator = {
     }
   },
 
-  get webappsDir() {
+  get tempDir() {
     let sandbox = {};
     XPCOMUtils.defineLazyServiceGetter(sandbox, "gDirService",
                                        "@mozilla.org/file/directory_service;1",
                                        "nsIProperties");
-    return File.join(URL.toFilename("file://"+sandbox.gDirService.get("TmpD", Ci.nsILocalFile).path), "b2g");
+
+    let basePath = sandbox.gDirService.get("TmpD", Ci.nsIFile).path;
+    return File.join(basePath, "b2g");
   },
 
   updateApp: function(id, manual) {
     console.log("Simulator.updateApp " + id);
 
     let webappsDir = URL.toFilename(profileURL + "webapps");
-    let TmpDwebappsDir = this.webappsDir;
+    let tempDir = this.tempDir;
     let webappsFile = File.join(webappsDir, "webapps.json");
     let webapps = JSON.parse(File.read(webappsFile));
 
@@ -152,7 +154,12 @@ let simulator = {
 
     if (!config.xid) {
       config.xid = ++[id for each ({ localId: id } in webapps)].sort(function(a, b) b - a)[0];
-      config.xkey = "myapp" + config.xid + ".gaiamobile.org";
+
+      // NOTE: we can't currently use localId as code because webappsActor
+      // if the emulator doesn't exit clean DOMApplicationRegistry._nextLocalId
+      // will not be able to serialize "dom.mozApps.maxLocalId" and will be the
+      // same for new app installed
+      config.xkey = UUID.uuid().toString(); //"myapp" + config.xid + ".gaiamobile.org";
 
       if (!config.origin) {
         config.origin = "app://" + config.xkey;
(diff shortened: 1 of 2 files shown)


digitarald:
Hooray!

Only issue I run into when using the actor (from my own xpcshell script), was that re-installs didn't work as expected. Like adding/removing fullscreen or other properties from the manifest wasn't recognized until I uninstalled the app and rebooted the phone.

rpl:
Currently I left intact the method we use to remove applications and I get a reasonable doubt that in this way some information (such as permissions that are registered and de-registered during the installation and uninstallation) will be left in the system.

I need to go a bit deeper (although in the end the real solution is surely to provide the opposite function into an debugger server actor)

addon/lib/remote-simulator-client.js
```diff
@@ -225,6 +226,15 @@ const RemoteSimulatorClient = Class({
                           onResponse);
   },
 
+  // send an install request to the remote webappsActor
+  install: function(appId, appType, onResponse) {
+    let remote = this._remote;
+
+    remote.client.request({to: remote.webapps, type: "install", appId: appId,
+                           appType: appType}, 
+                          onResponse);
```
mykmelez: Nit: this would be easier to read if the object argument was more structured:

```
remote.client.request({ to: remote.webapps,
                        type: "install",
                        appId: appId,
                        appType: appType },
                      onResponse);
```

That also makes the lines short enough to enable you to call request() directly on this._remote.client without exceeding 80 characters per line, so you don't have to assign it a local variable that then only gets used once:

```
this._remote.client.request({ to: remote.webapps,
                              type: "install",
                              appId: appId,
                              appType: appType },
                            onResponse);
```

addon/lib/main.js
```diff
@@ -126,10 +126,19 @@ let simulator = {
     }
   },
 
+  get webappsDir() {
+    let sandbox = {};
+    XPCOMUtils.defineLazyServiceGetter(sandbox, "gDirService",
+                                       "@mozilla.org/file/directory_service;1",
+                                       "nsIProperties");
+    return File.join(URL.toFilename("file://"+sandbox.gDirService.get("TmpD", Ci.nsILocalFile).path), "b2g");
```
mykmelez: I don't understand this line of code. `File.join()` takes two path parts, and this generates the first by 1. getting a path, 2. converting it to a URL via string concatenation, 3. converting it back to a path via `URL.toFilename()`. But isn't the first step sufficient? I.e. simply:

```
return File.join(sandbox.gDirService.get("TmpD", Ci.nsILocalFile).path, "b2g");
```

If you really do need to convert the path to a URL, you should use `URL.fromFilename()` to do it, which would make this:

```
return File.join(URL.toFilename(URL.fromFilename(sandbox.gDirService.get("TmpD", Ci.nsILocalFile).path)), "b2g");
```

But then you should really add a code comment explaining why it's necessary to convert the path to a URL only to turn right around and convert it back again. And wrap the line so it doesn't exceed 80 characters in length.

mykmelez: One more nit: nsILocalFile is deprecated, so use nsIFile instead.

addon/lib/main.js
```diff
@@ -126,10 +126,19 @@ let simulator = {
     }
   },
 
+  get webappsDir() {
```
mykmelez: Nit: it would be clearer to call this getter something like _tempDir_, since there are other directories in this codebase that are called _webappsDir_.

addon/lib/main.js
```diff
@@ -173,90 +182,100 @@ let simulator = {
...
+        let metadata = {
+          origin: config.origin,
+          manifestURL: id,
+        };
+        File.open(metadataFile, "w").writeAsync(JSON.stringify(metadata, null, 2), function(err) {
+          simulator.info(config.name + " (hosted app) installed in Firefox OS");
+        
+          // Complete install (Hosted)
+          if (manual) {
+            simulator.defaultApp = null; // id; // DISABLED: because on the first run the app will be not already installed
```
mykmelez: Nit: same as for above comment.

addon/lib/main.js
```diff
@@ -664,12 +691,7 @@ let simulator = {
 
         // NOTE: if a b2g instance is already running send request
         //       or start a new instance and send request on ready
-        if (this.isRunning) {
-          cmd();
-        } else {
-          this.remoteSimulator.once("ready", function() cmd());
-          this.run();
-        }
+        this.run(cmd);
```
mykmelez, line 686: Nice simplification! It makes more sense to encapsulate this logic in the `run()` function.

addon/lib/main.js
mykmelez: Erm, and I just realized: _main.js_ imports _Services.jsm_, which provides _dirsvc_, so we don't need to get it ourselves, and this can be as simple as:

```
get tempDir() {
  return File.join(Services.dirsvc.get("TmpD", Ci.nsIFile).path, "b2g");
},
```

rpl:
I've introduced suggested code cleanup and fixed a regression (due to new app install method and b2g install forced kill)

In the last commit I've removed from tab worker message handler any simulator.kill used from the previous install/upgrade method.

I've not been yet able to study more deeply potential problems related to the cases mentioned by @digitarald

mykmelez:
Thanks for the fixes!

PR #290, opened 2013-02-04
Show connect devtools button or allocated remote debugger port 

rpl:
Change needed to help developers to connect the upcoming remote developer tools to debug or interact with remote webconsole on the running b2g instance:
- show connect button on Firefox >= 20.0a1 (where connect devtools xhtml page is available)
- show allocated remote debugger port on previous Firefox releases as fallback

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -76,6 +76,12 @@ var Simulator = {
               $(Simulator.toggler).prop('checked', true);
               remoteDebuggerPortEl.html(message.remoteDebuggerPort);
               remoteDebuggerPortEl.parents('label').show();
+              // NOTE: show connect devtools buttons only when it's supported
+              if (message.hasConnectDevTools) {
+                $("#open-connect-devtools").show();
+              } else {
+                $("#open-connect-devtools").hide();
+              }
             }
             else {
               $(Simulator.toggler).prop('checked', false);
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -242,6 +242,10 @@ var Simulator = {
 
   addAppByDirectory: function() {
     window.postMessage({ name: "addAppByDirectory" }, "*");
+  },
+
+  openConnectDevTools: function() {
+    window.postMessage({ name: "openConnectDevTools" }, "*");
   }
 
 };
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -71,11 +71,16 @@ var Simulator = {
         switch (message.name) {
           case "isRunning":
             $(Simulator.toggler).prop('indeterminate', false);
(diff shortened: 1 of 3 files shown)


mykmelez:
@rpl and I chatted a bit over email about showing the port on older versions of Firefox:

@rpl wrote:

> What about if we show the port on the ui at least on firefox releases where connect.xhtml is not available? because a developer can choose to connect manually from a different firefox instance (where remote tools are available)

Then I responded:

> Hmm, that seems like a complicated way to provide this functionality. Generally, when a feature becomes available in a newer version of Firefox, we tell its users to use that version to test the feature. But in this case we would be telling them to use two different versions simultaneously!
> 
> Even if we could give developers this option, that doesn't mean we should do it. I'd rather constrain the functionality we provide to a limited set of features that work well than to give developers as many options as possible.

But upon reviewing this patch I noticed that the current stable release of Firefox (18) exposes the Remote Debugger, and it's possible to connect to the Simulator using that tool. So the port is actually useful on the stable release (even if it's more useful on a newer version of Firefox).

Thus I'm ok with exposing the port on older versions of the browser. It isn't the ideal UX, but then neither is the current one. We'll keep evolving it until we get to the optimal state.

addon/data/content/js/main.js
```diff
@@ -71,11 +71,25 @@ var Simulator = {
...
+              remoteDebuggerPortEl.html(message.remoteDebuggerPort);
+              remoteDebuggerPortEl.parents('label').show();
+              // NOTE: show connect devtools buttons where it's supported
+              //       and show allocated debugger port on previous firefox releases
+              if (message.hasConnectDevtools) {
+                $("#show-debugger-port").hide();
+                $("#open-connect-devtools").show();
+              } else {
+                $("#show-debugger-port").show();
+                $("#open-connect-devtools").hide();
```
mykmelez, line 86: In my tests, clicking on the port triggers the button, which then opens a chrome: URL that doesn't exist in the older versions of Firefox for which we display the port. And since users will want to click the port number to select it (in order to copy/paste it), they are likely to experience this problem if they use the remote port at all.

That's peculiar, since hiding the button sets it to `display: none`, which should completely remove it from the DOM. But in any case disabling the button after hiding it seems to work around the problem:

```
               //       and show allocated debugger port on previous firefox releases
               if (message.hasConnectDevtools) {
                 $("#show-debugger-port").hide();
+                $("#open-connect-devtools").prop("disabled", false);
                 $("#open-connect-devtools").show();
               } else {
                 $("#show-debugger-port").show();
                 $("#open-connect-devtools").hide();
+                $("#open-connect-devtools").prop("disabled", true);
               }
             }
             else {
```

mykmelez:
I fixed the ellipsis nit in 2b72ba8920479f6b20ee8eb2c5255df7bf9b2883. I fixed the button issue in a846d589ca4bafbda54997c862e4c271a3737fe4.

PR #291, opened 2013-02-04
remove app using mozApps uninstall and reinstall on undo remove

rpl:
This change adds an unistallApp command to the custom simulatorActor, which use mozApps.mgmt.uninstall to uninstall app by origin.

The app will be removed immediately when the user click the remove link (as most of the users expect to happen) and will be reinstalled immediately if the user click the undo button.

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -121,15 +121,25 @@ let simulator = {
       }
       console.log("Stored " + JSON.stringify(apps[webappFile]));
 
-      this.updateApp(webappFile);
+      this.updateApp(webappFile, function next(error, appId) {
+        // app reinstall completed
+        // success/error detection and report to the user
+        if (error) {
+          simulator.error(error);
+        }
+        simulator.sendListApps();
+      });
     }
   },
 
   updateAll: function() {
     this.run(function () {
-      function next() {
+      function next(error, appId) {
         // Call iterator.next() in a timeout to ensure updateApp() has returned;
         // otherwise we might raise "TypeError: already executing generator".
+        if (error) {
+          simulator.error(error);
+        }
         Timer.setTimeout(function() {
           try {
             iterator.next();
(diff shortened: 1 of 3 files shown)


addon/lib/main.js
```diff
@@ -228,12 +238,18 @@ let simulator = {
...
       simulator.run(function() {
         simulator.remoteSimulator.install(config.xkey, null, function(res) {
-          console.debug("INSTALL RESPONSE: ",JSON.stringify(res));
+          console.debug("webappsActor install packaged app reply: ",
+                        JSON.stringify(res));
           if (next) {
-            next();
+            // detect success/error and report to the "next" callback
+            if (res.error) {
+              next(res.error + ": "+res.message, res.appId);
```
mykmelez, line 249: Nit: we should be consistent with space around the plus operator, so this should be either:

```
          next(res.error + ": " + res.message, res.appId);
```

Or:

```
          next(res.error+": "+res.message, res.appId);
```

Of which the former is more conventional in Firefox code.

addon/lib/main.js
```diff
@@ -228,12 +238,18 @@ let simulator = {
       // NOTE: remote simulator.defaultApp because on the first run the app
       //       will be not already installed
       simulator.defaultApp = null;
-      console.log("INSTALLING ",config.xkey);
+      console.log("Requesting webappsActor to install packaged app: ",config.xkey);
```
mykmelez, line 241: Nit: here and elsewhere, put a space after a comma.

addon/lib/main.js
```diff
@@ -121,15 +121,25 @@ let simulator = {
       }
       console.log("Stored " + JSON.stringify(apps[webappFile]));
 
-      this.updateApp(webappFile, true);
+      this.updateApp(webappFile, function next(error, appId) {
+        // app reinstall completed
+        // success/error detection and report to the user
+        if (error) {
+          simulator.error(error);
```
mykmelez, line 128: I'm concerned about the number of calls to `simulator.error()` we're adding, especially since they're all passing it a message they didn't create, since that function displays a message to the developer in the primary interface, and we should be very careful about such messagse (unlike messages we write to a secondary interface like a console), because they are disruptive.

We should audit all such calls and make sure we're displaying information that it is absolutely necessary to display and that makes sense to the developers who will be seeing them.

mykmelez:
Since all the issues were nits, I merged this without making any changes. But we should make sure to address these nits in future changes; and audit those error messages.

PR #302, opened 2013-02-09
zip packaged apps asynchronously

mykmelez:
We should do expensive operations asynchronously. Like the archiving of packaged app files. Here's a change that switches that archiving to nsIZipWriter's asynchronous API.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -233,27 +233,28 @@ let simulator = module.exports = {
       let archiveFile = File.join(tempWebappDir, "application.zip");
 
       console.log("Zipping " + sourceDir + " to " + archiveFile);
-      archiveDir(archiveFile, sourceDir);
+      archiveDir(archiveFile, sourceDir, function() {
+        simulator.info(config.name + " (packaged app) installed in Firefox OS");
+        // Complete install (Packaged)
 
-      simulator.info(config.name + " (packaged app) installed in Firefox OS");
-      // Complete install (Packaged)
-
-      // NOTE: remote simulator.defaultApp because on the first run the app
-      //       will be not already installed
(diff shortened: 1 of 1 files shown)


mykmelez:
@rpl: how does this look to you? It seems to work and is a relatively straightforward change. I'm not sure throwing the status is the best way to recover from an error, though.

addon/lib/simulator.js
```diff
@@ -858,7 +859,17 @@ function archiveDir(zipFile, dirToArchive) {
...
+  writer.processQueue({
+    onStartRequest: function onStartRequest(request, context) {},
+    onStopRequest: function onStopRequest(request, context, status) {
+      if (status == Cr.NS_OK) {
+        writer.close();
+        console.log("archived dir " + dirToArchive);
+        callback();
+      }
+      else {
+        throw status;
```
rpl: here we can use the same approach used in the updateApp callback and pass the error as an optional callback first parameter (and often used in the nodejs async code callbacks)

```
let error = ...
callback(error);
```

and let the callback to handle success/error cases (e.g. what/how to report the result to the user)

mykmelez: Nice pattern! I made this change.

mykmelez:
@rpl: how does this look with the change to error reporting?

rpl:
:+1: 
it looks good to me

mykmelez:
@rpl: Ok, thanks, I've merged it!

By the way, in the future, when I ask you to look at something like this, and you're satisfied with it, feel free to simply merge it yourself! In such a situation, you are serving as the reviewer of my code, so once you've looked at it and are satisfied with it, that means it's ready to go into the tree.

PR #303, opened 2013-02-09
fix "apps undeclared" error (missing let in local var declation)

rpl:
thanks to the new cleanup and "use strict" directive, we've found a silent problem:
apps wasn't declared in this assignments.


mykmelez:
Nice catch! I found one of these too, but I missed these two. :-)

PR #304, opened 2013-02-11
change runApp command to use mozApps API and app origin

rpl:
Change simulatorActor runApp command to use app origin to find and launch an installed app.
This change prevents the b2g-instance to exit if the application is not currently installed and doesn't launch the wrong app if you app share the same name.

 NOTE: runApp doesn't always show the app when the simulator instance
 is not already running, because the homescreen is not yet fully loaded and
 the app will start in background (reachable by pressing the home button to reach
 the gaia tasks list)

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -124,7 +124,7 @@ let simulator = module.exports = {
       }
       console.log("Stored " + JSON.stringify(apps[webappFile]));
 
-      this.updateApp(webappFile, function next(error, appId) {
+      this.updateApp(webappFile, function next(error, app) {
         // app reinstall completed
         // success/error detection and report to the user
         if (error) {
@@ -139,7 +139,7 @@ let simulator = module.exports = {
 
   updateAll: function() {
     this.run(function () {
-      function next(error, appId) {
+      function next(error, app) {
         // Call iterator.next() in a timeout to ensure updateApp() has returned;
         // otherwise we might raise "TypeError: already executing generator".
         if (error) {
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -129,8 +129,10 @@ let simulator = module.exports = {
         // success/error detection and report to the user
         if (error) {
           simulator.error(error);
+        } else {
+          simulator.sendListApps();
+          simulator.runApp(app.origin);
         }
-        simulator.sendListApps();
       });
     }
   },
@@ -255,15 +257,14 @@ let simulator = module.exports = {
               if (next) {
                 // detect success/error and report to the "next" callback
                 if (res.error) {
-                  next(res.error + ": " + res.message, res.appId);
+                  next(res.error + ": " + res.message, config);
                 } else {
-                  next(null, res.appId);
+                  next(null, config);
                 }
               }
             });
           });
         }
-
       });
     } else {
       // Hosted App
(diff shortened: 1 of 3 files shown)


addon/lib/simulator.js
```diff
@@ -124,13 +124,15 @@ let simulator = module.exports = {
       }
       console.log("Stored " + JSON.stringify(apps[webappFile]));
 
-      this.updateApp(webappFile, function next(error, appId) {
+      this.updateApp(webappFile, function next(error, app) {
```
mykmelez, line 127: This and the other _next_ functions have been all been made to take an _app_ object as their second argument, which seems fine. But _updateApp_ still passes those functions an app ID if there is an error. Granted, none of the _next_ functions do anything with their second argument if there is an error. But we should still be consistent about what we pass, so all callers pass the same thing, and the _next_ functions always know what to expect.

addon/lib/simulator.js
```diff
@@ -251,7 +253,7 @@ let simulator = module.exports = {
             if (res.error) {
               next(res.error + ": "+res.message, res.appId);
             } else {
-              next(null, res.appId);
+              next(null, config);
```
mykmelez: The change doesn't merge automatically, and I'm getting a strange conflict when I try to merge manually:

```
      archiveDir(archiveFile, sourceDir, function(error) {
        if (error) {
          if (next) {
<<<<<<< HEAD
            next(error);
=======
            // detect success/error and report to the "next" callback
            if (res.error) {
              next(res.error + ": "+res.message, res.appId);
            } else {
              next(null, config);
            }
>>>>>>> 514ebde8246e229a7d6f80ef4b7ec8f49aa31ca6
```

It isn't clear where the chunk from 514ebde8246e229a7d6f80ef4b7ec8f49aa31ca6 is coming from, but it does seem to be in the wrong place, as _archiveDir()_'s callback doesn't have a _res_ parameter, it has an _error_ parameter.

There is a code block like the one in the chunk a few lines down from there. But that one is indented twenty spaces rather than twelve. Nevertheless, it doesn't contain the necessary change, so I'm guessing that it is where the change should be made, and Git's merge algorithm somehow got it wrong (and changed the indent in the process?).

addon/lib/simulator.js
```diff
@@ -677,6 +684,13 @@ let simulator = module.exports = {
     }
   },
 
+  runApp: function(appOrigin, next) {
+    let cmd = (function () {
+      this.remoteSimulator.runApp(appOrigin, next);
+    }).bind(this);
+    this.run(cmd);
```
mykmelez: I don't mind binding functions. In fact, I rather like it. But we don't do it much in this module. Since _simulator_ is a singleton, we instead typically just reference it directly, i.e.:

```
simulator.remoteSimulator.runApp(appOrigin, next);
```

So I wonder if this bind is strictly necessary.

prosthesis/content/simulator-actors.js
```diff
@@ -61,19 +61,72 @@ SimulatorActor.prototype = {
   },
 
   onRunApp: function(aRequest) {
-    log("simulator actor received a 'runApp' command");
+    log("simulator actor received a 'runApp' command:"+aRequest.origin);
```
mykmelez: Nit: add whitespace around plus operator, i.e.:

```
log("simulator actor received a 'runApp' command:" + aRequest.origin);
```

rpl:
@mykmelez: how does this look with the suggested changes (and fix from commit 7f190d5)?

mykmelez:
Looks great, thanks @rpl!

Issue #6, opened 2012-09-19
Add a way to change orientation.


JoeStagner:
From Portrait to Landscape

hkgrild:
I agres with that suggestion, for dev's it would be awesome

leandroruel:
i agree too...i want it too

nhoizey:
It's a must have, of course.

abogdanovich:
+1 - we really need this option. Thanks guys!

rpl:
fixed by #305
(it will be released in the next simulator preview build)

Issue #270, opened 2013-01-26
No support for non-portrait orientations


mattbasta:
There's no way to test apps that don't run in portrait mode

mykmelez:
This is a duplicate of #6.

Issue #195, opened 2012-12-12
Add rotate button


waldyrious:
...possibly next to the home button. Spawned from #41.

mykmelez:
This is a more specific manifestation of issue #6, which requests a way to change orientation, so let's track it in that issue.

Also note that there are multiple ways we might implement the feature, and the issue tracks implementation generally, not the specifics of how we do it. Nevertheless, in recent conversations with @rpl, I suggested this very solution: a button on the toolbar that rotates the Simulator window.

Issue #123, opened 2012-11-18
Simulator not working properly on lubuntu 12.10 Firefox 19.0a1 (2012-11-07)


thomasysliu:
The Firefox nighty is from PPA

deb http://ppa.launchpad.net/ubuntu-mozilla-daily/ppa/ubuntu precise main
deb-src http://ppa.launchpad.net/ubuntu-mozilla-daily/ppa/ubuntu precise main

The simulator just not start any new window. And the switch turn off soon after it start.

thomasysliu:
It also happened on Aurora

The error message is 
/////
Running global cleanup code from study base classes.
info: r2d2b2g: Simulator.onMessage getIsRunning
info: r2d2b2g: Simulator.onMessage listApps
info: r2d2b2g: Simulator.sendListApps
info: r2d2b2g: Simulator.onMessage listTabs
info: r2d2b2g: Simulator.onMessage getPreference
info: r2d2b2g: Simulator.onMessage toggle
###!!! ABORT: X_GLXQueryServerString: BadRequest (invalid request code or no such operation): file /home/myk/Mozilla/central/toolkit/xre/nsX11ErrorHandler.cpp, line 157
###!!! ABORT: X_GLXQueryServerString: BadRequest (invalid request code or no such operation): file /home/myk/Mozilla/central/toolkit/xre/nsX11ErrorHandler.cpp, line 157
info: r2d2b2g: linux/b2g/b2g-bin terminated with 139

/////////////////

info: r2d2b2g: Simulator.onMessage toggle
###!!! ABORT: X_GLXQueryServerString: BadRequest (invalid request code or no such operation): file /home/myk/Mozilla/central/toolkit/xre/nsX11ErrorHandler.cpp, line 157
###!!! ABORT: X_GLXQueryServerString: BadRequest (invalid request code or no such operation): file /home/myk/Mozilla/central/toolkit/xre/nsX11ErrorHandler.cpp, line 157
info: r2d2b2g: linux/b2g/b2g-bin terminated with 139
Failed to open VDPAU backend libvdpau_nvidia.so: 無法開啟共用目的檔: 沒有此一檔案或目錄

thomasysliu:
Crash gdb information

liu@thomas-laptop:~/.mozilla/firefox-trunk/ovqpozp5.default/extensions/r2d2b2g@mozilla.org/resources/r2d2b2g/data/linux/b2g$ ./b2g 
[...]

PR #305, opened 2013-02-11
FEATURE: simulate screen orientation lock/unlock and phone resolution

rpl:
fix issue #6 and #7  
- simulate screen orientation, lock/unlock and dispatch mozorientationchange event
- flip screen button (enabled/disable on screen orientation lock changes)

DEMO:
- http://youtu.be/iw_s4ppjiLY


past:
The demo looks very cool indeed!
I was wondering if you have seen this as well, and whether it is something we could incorporate into the simulator:
http://louisremi.github.com/orientation-devtool/

wfwalker:
AWESOME, thank you so much for your hard work so far!

rpl:
@past: I tried to start from something simpler like the screen orientation api: 
http://www.w3.org/TR/screen-orientation/
http://people.mozilla.com/~gmealer/exercises/orientation/

and "flip screen" feature is highly requested (#6, #123, #195, #270) and currently b2g-desktop nsIDOMScreen reports wrong resolution and orientation (desktop width/height and landscape-primary orientation).

@wfwalker: thanks! it started as a test but I think it's promising enough to be a good starting point to a more complete and correct implementation

past:
Agreed, simple is how we should start. Filed issue #312 for improving this further.

mykmelez:
Is this now ready to review?

rpl:
I want to rework prosthesis/components/SimulatorScreen.js to be modeled on DOMFMRadioChild 
(http://mxr.mozilla.org/mozilla-central/source/dom/fm/DOMFMRadioChild.js)
and try to remove (or at least reduce) the prosthesis/modules/GlobalSimulatorScreen.jsm

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,176 @@
...
+    flags: Ci.nsIClassInfo.DOM_OBJECT
+  }),
+
+  init: function (aWindow) {
+    dump("SIMULATOR SCREEN INIT CALLED\n");
+    //this._aWindow = Cu.getWeakReference(aWindow);
+    this._aWindow = XPCNativeWrapper.unwrap(aWindow);
+    this._winUtils = this._aWindow.
+      QueryInterface(Components.interfaces.nsIInterfaceRequestor)
+      .getInterface(Components.interfaces.nsIDOMWindowUtils);
```
mykmelez: Nit: Components.interfaces -> Ci
Nit: put dereference operator (i.e. period) at end of previous line.

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,176 @@
...
+
+    let self = this;
+    self._globalScreen = GlobalSimulatorScreen;
+    dump("SCREEN: "+self._globalScreen.mozOrientation+"\n");
+    self._onmozorientationchange = null;
+
+    let chromeObject = {
```
mykmelez: All of the functions defined in _chromeObject_, whether getters/setters or methods, close over the scope of the _init()_ function, which means they have access to both the _GlobalSimulatorScreen_ global and the _aWindow_ argument. So you don't need to assign those values to properties of SimulatorScreen, and you don't need to assign _this_ to _self_. Just use _GlobalSimulatorScreen_ and _aWindow_ directly in all the functions.

As for __onmozorientationchange_, it is only ever defined and used by _chromeObject_, so it can be simply a property of that object. But note that you specify it as "rw", yet you don't include a getter for it, only a setter.

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,176 @@
...
+      get availLeft() {
+        return 0;
+      },
+      get mozOrientation() {
+        return self._globalScreen.mozOrientation;
+      },
+      addEventListener: self._aWindow.addEventListener,
+      removeEventListener: self._aWindow.removeEventListener,
+      set onmozorientationchange(value) {
+        if(self._onmozorientationchange) {
```
mykmelez: Nit: add a space between "if" and the opening parenthesis.

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,176 @@
...
+            evt.initCustomEvent('mozorientationchange', true, false, {
+              orientation: orientation
+            });
+            //self._winUtils.dispatchDOMEventViaPresShell(window, evt, true);
+            self._aWindow.dispatchEvent(evt);
+          }
+
+          return true;
+        }
+        dump("orientation not found: '"+orientation+"'\n");
```
mykmelez: Nit: add space around plus signs.

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,176 @@
...
+    const properties = {
+      top: genPropDesc('top'),
+      left: genPropDesc('left'),
+      width: genPropDesc('width'),
+      height: genPropDesc('height'),
+    };
+    Object.defineProperties(contentObj, properties);
+    Cu.makeObjectPropsNormal(contentObj);
+
+    return contentObj;
```
mykmelez: Presumably this is all no longer necessary, since you return _chromeObject_ after defining it.

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+  get rotateButton() {
+    if (rotateButtonEl) {
+      return rotateButtonEl;
+    }
+
+    var window = XPCNativeWrapper.unwrap(
+        Cc['@mozilla.org/appshell/window-mediator;1']
+          .getService(Ci.nsIWindowMediator)
+          .getMostRecentWindow("navigator:browser")
+      );
```
mykmelez: Nit: it'd be preferable to use `let` in place of `var` in this file.

Also, it looks like the second and subsequent lines of this statement are indented two spaces too much.

Finally, nit: put the dereference operators (i.e. periods) on the ends of the previous lines.

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+    GlobalSimulatorScreen.rotateButton.classList.remove("active");
+  },
+  unlocked: function() {
+    GlobalSimulatorScreen.mozOrientationLocked = false;
+    GlobalSimulatorScreen.rotateButton.classList.add("active");
+  },
+  flipScreen: function() {
+    if (GlobalSimulatorScreen.mozOrientationLocked) {
+      // disabled
+    } else {
```
mykmelez: Nit: in functions like this, where the function exits early if some condition is true, it's simpler to make the function explicitly return, since then the rest of the function doesn't have to live in an _else_ block, i.e.:

```
flipScreen: function() {
  if (GlobalSimulatorScreen.mozOrientationLocked) {
    return;
  }

  ...
```

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+      GlobalSimulatorScreen.width = 320;
+      GlobalSimulatorScreen.height = 480;
+    } else if (GlobalSimulatorScreen.mozOrientation.match(/^landscape/)) {
+      GlobalSimulatorScreen.width = 480;
+      GlobalSimulatorScreen.height = 320;
+    }
+
+    var width = GlobalSimulatorScreen.width+"px";
+    var height = GlobalSimulatorScreen.height+"px";
+    dump("ROTATE: "+width+" "+height+"\n");
```
mykmelez: Nit: add space around plus signs here and elsewhere.

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+    });
+    ["height", "min-height", "max-height"].forEach(function (i) {
+      shell.style[i] = height;
+      homescreen.style[i] = height;
+    });
+    // WORKAROUND: run window.resizeTo immediately will not resize it correctly
+    window.setTimeout(function () {
+      dump("RESIZE TO: "+width+" "+height+"\n");
+      window.resizeTo(parseInt(width), parseInt(height));
+    },100);
```
mykmelez: Is it necessary to wait at least 100ms, or is to just necessary to wait any amount of time (i.e. to spin the event loop)? If the latter, `setTimeout(…, 0)` (i.e. zero), which will spin the event loop without waiting any longer than necessary and is the convention in Firefox code when we have to do something after spinning the event loop.

If we do have to wait at least 100ms, however, then that's a problem, because we're probably going to have to wait longer on some machine that is slower than the ones we're testing on. In that case, we should figure out what it is we're actually waiting on and make sure that finishes before we resize the window.

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+  },
+
+}
\ No newline at end of file
```
mykmelez: Nit: add newline to end of file.

mykmelez:
Where does rotate.png come from, and what is the license under which it is made available? I just want to make sure we have the legal ability to use it!

rpl:
I created rotate.png in gimp as a test
I think the original image is: http://www.defaulticon.com/v1/
which unfortunately is "Creative Commons Attribution-No Derivative Works 3.0"
:-(

mykmelez:
That's an ok license if we credit the author and use the original image, although that "no derivative" part means we can't change it. Let's do whatever is easiest, since these images are temporary, and we'll replace them with custom ones created by a Mozilla visual designer at some point.

mykmelez:
Also see this ["rotate iPad" icon](http://thenounproject.com/noun/rotate-ipad/#icon-No11649) from the noun project, whose CC license does permit modification.

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,154 @@
...
+    classDescription: "mozSimulatorScreen",
+    interfaces: [Ci.nsIDOMGlobalPropertyInitializer,
+                 Ci.nsISupportsWeakReference],
+    flags: Ci.nsIClassInfo.DOM_OBJECT
+  }),
+
+  init: function (aWindow) {
+    dump("SIMULATOR SCREEN INIT CALLED\n");
+
+    let aWindow = XPCNativeWrapper.unwrap(aWindow);
```
mykmelez: Nit: this _let_ declaration redeclares the _aWindow_ argument. The statement should reassign _aWindow_ without redeclaring it or give the variable a different name.

prosthesis/components/SimulatorScreen.js
```diff
@@ -0,0 +1,154 @@
...
+      },
+      get availTop() {
+        return 0;
+      },
+      get availLeft() {
+        return 0;
+      },
+      get mozOrientation() {
+        return globalScreen.mozOrientation;
+      },
```
mykmelez: Nit: these simple getters be easier to read as one-line lambdas, f.e.:

```
get mozOrientation() globalScreen.mozOrientation,
```

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+    let window = GlobalSimulatorScreen.window;
+    this._rotateButtonEl = window.document.getElementById("rotateButton");
+
+    return this._rotateButtonEl;
+  },
+  locked: function() {
+    GlobalSimulatorScreen.mozOrientationLocked = true;
+    GlobalSimulatorScreen.rotateButton.classList.remove("active");
+  },
+  unlocked: function() {
```
mykmelez: Nit: these would be better named as the verbs _lock()_ and _unlock()_.

prosthesis/modules/GlobalSimulatorScreen.jsm
```diff
@@ -0,0 +1,114 @@
...
+
+    this._window = XPCNativeWrapper.unwrap(
+      Cc['@mozilla.org/appshell/window-mediator;1'].
+        getService(Ci.nsIWindowMediator).
+        getMostRecentWindow("navigator:browser")
+    );
+
+    return this._window;
+  },
+  get rotateButton() {
```
mykmelez, line 30: Nit: it'd be a bit easier to read this code if there was a line of whitespace between the multi-line property declarations, like these two for the _window_ and _rotateButton_ getters.

mykmelez:
This is awesome, and I only have nits, so this is ready to merge! Feel free to do so after fixing zero or more of the nits at your discretion.

And a thought for a future direction: it might be nice if the orientation change animated, so it wasn't so abrupt; and nifty if it changed in a way that left the rotate button in the same position under the mouse, so a developer could click multiple times to change the orientation and then return to the original orientation without having to move their mouse.

rpl:
Merged! 
(Tested for regression using last updated b2g-desktop and gaia components)

PR #314, opened 2013-02-13
FIX: application update command should update packaged app manifest and content 

### LINKED ISSUE #269: Update function on Dashboard cannot work to refresh the app. on local file storage
I cannot use [Update] button on the Dashboard to refresh the app after I have made some modifications to the implementation.
# STR
1. Load an app from local file system.
2. Make some modifications to the HTML files of the app
3. Press [Update] button - it will show an banner on simulator that the app is installed
4. relaunch the app. to check

Expected:
   The app has been updated.

Actual:
  It still shows the old state of the app.
## 

Note:
After deleting the offline cache in the Gaia profile, it could work to refresh the app.
However, I think it would be better to resolve this issue for developer to modify and try their app.

Thanks.

### LINKED ISSUE #308: updating all apps on addon update causes Simulator to start unexpectedly
Updating an app causes the Simulator to start if it isn't already running. That's ok when the user triggers the update by pressing the app's Update button, since it's clear that the Simulator started as a result of the button press, and the user probably wants to test the app after updating it.

But it's a problem when an addon update triggers updating all apps, because it isn't clear why the Simulator started in that case, especially if the Dashboard is closed. Launching the Simulator then is unexpected and startles the user.

We have to start the Simulator to update an app. But that is a technical implementation detail. And we shouldn't expose the user to it. The Simulator should only start when the user asks it to, either because they press the Start button or because they press the Update button for a particular app.

We could fix this by starting the Simulator in a special "install app" mode when updating all apps while the Simulator is not running. The mode, triggered by a command-line flag, would install the app specified on the command line (or multiple apps specified in a temporary file) and then quit. All without ever opening a window. So the user will never see the Simulator.

cc: @rpl

rpl:
- this change should fix #269
- introduce a temporary webappsActor renamed to simulatorwebappsActor to fix a couple of issues with reinstall application using its install command
- kill the app if it's already running to run the new version correctly
- install app if not installed on run button press 
- reinstall apps on first run after upgrade/downgrade/remote+install (fix #308)


addon/lib/remote-simulator-client.js
```diff
@@ -76,7 +76,7 @@ const RemoteSimulatorClient = Class({
           tabs: reply.tabs,
           selected: reply.selected,
           simulator: reply.simulatorActor,
-          webapps: reply.webappsActor
+          webapps: reply.simulatorwebappsActor
```
rpl: use the temporary fixed webappsActor renamed to simulatorwebappsActor

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -0,0 +1,281 @@
...
+function WebappsActor(aConnection) { debug("init"); }
+
+WebappsActor.prototype = {
+  actorPrefix: "simulatorwebapps",
+
+  _registerApp: function wa_actorRegisterApp(aApp, aId, aDir) {
+    let reg = DOMApplicationRegistry;
+    let self = this;
+
+    if (aId in reg._manifestCache) {
```
rpl, line 36: clean the cached manifest (e.g. delete DOMApplicationRegistry._manifestCache[appId]) as done here: http://mxr.mozilla.org/mozilla-central/source/dom/apps/src/Webapps.jsm#1084

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -0,0 +1,281 @@
...
+          zipFile.moveTo(installDir, "application.zip");
+
+
+
+          // Extract the manifest.webapp file from the zip.
+          zipFile = installDir.clone();
+          zipFile.append("application.zip");
+
+          // Refresh applicazion.zip content (e.g. reinstall app)
+          Services.obs.notifyObservers(zipFile, "flush-cache-entry", null);
```
rpl, line 183: flush zip reader cache  as done here: http://mxr.mozilla.org/mozilla-central/source/dom/apps/src/Webapps.jsm#1117

prosthesis/content/simulator-actors.js
```diff
@@ -63,43 +63,62 @@ SimulatorActor.prototype = {
...
         } catch(e) {
-          log("EXCEPTION: " + e);
+          log("RUNAPP EXCEPTION: " + e);
         }
       },
       unlockScreen: function(cb) {
-        let setReq = window.navigator.mozSettings
-          .createLock().set({'lockscreen.enabled': false});
-        setReq.onsuccess = function() {
+        // WORKAROUND: currently we're not able to detect when the firefoxos is fully loaded
```
rpl: I would like a better workaround to this problem :-(

prosthesis/content/simulator-actors.js
```diff
@@ -63,43 +63,62 @@ SimulatorActor.prototype = {
...
+          };
+          setReq.onerror = function() {
+            cb("unlock error");
+          };
+        }, 500);
+      },
+      killAppByOrigin: function(origin, cb) {
+        try {
+          WindowManager.kill(origin);
+          // WORKAROUND: currently WindowManager.kill doesn't always call
```
rpl: I would like a better workaround to this problem :-( :-( :-(

addon/lib/remote-simulator-client.js
mykmelez: Nit: for names like this, capitalize the previously-initial character of the old name (i.e. simulatorWebappsActor) or separate the new prefix from the old name with an underscore character (i.e. simulator_webappsActor).

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -0,0 +1,281 @@
...
+          zipFile.append("application.zip");
+          zipFile.moveTo(installDir, "application.zip");
+
+
+
+          // Extract the manifest.webapp file from the zip.
+          zipFile = installDir.clone();
+          zipFile.append("application.zip");
+
+          // Refresh applicazion.zip content (e.g. reinstall app)
```
mykmelez: Nit: applicazion -> application

prosthesis/content/simulator-actors.js
mykmelez: Check `(WindowManager.getRunningApps().hasOwnProperty[origin] || !WindowManager.getRunningApps()[origin].killed)` before killing the app, then listen for an _appterminated_ event instead of passing a callback.

_kill()_ only returns early if `!isRunning(origin)` or `runningApps[origin].killed`. And _isRunning()_ just checks `runningApps.hasOwnProperty(origin)`. Where _runningApps_ is the object returned by _getRunningApps()_. It still doesn't always call the callback, even if it doesn't return early. But it does always dispatch the event if it doesn't return early.

prosthesis/content/simulator-actors.js
mykmelez: Can we disable the lockscreen permanently, f.e. via build/override-settings.json? It isn't useful for app developers right now, it's just a hoop to jump through. And in the rare case where a developer really needs to see the lockscreen, presumably they can do so by pressing the End key (Fn+RightArrow on Mac).

rpl:
renamed prosthesis/content/simulator-actors.js to prosthesis/content/dbg-simulator-actors.js
(follow debugging actors definition files naming conventions)

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -63,43 +63,93 @@ SimulatorActor.prototype = {
...
+                }
+              });
             });
           });
         } catch(e) {
-          log("EXCEPTION: " + e);
+          log("RUNAPP EXCEPTION: " + e);
+        }
+      },
+      waitHomescreenReady: function(cb) {
```
rpl, line 102: wait for "homescreen.ready" setting to be true:
- b2g instance already started
- b2g instance starting, lockscreen is fully loaded and correctly unlocked

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -63,43 +63,93 @@ SimulatorActor.prototype = {
...
+            let app = WindowManager.getRunningApps()[origin];
+            log("RUNAPP: killAppByOrigin - wait for appterminated");
+            let once = function (evt) {
+              if (evt.detail.origin !== origin)
+                return
+              log("RUNAPP: killAppByOrigin - appterminated received");
+              homescreen.removeEventListener("appterminated", once);
+              // WORKAROUND: bug (probably related to disabled oop) restarted
+              // app keep to be flagged as killed and never restarted
+              delete WindowManager.getRunningApps()[origin];
```
rpl, line 143: I'm not sure why happening, but after a couple of times a user click the "update" app button,
window_manager keeps to terminated app as killed in the running apps list and never restarts
it again:
- !!WindowManager.getRunningApps()[origin] == true
- WindowManager.getRunningApps()[origin].killed == true

prosthesis/content/lockscreen.js
```diff
@@ -0,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// NOTE: disable lockscreen by default on FirefoxOS Simulator
+// disable lockscreen on startup
+SettingsListener.observe("lockscreen.enabled", false, function(value) {
```
rpl, line 9: lockscreen.enabled is set and never read by LockScreen,
we use it to detect when lockscreen start loading.

prosthesis/content/lockscreen.js
```diff
@@ -0,0 +1,39 @@
...
+  if (!value)
+    return
+
+  navigator.mozSettings
+    .createLock().set({'homescreen.ready': false});
+
+  let homescreen = document.getElementById("homescreen").contentWindow.wrappedJSObject;
+  dump("simulator - HOMESCREEN LOCKED: wait for lockscreen.locked \n");
+  // keep the lockscreen unlocked
+  SettingsListener.observe("lockscreen.locked", true, function(value) {
```
rpl, line 21: lockscreen.locked is set by LockScreen on lock/unlock

this handler keep the lockscreen unlocked and set "homescreen.ready = true"

addon/lib/simulator.js
```diff
@@ -764,13 +764,26 @@ let simulator = module.exports = {
...
         });
         break;
       case "runApp":
-        let appOrigin = this.apps[message.id].origin;
-        simulator.runApp(appOrigin);
+        let app = this.apps[message.id];
+        simulator.runApp(app, function (res) {
+          if (res.success === false) {
+            if (res.error === 'app-not-installed') {
+              // install and run if not installed
```
rpl, line 801: we're now able to detect if an app is not installed, so we can try to install and run it again.
thanks to this change we can lazily reinstalled an app when:
- user remove the app manually from the homescreen
- user upgrade or reinstall the simulator addon

reinstalling the addon xpi will overwrite the b2g profile and all the app registered in the dashboard
will became "not installed".

addon/lib/main.js
```diff
@@ -76,8 +78,8 @@ switch (Self.loadReason) {
     if (Services.vc.compare(lastVersion, "2.0pre5") < 0) {
       ensureXkeysUnique();
     }
-
-    Simulator.updateAll();
+    // NOTE: updateAll disabled (lazy install if not installed on runApp)
+    //Simulator.updateAll();
```
rpl: thanks to the "install if not installed" feature in the simulator runApp method,
we can now alleviate #308 ensuring xkey uniqueness on upgrade and leave
apps to lazily reinstall when the user press the run button.

mykmelez:
Oh, and thanks for removing the trailing spaces and unused code! In the long run, it'll make a difference to keep our code tidy in these and other ways.

addon/lib/main.js
```diff
@@ -62,23 +62,26 @@ if (SStorage.storage.lastVersion != Self.version) {
   SStorage.storage.lastVersion = Self.version;
 }
 
+// on remove+install, downgrade and upgrade:
+// - ensure apps xkeys are unique
+// - flag needsUpdateAll if there are active apps registered
+if (["install", "downgrade", "upgrade"].indexOf(Self.loadReason) >= 0) {
```
rpl, line 68: on remove+install, downgrade and upgrade we just ensure apps xkeys are unique and flag needsUpdateAll if there are active apps registered
("deleted" apps should not be reinstalled)

addon/lib/simulator.js
```diff
@@ -131,13 +131,14 @@ let simulator = module.exports = {
...
-          simulator.runApp(app.origin);
+          simulator.runApp(app);
         }
       });
     }
   },
 
-  updateAll: function() {
+  updateAll: function(oncompleted) {
+    simulator.remoteShowNotification("reinstalling registered apps...");
```
rpl: add notification on b2g side to keep the user informed of reinstall apps procedure

addon/lib/simulator.js
```diff
@@ -148,14 +149,29 @@ let simulator = module.exports = {
         Timer.setTimeout(function() {
           try {
             iterator.next();
-          } catch (err if err instanceof StopIteration) {}
+          } catch (err if err instanceof StopIteration) {
+            simulator.remoteShowNotification("reinstalling completed.");
```
rpl: add another notification on b2g side to signal completed reinstall apps procedure

addon/lib/simulator.js
```diff
@@ -148,14 +149,29 @@ let simulator = module.exports = {
...
+            if (typeof oncompleted === "function") {
+              oncompleted();
+            }
+          }
         }, 0);
       }
-      let iterator = (simulator.updateApp(id, next) for (id in simulator.apps));
+      // only active apps needs to be reinstalled
```
rpl, line 160: only active apps needs to be reinstalled

addon/lib/simulator.js
```diff
@@ -677,14 +693,27 @@ let simulator = module.exports = {
       this.defaultApp = null;
     }
 
+    let next = null;
+    // if needsUpdateAll try to reinstall all active registered app
+    if (SStorage.storage.needsUpdateAll) {
```
rpl, line 698: if needsUpdateAll try to reinstall all active registered app

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -154,7 +220,19 @@ SimulatorActor.prototype = {
     Services.tm.currentThread.dispatch(runnable,
                                        Ci.nsIThread.DISPATCH_NORMAL);
     return {
-      message: "uninstallApp request received"
+      message: "uninstallApp request received",
+      success: true
+    };
+  },
+
+  onShowNotification: function (aRequest) {
```
rpl, line 228: use AlertHelper to show notification on the b2g side

addon/lib/simulator.js
```diff
@@ -148,14 +149,29 @@ let simulator = module.exports = {
...
+        .filter(function (appId) !simulator.apps[appId].deleted);
+      let iterator = (simulator.updateApp(activeAppIds[i], next) for (i in activeAppIds));
       next();
     });
   },
 
+  showRemoteNotification: function(userMessage) {
+    this.run(function () {
+      simulator.remoteSimulator
+        .showNotification(userMessage, null);
```
mykmelez, line 171: Nit: since the _onResponse_ argument is optional, just leave it out instead of passing _null_.

mykmelez:
Merged with nit fixed!

PR #316, opened 2013-02-16
more info on how to contribute

nickdesaulniers:
review? @mykmelez

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+# Running a Fork #
+
+1. Fork on github: https://github.com/mozilla/r2d2b2g
+2. clone your fork: `git clone git@github.com:your_github_username_here/r2d2b2g.git`
+3. set r2d2b2g/master as upstream: `git remote add --track master upstream https://github.com/mozilla/r2d2b2g.git`
+4. Make a new directory for your firefox profile outside of the cloned fork. ex. `mkdir arbitraryProfileName`
+5. cd into your cloned fork
+6. initialize submodules `git submodule init && git submodule update`
+7. make `make`
+8. initialize addon sdk `cd addon-sdk && . bin/activate`
+9. change into addon dir `cd ../addon`
+10. Run it! `cfx run --templatedir template/ --profiledir ../../arbitraryProfileName/`
+
+Note: to close it use commmand quit in osx, control-c from the command line will not save any newly installed apps.
+
+# Contributing #
+
+Pick up an issue from the [issue tracker](https://github.com/mozilla/r2d2b2g/issues?state=open) and checkout a new branch that describes the issue.
+ex. `git checkout -b fixThatBrokenThingYouKnow`
+
+squashed commits preferred, then issue a pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,13 +3,12 @@
 1. Fork on github: https://github.com/mozilla/r2d2b2g
 2. clone your fork: `git clone git@github.com:your_github_username_here/r2d2b2g.git`
 3. set r2d2b2g/master as upstream: `git remote add --track master upstream https://github.com/mozilla/r2d2b2g.git`
-4. Make a new directory for your firefox profile outside of the cloned fork. ex. `mkdir arbitraryProfileName`
-5. cd into your cloned fork
-6. initialize submodules `git submodule init && git submodule update`
-7. make `make`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
+# Running a Fork #
+
+1. Fork on github: https://github.com/mozilla/r2d2b2g
+2. clone your fork: `git clone git@github.com:your_github_username_here/r2d2b2g.git`
+3. set r2d2b2g/master as upstream: `git remote add --track master upstream https://github.com/mozilla/r2d2b2g.git`
+4. Make a new directory for your firefox profile outside of the cloned fork. ex. `mkdir arbitraryProfileName`
```
mykmelez: It doesn't have to be outside of your cloned fork, although if you put it inside, then Git will complain about it being an untracked dir until you add it to your global _.gitignore_. But we probably don't need to go into this level of detail.

However, I would reference _/path/to/arbitraryProfileName_ to make it clearer that you're creating a directory on some path outside the cloned fork.

CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
...
+5. cd into your cloned fork
+6. initialize submodules `git submodule init && git submodule update`
+7. make `make`
+8. initialize addon sdk `cd addon-sdk && . bin/activate`
+9. change into addon dir `cd ../addon`
+10. Run it! `cfx run --templatedir template/ --profiledir ../../arbitraryProfileName/`
```
mykmelez: The point of `make run` is to abstract away the complexity of `cfx`, so I'd prefer to simply tell people to use `make run` here, since these look like instructions for getting up and running quickly.

But let's add a _Using cfx_ section to the end of the document that explains how to activate _cfx_ for folks who want more control over their development environment.

And we should also make `make run` reuse a profile directory by default, creating it if it doesn't already exist, something like a top-level _profile_ directory that we add to _.gitignore_ and delete on `make clean`. I filed issue #317 on that.

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
...
+3. set r2d2b2g/master as upstream: `git remote add --track master upstream https://github.com/mozilla/r2d2b2g.git`
+4. cd into your cloned fork
+5. initialize submodules `git submodule init && git submodule update`
+6. make `make`
+7. initialize addon sdk `cd addon-sdk && . bin/activate`
+8. change into addon dir `cd ../addon`
+9. Run it! `cfx run --templatedir template/ --profiledir /path/to/arbitrary/profile/`
+
+Note: to close it use commmand quit in osx, control-c from the command line will not save any newly installed apps.
+
```
mykmelez, line 14: To make this a little clearer, write it as:

Note: to quit the instance of Firefox started by `make run`, use Firefox's Quit/Exit menu item, as aborting the task on the command line with _Control-C_ will crash Firefox, and you'll lose changes to the apps on the Dashboard.

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
...
+# Contributing #
+
```
mykmelez, line 16: The whole document is called Contributing, so it's a little strange to have a section within it called Contributing. Perhaps this could be called _Fixing Issues_.

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
...
+Pick up an issue from the [issue tracker](https://github.com/mozilla/r2d2b2g/issues?state=open) and checkout a new branch that describes the issue.
+ex. `git checkout -b fixThatBrokenThingYouKnow`
+
+squashed commits preferred, then issue a pull request.
```
mykmelez, line 20: Nit: the formatting here is a little weird.

Also, here and elsewhere, wrap lines at 80 characters.

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
+# Running a Fork #
+
+1. Fork on github: https://github.com/mozilla/r2d2b2g
+2. clone your fork: `git clone git@github.com:your_github_username_here/r2d2b2g.git`
+3. set r2d2b2g/master as upstream: `git remote add --track master upstream https://github.com/mozilla/r2d2b2g.git`
+4. cd into your cloned fork
```
mykmelez, line 6: Steps 3 and 4 need to be switched, as you need to change into the working directory before you add the remote.

Also, nit: most of the steps show a command, but the `cd` step doesn't. It should show: `cd r2d2b2g/`.

And nit: some of these steps use a colon to separate the description from the command, while others don't; they should be consistent.

PR #318, opened 2013-02-16
FEATURE: optional reuse profile in make run/test

### LINKED ISSUE #317: make `make run` reuse profile dir
It's painful to develop with `make run` because the Firefox profile gets recreated each time. So I and other devs like @nickdesaulniers and @digitarald (not sure about @rpl) typically use `cfx` directly. So we should make `make run` reuse a profile directory by default, creating it if it doesn't already exist, something like a top-level `profile` directory that we add to .gitignore and delete on `make clean`.

rpl:
optional reuse profile using make run/test (fix #317):

```
 make run PROFILE=/path/to/profile
```

or

```
 make test PROFILE=/path/to/profile
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,10 @@ ifdef BIN
   BIN_ARG = -b $(BIN)
 endif
 
+ifdef PROFILE
+  PROFILE_ARG = --profiledir $(PROFILE)
+endif
+
 ifdef TEST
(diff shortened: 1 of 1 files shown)


mykmelez:
Looks good, although I still think we should reuse a profile dir by default.

Issue #277, opened 2013-01-27
Simulator 2.0pre5: geolocation dialog throws "frame is null" error


jrburke:
Using the 2.0pre5 build at firefox hack day. It seems like maybe the build of gaia in that versoin has a bug with the geolocation prompt: user clicks Accept, but no geolocation callback occurs, and there is a "frame is null" error in the web console.

I expect it will be sorted out as gaia is updated, but just noting it, happy to retest when a new build is out.

mykmelez:
The geolocation prompt seemed to work in earlier versions of Simulator, although only on some OSes. So this seems to have broken since then. Hopefully it'll get fixed in a newer version of Gaia, but if it's B2G Desktop-specific, then the Gaia developers probably aren't paying close attention to it, and we may have to figure it out ourselves.

mykmelez:
@kanru has contributed a fake geolocation provider, which we can use to make geolocation work on all platforms. I added it to the Simulator in this branch:

https://github.com/mykmelez/r2d2b2g/tree/geolocation
https://github.com/mykmelez/r2d2b2g/blob/geolocation/prosthesis/components/FakeGeolocationProvider.js

However, it still needs some work. As I [discussed with @kanru on dev.webapps](https://groups.google.com/forum/#!topic/mozilla.dev.webapps/phGFqhXrP6A), we'll want an interface for developers to specify the location (with the default being the current location, as provided by Firefox). And some parameters for movement, although we could do that in a second phase of development.

PR #325, opened 2013-02-22
Geolocation

nickdesaulniers:
review? @mykmelez
Issue #277

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -5,6 +5,8 @@
 
 'use strict';
 
+const { Cc, Ci, Cu, ChromeWorker } = require("chrome");
+
 const { EventTarget } = require("sdk/event/target");
 const { emit, off } = require("sdk/event/core");
 const { Class } = require("sdk/core/heritage");
@@ -15,10 +17,8 @@ const URL = require("url");
 const Subprocess = require("subprocess");
 const Prefs = require("preferences-service");
 
-const { Cc, Ci, Cu, ChromeWorker } = require("chrome");
-
-const { rootURI } = require('@loader/options');
-const profileURL = rootURI + "profile/";
+const { rootURI: ROOT_URI } = require('@loader/options');
+const PROFILE_URL = ROOT_URI + "profile/";
 
 const PingbackServer = require("pingback-server");
 
@@ -305,7 +305,7 @@ const RemoteSimulatorClient = Class({
 
   // compute current b2g filename
   get b2gFilename() {
-    if (this._executable) return this._executableFilename;
+    return this._executable ? this._executableFilename : "B2G";
   },
 
   // compute current b2g file handle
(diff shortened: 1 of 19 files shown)


addon/lib/remote-simulator-client.js
```diff
@@ -199,6 +203,25 @@ const RemoteSimulatorClient = Class({
       emit(this, "clientClosed", {client: client});
     }).bind(this));
 
+    client.addListener("geolocationRequest", (function() {
+      let geolocation = Cc["@mozilla.org/geolocation;1"].
+                        getService(Ci.nsIDOMGeoGeolocation);
+      geolocation.getCurrentPosition((function success(position) {
+        let [lat, lon] = [position.coords.latitude, position.coords.longitude];
```
mykmelez: Note that you can also do:

```
let { latitude: lat, longitude: lon } = position.coords;
```

Or make it even simpler by reusing the existing property names:

```
let { latitude, longitude } = position.coords;
...
  message: {
    lat: latitude,
    lon: longitude,
  },
```

But the current approach is also great.

gaia
```diff
@@ -1 +1 @@
-Subproject commit d082aa0f85c0dbad6faa7da8235357f1a9d18255
+Subproject commit 9f5a4a9f6d321dbb4805f1f98623e770b0e89698
```
mykmelez: This looks like an extraneous change from a bad merge.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -24,6 +26,17 @@ function SimulatorActor(aConnection)
...
+
+  let self = this;
+  Services.obs.addObserver(function () {
+    if (self.clientReady) {
+      self._connection.send({
+        from: self.actorID,
+        type: "geolocationRequest"
+      });
+    }
+  }, "r2d2b2g-geolocation-request", false);
```
mykmelez: Nit: it looks like you switched to `Function.bind()` in two other places, so you should do so here as well!

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -36,6 +49,9 @@ SimulatorActor.prototype = {
 
   onPing: function(aRequest) {
     log("simulator actor received a 'ping' command");
+    if (!this.clientReady) {
+      this.clientReady = true;
+    }
```
mykmelez: There's no need to check its value before setting it, since you don't do anything else in that case. Thus you might as well just set it!

addon/lib/remote-simulator-client.js
nickdesaulniers: how about just:

``` javascript
message: {
    lat: position.coords.latitude,
    lon: position.coords.longitude,
  },
```

as in no intermediary variable?  What do you think?

mykmelez: Even better!

nickdesaulniers:
much thanks @past

PR #339, opened 2013-03-04
FIX: adb executable uri is "data/linux64" on XPCOMABI x86_64

rpl:
Last Makefile changes use same conventions as b2g binaries and on Linux platform
and puts executables into data/linux64 on 64bit systems and linux on 32bit systems.

This change fix the adb.js package to use the right path on 64bit systems.


mykmelez:
Thanks for the fix!

Issue #332, opened 2013-02-26
Add geolocation mock data support - initial longitude and latitude choice


nickdesaulniers:
The user should be able to click a button on the bottom of the simulator window (as opposed to the addon window) next to the home button.  This should open a dialog external to the simulator window.  The dialog should have radio buttons that default to "use my current location", the second non-default choice to be "enter in a longitude and latitude coordinate pair", and then maybe in a future iteration have room for a full map.  The user can then click apply or cancel to have their default changed for the duration of the simulator processes lifetime.

mykmelez:
We might want to persist a selection location across Simulator sessions, but persisting for the current session is a good first step.

nickdesaulniers:
Big thanks to @rpl , I copied most of your code from the rotation button for the geolocation button.  It gave me a lot of inspiration.  Also @kanru, who did all of the work on the FakeGeolocationProvider!

PR #340, opened 2013-03-05
add geolocation mock data support Issue #332

nickdesaulniers:
Issue #332
review? @mykmelez

I was using this app for testing: https://github.com/nickdesaulniers/GeoLocTest


.gitignore
```diff
@@ -13,3 +13,6 @@ addon/data/mac64/adb
 addon/data/win32/adb.exe
 addon/data/win32/AdbWinApi.dll
 addon/data/win32/AdbWinUsbApi.dll
+
+# Downloaded b2g builds
+*.dmg
```
mykmelez: Git best practice is to put this into a global .gitignore file, as described in [Using Git / Ignoring files](https://help.github.com/articles/ignoring-files).

prosthesis/components/FakeGeolocationProvider.js
```diff
@@ -50,20 +64,36 @@ FakeGeoPositionProvider.prototype = {
                                            Ci.nsIFakeListener,
                                            Ci.nsITimerCallback]),
   startup:  function() {
-    Services.obs.notifyObservers(null, "r2d2b2g-geolocation-request", null);
-    Services.obs.addObserver((function (message) {
-      this.currentLoc = new FakeGeoPositionObject(message.wrappedJSObject.lat,
-                                                  message.wrappedJSObject.lon);
-      this.callback.update(this.currentLoc);
-    }).bind(this), "r2d2b2g-geolocation-response", false);
+    if (this.started) return;
```
mykmelez: Nit: Mozilla convention is to wrap even single-line conditional blocks in curly braces.

prosthesis/components/FakeGeolocationProvider.js
```diff
@@ -50,20 +64,36 @@ FakeGeoPositionProvider.prototype = {
...
+    this.started = true;
+    this.walk();
+    this.updateTimer = Cc["@mozilla.org/timer;1"].createInstance(Ci.nsITimer);
+    this.updateTimer.initWithCallback(this, 1000, this.updateTimer.TYPE_REPEATING_SLACK);
```
mykmelez, line 69: One second is a very frequent update interval that will cause a lot of unnecessary IPC, since the developer's geolocation won't change nearly as often (if at all).

And it seems to me that we don't actually need to query the geolocation on an interval at all. We can instead send a single message to the Firefox process requesting the geolocation, after which the Firefox process can automatically push updates to the Simulator process as they occur until the Simulator process tells it to stop. That seems like a better approach, since then we don't do unnecessary IPC, and updates only travel over the wire as often as they happen.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -28,14 +28,14 @@ function SimulatorActor(aConnection)
...
-  Services.obs.addObserver((function() {
+  /*Services.obs.addObserver((function() {
     if (this.clientReady) {
       this._connection.send({
         from: this.actorID,
         type: "geolocationRequest"
       });
     }
-  }).bind(this), "r2d2b2g-geolocation-request", false);
+  }).bind(this), "r2d2b2g-geolocation-request", false);*/
```
mykmelez: Let's remove this code if we aren't going to use it.

prosthesis/components/FakeGeolocationProvider.js
nickdesaulniers, line 69: There is no IPC going on in this case.  IPC occurs once when the simulator is started, and the resulting coordinates are stored within prosthesis as the initial latitude and initial longitude.  The timer is used in this case for calls to navigation.geolocation.watchPosition, to query the current latitude and longitude, which may differ from the initial coordinates if the user chooses mock coordinates.  This allows you to change the coordinates being reported as the position is being watched.  I agree that hard coding the interval is not the best thing to do, but in future iterations on this code, it would be nice to allow the developer to change this.

prosthesis/components/FakeGeolocationProvider.js
```diff
@@ -2,11 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-Components.utils.import("resource://gre/modules/XPCOMUtils.jsm");
-Components.utils.import("resource://gre/modules/Services.jsm");
+// http://www.w3.org/TR/2010/CR-geolocation-API-20100907/#api_description
+// http://mxr.mozilla.org/mozilla-central/source/xpcom/system/nsIGeolocationProvider.idl#32
+// http://mxr.mozilla.org/mozilla-central/source/dom/system/NetworkGeolocationProvider.js#279
```
nickdesaulniers: Whoops, don't need these!

.gitignore
```diff
@@ -13,3 +13,4 @@ addon/data/mac64/adb
 addon/data/win32/adb.exe
 addon/data/win32/AdbWinApi.dll
 addon/data/win32/AdbWinUsbApi.dll
+
```
nickdesaulniers: @mykmelez , should I fix this?  I added newlines to the end of some files.

prosthesis/content/geolocation.xul
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
+
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this file,
+   - You can obtain one at http://mozilla.org/MPL/2.0/.  -->
+
+<?xml-stylesheet href="chrome://global/skin/" type="text/css"?>
+<?xml-stylesheet href="chrome://prosthesis/skin/geo.css" type="text/css"?>
```
mykmelez: I looked into why _chrome://global/skin/_ doesn't seem to be applied, and I discovered that it is applied, as are other stylesheets that the platform applies by default (namely, _ua.css_ and _xul.css_); but another stylesheet, B2G's _chrome://browser/content/content.css_, disables some of the default styles (like those for XUL buttons).

We should reenable those styles, and it's easy to do. Just add the following line to _prosthesis/chrome.manifest_:

```
override chrome://browser/content/content.css chrome://prosthesis/content/global.css
```

This overrides that B2G stylesheet with our own. Note that our own stylesheet doesn't actually have to exist. Even without a stylesheet at that URL, the platform will still override the B2G stylesheet.

This change may regress B2G stylesheet rules that we want to retain. So after we do this, we'll want to watch out for those and resolve them by copying styles from the B2G stylesheet to our own. In particular, I expect that scrollbars will behave differently. But we can do that work in a separate changeset.

prosthesis/content/geolocation.xul
```diff
@@ -0,0 +1,51 @@
...
+                 increment="0.1" decimalplaces="5" disabled="true"/>
+      </row>
+    </rows>
+  </grid>
+  <hbox>
+    <spacer flex="1"/>
+    <button label="Apply" class="choice" id="apply"/>
+    <button label="Cancel" class="choice" id="cancel"/>
+    <spacer flex="1"/>
+  </hbox>
```
mykmelez: Instead of creating these buttons yourself, change the top-level element from _&lt;window>_ to _&lt;dialog>_, which will automagically create nativeish accept/cancel buttons for you using the [dialog XBL binding](http://mxr.mozilla.org/mozilla-central/source/toolkit/content/widgets/dialog.xml), which dispatches _dialogaccept_ and _cancel_ events when the user presses the buttons. See the [print page setup dialog](http://mxr.mozilla.org/mozilla-central/source/toolkit/components/printing/content/printPageSetup.xul), for an example, noting especially its _<dialog>_ element's _ondialogaccept_ and _oncancel_ attributes.

prosthesis/content/shell.js
```diff
@@ -20,3 +20,50 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+        } else {
+          latitude = params.output.lat || latitude;
+          longitude = params.output.lon || longitude;
+        }
+      },
+      gotCoords = function gotCoords(message) {
+        latitude = initialLatitude = message.wrappedJSObject.lat;
+        longitude = initialLongitude = message.wrappedJSObject.lon;
+        document.getElementById("geolocationButton")
+                .addEventListener("click", openWin);
```
mykmelez, line 55: We shouldn't wait until we get the developer's current geolocation to react to button clicks, because that can take a long time (up to and including forever), and it's quite confusing to click that button and have nothing happen.

.gitignore
mykmelez: Yes, thanks! Whitespace corrections are always welcome. :-)

mykmelez: Erm, although in this case it looks like you added an extra newline, which is unnecessary.

prosthesis/components/FakeGeolocationProvider.js
mykmelez, line 69: Erm, sorry, my bad on the IPC. I've now reviewed the whole changeset and think I understand it!

So, it looks like what's happening is that the FakeGeoPositionProvider, once instantiated by a geolocation request from an app, starts requesting a geolocation and updating its callback once per second.

That interval is too frequent, since real geolocation providers are unlikely to call their callbacks that frequently, even when the geolocation is changing (and in this case it mostly won't be changing).

Instead of updating its callback on a fixed schedule, the provider should update it only when the geolocation actually changes, whether because the developer specifies a new custom location or Firefox reports a new current location.

Also, the provider doesn't have to poll for changes by sending r2d2b2g-geolocation-request. It can simply observe r2d2b2g-geolocation-response, which the code in shell.js can then notify it about whenever the geolocation changes.

prosthesis/components/FakeGeolocationProvider.js
```diff
@@ -40,8 +45,17 @@ FakeGeoPositionObject.prototype = {
...
+  this.started = false;
   this.callback = null;
   this.currentLoc = null;
+
+  Services.obs.addObserver((function (message) {
+    dump(message.wrappedJSObject.lat + ',' + message.wrappedJSObject.lon + '\n');
+    this.currentLoc = new FakeGeoPositionObject(message.wrappedJSObject.lat,
+                                                message.wrappedJSObject.lon);
+    this.callback.update(this.currentLoc);
+  }).bind(this), "r2d2b2g-geolocation-response", false);
```
mykmelez, line 53: _this.callback_ is not guaranteed to exist until _FakeGeoPositionProvider.watch()_ is called, so this observer should not assume it exists. But you don't have to test for it; instead, just move this observer registration to the _FakeGeoPositionProvider.watch()_ method, since it doesn't matter until then anyway.

prosthesis/content/shell.js
```diff
@@ -20,3 +20,50 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+      sendCoords = function sendCoords() {
+        Services.obs.notifyObservers({
+          wrappedJSObject: {
+            lat: latitude,
+            lon: longitude,
+          }
+        }, "r2d2b2g-geolocation-response", null);
+      };
+
+  Services.obs.addObserver(gotCoords, "r2d2b2g-geolocation-setup", false);
```
mykmelez, line 66: I wonder if retrieval of the current geolocation in dbg-simulator-actors.js could somehow race this observer registration.

But in any case my more significant concern is that this makes prosthesis retrieve the current geolocation on startup without first prompting the developer to provide it. Whereas in the current codebase we wait to retrieve the geolocation until an app requests it and the developer confirms the access prompt.

Nevertheless, I do see some value in defaulting the latitude/longitude fields to the developer's current geolocation. I'm just not sure how to reconcile that with our interest in confirming that the Simulator should have access to that geolocation.

prosthesis/skin/shell.css
```diff
@@ -38,3 +38,13 @@ toolbarbutton > .toolbarbutton-text {
...
+
+#geolocationButton {
+  list-style-image: url("chrome://prosthesis/skin/geo.png");
+  -moz-image-region: rect(0, 9px, 14px, 0);
+}
+
+#geolocationButton:hover {
+  -moz-image-region: rect(0, 18px, 14px, 9px);
+}
+
```
mykmelez, line 50: Note to self: move all prosthesis/skin/ files to prosthesis/content/ and remove the prosthesis/skin/ directory to fix issue #343.

prosthesis/content/geolocation.xul
nickdesaulniers: from _window_ to _dialog_ ?

mykmelez: Heh, yes! From _&lt;window>_ to _&lt;dialog>_. I didn't realize that GitHub would strip my tag references from my comment.

prosthesis/content/shell.js
nickdesaulniers, line 55: On the other hand, if we open the window without the coordinates, we can't update them with the window open (they are passed into the window as arguments when opening).  The internal coordinates will be updated, just the developer would have to close and reopen the geolocation window to see them.  This is assuming they open the geolocation window before it has received the geolocation coordinates.  Is that ok?

mykmelez, line 55: You can update the coordinates after the dialog has been opened, you just need to observe the same notifications in the dialog. Or we can instead not use nor display the coordinates in the dialog at all. Which would also solve the problem with requesting the geolocation before the developer has been prompted about it. Although I do see the value in using the coordinates in this dialog. And we should find a better way to get the developer to confirm that we can use the current geolocation.

prosthesis/content/geolocation.xul
nickdesaulniers: Scroll bars look fine, just keep an eye for any other regressions in the future.

prosthesis/components/FakeGeolocationProvider.js
nickdesaulniers, line 53: I think it would be better to keep it in the constructor: a geolocation provider is treated as a singleton, it is created once and only once.  Watch may be called multiple times.  Moving the observer into watch will just attach additional listeners.  

I could add:

``` javascript
if (typeof this.callback.update === "function") {
  this.callback.update(this.currentLoc);
}
```

What do you think?

prosthesis/content/shell.js
nickdesaulniers, line 55: Maybe have 3 choices: default to something like 0,0 or Mountain View, use my current, or use these specified?  What do you think?

mykmelez, line 55: I like it!

prosthesis/content/shell.js
nickdesaulniers, line 66: From IRC:
Jorgev: if it's opt-in, there's no problem [1:26pm]
Jorgev: it doesn't matter you're not using the regular UI in this case [1:26pm]

prosthesis/content/shell.js
nickdesaulniers, line 55: I think the 3 choices can be condensed into 2 choices, and my convo with jorgev enhances my belief:
- Custom (default) (defaults to mozilla HQ gps)
- Current (this addon may fetch and use my current gps coordinates)

I'll start reworking the code to only request the current coords if the developer "opts in" by choosing the "Use my current location" radio button.

nickdesaulniers, line 55: @mykmelez , should the "Current location" choice rerequest it every time or memoize it?  I don't think developers will ever move around while developing, but maybe there's something that I'm overlooking?  What do you think?

mykmelez, line 55: Developers won't move around most of the time, but they might do so occasionally. For example, a developer might work from a train while commuting. They might even do so intentionally, taking advantage of their daily commute to test geolocation changes in their app.

So it'd be useful to keep the current geolocation updated, although it isn't essential. We can either tackle it here or in a followup change. Either way, I expect that it's fairly straightforward to implement.

Currently, prosthesis sends a request to which the addon then sends a single response after getting the position via _nsIDOMGeoGeolocation.getCurrentPosition()_. To keep the current location up-to-date, we just have to make the addon register a listener via _nsIDOMGeoGeolocation.watchPosition()_ and then send a response every time it receives an update.

nickdesaulniers, line 55: Yeah sounds good, I've spun these out into two separate issues, #356 and #357 .  I'd like to see the code here in this pull request land and then take on each one separately.

mykmelez:
It looks like the only remaining issues are that the Simulator requests the current geolocation on startup, regardless of whether or not it's needed; and the FakeGeolocationProvider updates the callback once per second, whether or not the geolocation has changed.

To fix these issues, we're going to have to revamp the way the various JavaScript contexts (shell, FakeGeolocationProvider, dbg-simulator-actors) communicate with each other. I'm inclined to say that we should do that in a followup change, since this changeset is already involved. But it's essential that we do it, as those are both significant issues.

PR #347, opened 2013-03-06
trim curly braces from UUIDs used as app IDs

mykmelez:
[Over in dev-b2g](https://groups.google.com/d/topic/mozilla.dev.b2g/j-r8ed6yAS4/discussion), @ArnaudD reported that an app he was testing in the Simulator had an app:// URL with a GUID containing curly braces, which violates RFC 3986.

@fabricedesre is fixing B2G's app ID generator to trim the braces in [bug 845344](https://bugzilla.mozilla.org/show_bug.cgi?id=845344). We should similarly fix the Simulator's app ID generator.

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -41,7 +41,7 @@ function ensureXkeysUnique() {
     let app = Simulator.apps[key];
 
     // Give the app a new unique xkey.
-    app.xkey = UUID.uuid().toString();
+    app.xkey = UUID.uuid().toString().slice(1, -1);
 
     // For "local" (i.e. packaged) apps, make the origin and manifest URL match
     // the xkey, since we'll use the xkey as the ID of the app in
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -34,9 +34,11 @@ PageMod({
 });
 
 /**
- * Ensure app xkeys are unique.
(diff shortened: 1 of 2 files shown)


mykmelez:
@rpl How does this look to you?

rpl:
@mykmelez we should probably regenerate xkeys on registered app during next user upgrades,
something like: 

```
diff --git a/addon/lib/main.js b/addon/lib/main.js
index 72e1650..c4c4960 100644
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -74,6 +74,11 @@ if (["install", "downgrade", "upgrade"].indexOf(Self.loadReason) >= 0) {
       if (Services.vc.compare(lastVersion, "2.0pre5") < 0) {
         ensureXkeysUnique();
       }
+      if (Services.vc.compare(lastVersion, "3.0pre2") < 0) {
+        // NOTE: trim curly braces from xkeys UUIDs
+        // generated from older simulator releases
+        ensureXkeysUnique();
+      }
```

mykmelez:
Good catch! I think this is the right fix, as we don't need to check for both < 2.0pre5 and < 3.0pre3, since we do the same thing in either case.

rpl:
Indeed! it looks good, merging now.

ArnaudD:
Great ! Thanks !

PR #355, opened 2013-03-07
FIX: prevents default on "validation errors link" click handler

### LINKED ISSUE #352: Refresh after clicking on an error does not display app list
STR:
1. Start simulator
2. Install app with invalid manifest so that an error is reported
3. Click on the (X errors and X warnings)
4. refresh the page

Expected results:
Installed apps are re-listed

Actual results:
Installed apps are not re-listed

Looks like the apps are not re-listed when there's an empty fragment identifier on the end of the URL:
resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/index.html#

rpl:
fix #352

IMPLEMENTATION NOTE: "return false;" in a jQuery event handler is effectively the same as calling both event.preventDefault and event.stopPropagation (http://stackoverflow.com/a/1357151)

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -268,6 +268,7 @@ var Simulator = {
                       .prop("title", "expand validation messages")
                       .click(function(evt) {
                         listContainerEl.toggle();
+                        return false;
                       }),
                     listContainerEl);
                 } else {


mykmelez:
Excellent.

Issue #365, opened 2013-03-08
Validation Result INVALID for just warnings


digitarald:
The validator current marks the validation as failed when only 1 warning was found.

Warnings shouldn't mark validation failed.

On wording: Does something actually fail, or would it make more sense to label it "Invalid Manifest"? We still allow all actions on the app, it's a bit confusing.

rpl:
should be fixed by #375 

now app with warnings are market "WARNINGS" instead of "INVALID"

I'm closing it, but please reopen if the current solution is not enough.

PR #375, opened 2013-03-14
FIX: prevent silent errors on add/undo error

### LINKED ISSUE #374: "add to Dashboard" silent fails to appear if validation doesn't manage to start simulator
When you add an app to the dashboard, the validator runs the simulator to validate the permissions. But if the simulator doesn't start for some reason, then the validation process never finishes, and the app doesn't appear on the dashboard.

Reloading the dashboard causes the app to show up. But it hasn't been validated, and I suspect it also isn't initialized correctly (since much initialization happens in _updateApp(), which doesn't get called until updateApp() gets called back by the validator).

I noticed this because it happened to me on several occasions over the last couple days. Unfortunately, I don't have steps to reproduce consistently. When it occurs, I can often cause it to recur a few times by removing and readding the same app. But eventually it stops recurring. No error messages are written to the consoles.

We should find and fix whatever problem is preventing the simulator from starting up. But I wonder if we should also remove the dependency between the validator and the simulator, given that apps you add to the dashboard should always show up on it. They should never silently fail to appear.

@rpl, what do you think?

rpl:
prevent silent errors in the dashboard due to an b2g-desktop instance stalled:
- call simulator.sendListApps on add/undo error to immediatelly list the new app
- update and check cached manifest before validation (to be able to update the app name in the dashboard)
- define a new app.validation.running attribute
- dashboard app validation result text: RUNNING, WARNINGS, INVALID, OK

should fix #374 and #365

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -178,23 +178,46 @@ let simulator = module.exports = {
 
   updateApp: function(id, next) {
     console.log("Simulator.updateApp " + id);
-    // refresh app list on the dashboard before validation
-    // (prevents silent errors on a stalled validation)
-    simulator.sendListApps();
 
-    simulator.validateApp(id, function(error, app) {
-      // update dashboard app validation info
-      simulator.sendListApps();
+    simulator._updateCachedManifest(id, function(error, manifest) {
+      let app = simulator.apps[id];
+      app.validation = {errors: [], warnings: []};
 
-      if (!error) {
-        // NOTE: try to updateApp if there isn't any blocking error
-        simulator._updateApp(id, next);
-      } else {
-        // validation error
+      if (error) {
+        app.validation.errors.push("Error updating cached Manifest: " + error);
         if (typeof next === "function") {
-          next(error, app);
+          // NOTE: blocking error
+          next(Error("Unable to read manifest: '" + id + "'."), app);
+        }
+        return;
+      }
+
+      if (!app.manifest) {
+        app.validation.errors.push("Missing Manifest.");
+        if (typeof next === "function") {
+          // NOTE: blocking error
+          next(Error("Missing manifest: '" + id + "'."), app);
+          return;
         }
       }
+      // update dashboard app validation info to prevents silent errors
+      // on stalled validations
+      simulator.sendListApps();
+
+      simulator.validateApp(id, function(error, app) {
+        // update dashboard app validation info
+        simulator.sendListApps();
+
+        if (!error) {
+          // NOTE: try to updateApp if there isn't any blocking error
+          simulator._updateApp(id, next);
+        } else {
+          // validation error
+          if (typeof next === "function") {
+            next(error, app);
+          }
+        }
+      });
     });
   },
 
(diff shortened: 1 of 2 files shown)


mykmelez:
If the _updateApp_ callback is going to call _sendListApps_ unconditionally, regardless of whether or not there was an error, then that call should happen outside the `if (error)` conditional.

But calling _sendListApps_ unconditionally in the _updateApp_ callback doesn't solve the problem anyway, since _updateApp_ waits on _validateApp_, and _validateApp_ never calls its callback when the simulator fails to start.

I think what we need to do here is make _addAppByDirectory_, _undoRemoveApp_, and _addManifest_ add the app to the list via _sendListApps_ before updating the app via _updateApp_. In the case of _addAppByDirectory_, this will also mean reading the manifest and adding it to the app record.

rpl:
indeed! now I understand better the issue.

@mykmelez I just pushed a tweak: sendListApps as the first action in updateApp callbacks (excluding updateAll)
and add a sendListApps inside the updateApp method just before the validation sequence (this should list the app into the
dashboard just after any "add*" and prevents silent errors on stalled validations)

addon/lib/simulator.js
```diff
@@ -177,6 +178,10 @@ let simulator = module.exports = {
 
   updateApp: function(id, next) {
     console.log("Simulator.updateApp " + id);
+    // refresh app list on the dashboard before validation
+    // (prevents silent errors on a stalled validation)
+    simulator.sendListApps();
```
mykmelez: This is too early, as __updateCachedManifest_ hasn't been called yet, so we don't have enough information about the new app yet, and its listing looks broken (in particular, its name is missing).

mykmelez: We could fix this by moving the __updateCachedManifest_ call from _validateApp_ to _updateApp_, which seems like the minimal change needed to display the app correctly whether or not validation completes.

Then, in the long run, let's consider what else we can do to distinguish more clearly between adding an app, updating it, and validating it, since we seem to be conflating those operations, but they are discrete, even if we currently always update and validate an app after adding it.

rpl:
@mykmelez this time I should be able to better cover the different cases ;-) 
- update and check cached manifest before validation (to be able to update the app name in the dashboard)
- define a new app.validation.running attribute
- dashboard app validation result text: RUNNING, WARNINGS, INVALID, OK
  (should fix #365)

in a separate feature branch, could be interesting if I extract an "app-manager.js" module from the "simulator.js" module
(as in my old job_scheduler experimental branch) to create a more clear app management api and workflow?

mykmelez:
> in a separate feature branch, could be interesting if I extract an "app-manager.js" module from the "simulator.js" module (as in my old job_scheduler experimental branch) to create a more clear app management api and workflow?

I do think we should refactor the _Simulator_ object, which is large and performs many functions that aren't closely related. I'm not sure about an _app-manager_ module specifically, though; can you describe which functions you would move to it?

rpl:
> In that experimental branch, _app-manager_ was responsible for:

At first glance, this seems like a reasonable set of functions to factor out into an app management module, although there are probably some details to consider.

PR #388, opened 2013-03-23
Update main.css

### LINKED ISSUE #387: Hidden text in the manifest.webapp input box
I got confused by the extension of the manifest file and because only "URL for page or manifest.weba" is visible, I thought that the extension was ".weba".

This happens because the input box is too small.

davidgomes:
Fixed #387 by increasing width of the input box on main.css.

### DIFF
--- a/addon/data/content/css/main.css
+++ b/addon/data/content/css/main.css
@@ -591,7 +591,9 @@ figure {
     border: 2px darkred dashed;
     background-color: #E5CBCB;
 }
-
+#add-app-url {
+    width: 220px;
+}
 button, input[type="url"], input[type="text"] {
     font-size: 0.8em;
     color: #444;


mykmelez:
As @nickdesaulniers noted in #387, this probably depends on the screen resolution (or perhaps the default system font, or something else that affects text width). And I don't see it on my screen. Nevertheless, I can believe it happens!

And it's worth making that field a little wider anyway. After all, we enter URLs into it, and they frequently take up at least as much space as we currently provide. So I have merged the change.

@davidgomes: thanks for the fix!

cc: @digitarald so he's aware of this issue, as he (still?) has plans to change the way you add apps to the Dashboard.

mykmelez:
Followup change to add @davidgomes to the list of contributors: https://github.com/mozilla/r2d2b2g/commit/e00700fabe2601b6800afda569626542c14af808

PR #399, opened 2013-03-28
FIX: random allocated b2g debugger remoteport saved as connect page submit side-effect

rpl:
Trying to use the new experimental Browser Debugger in Firefox Nightly
to debug simulator addon code,
I notice that it doesn't work on my profile because connect.xhtml saved
random allocated port as a side effect of the submit action:

https://mxr.mozilla.org/mozilla-central/source/browser/devtools/framework/connect/connect.js#55

and so Firefox try to start a debugger server on the same port of the connected b2g instance.

proposed changes to workaround this bug:
- restore previously saved remote debugger port on connect.xhtml tab close
- on install/upgrade restore standard remote debugger port (6000) if previously
  installed version < 3.0pre5

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -56,6 +56,12 @@ function ensureXkeysValid() {
   }
 }
 
+// Restore standard remote debugger port
+// (autosaved on submit by connect.xhtml on simulator < 3.0pre5)
+function restoreStardardRemoteDebuggerPort() {
+  Services.prefs.setIntPref("devtools.debugger.remote-port", 6000);
+}
+
 // Retrieve the last addon version from storage, and update storage if it
 // has changed, so we can do work on addon upgrade/downgrade that depends on
 // the last version the user used.
@@ -76,7 +82,9 @@ if (["install", "downgrade", "upgrade"].indexOf(Self.loadReason) >= 0) {
       if (Services.vc.compare(lastVersion, "3.0pre3") < 0) {
         ensureXkeysValid();
       }
-      SStorage.storage.needsUpdateAll = true;
+      if (Services.vc.compare(lastVersion, "3.0pre5") < 0) {
+        SStorage.storage.needsUpdateAll = true;
+      }
     }
   }
 }
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -58,7 +58,7 @@ function ensureXkeysValid() {
 
 // Restore standard remote debugger port
 // (autosaved on submit by connect.xhtml on simulator < 3.0pre5)
-function restoreStardardRemoteDebuggerPort() {
+function restoreStandardRemoteDebuggerPort() {
   Services.prefs.setIntPref("devtools.debugger.remote-port", 6000);
 }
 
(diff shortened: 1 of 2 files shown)


addon/lib/main.js
```diff
@@ -56,6 +56,12 @@ function ensureXkeysValid() {
   }
 }
 
+// Restore standard remote debugger port
+// (autosaved on submit by connect.xhtml on simulator < 3.0pre5)
+function restoreStardardRemoteDebuggerPort() {
```
mykmelez: Nit: Stardard -> Standard

addon/lib/main.js
```diff
@@ -76,7 +82,9 @@ if (["install", "downgrade", "upgrade"].indexOf(Self.loadReason) >= 0) {
       if (Services.vc.compare(lastVersion, "3.0pre3") < 0) {
         ensureXkeysValid();
       }
-      SStorage.storage.needsUpdateAll = true;
+      if (Services.vc.compare(lastVersion, "3.0pre5") < 0) {
+        SStorage.storage.needsUpdateAll = true;
+      }
```
mykmelez, line 87: Can you explain this change further? I don't understand why we would make _needsUpdateAll_ conditional in this way.

addon/lib/simulator.js
```diff
@@ -860,12 +860,16 @@ let simulator = module.exports = {
...
+    let originalPort = Services.prefs.getIntPref("devtools.debugger.remote-port");
     Tabs.open({
       url: "chrome://browser/content/devtools/connect.xhtml",
       onReady: function(tab) {
         // NOTE: inject the allocated remoteDebuggerPort on the opened tab
         tab.attach({
           contentScript: "window.addEventListener('load', function() { document.getElementById('port').value = '"+port+"'; }, true);"
+        }).on('detach', function restoreOriginalRemotePort() {
+          // restore previous value (autosaved on submit by connect.xhtml)
+          Services.prefs.setIntPref("devtools.debugger.remote-port", originalPort);
```
mykmelez, line 872: Nice!

mykmelez:
I'm not sure how important it is to restore the port for users for older versions of the addon, but I suppose it doesn't hurt.

addon/lib/main.js
rpl, line 87: no reason, my faut

I'm pushing a fix.

rpl:
@mykmelez I pushed a couple of tweaks.

If we don't need to restore the port for users of older simulator versions, I can remove that part of this pull request in no time. 
(probably it's not important, but tcp 6000 is often advertised in articles about the upcoming browser debugger, so I thought this change could reduce risks to confuse users).

mykmelez:
> (probably it's not important, but tcp 6000 is often advertised in articles about the upcoming browser debugger, so I thought this change could reduce risks to confuse users).

That's a good point. Even if it isn't necessary, I bet it'll make life better for some of our users, and the code is short and sweet, so it doesn't add much complexity or size. Therefore, let's keep it!

mykmelez:
This looks great, merging...

Issue #343, opened 2013-03-05
remove unnecessary prosthesis/skin/ directory


mykmelez:
The prosthesis/skin/ directory is unnecessary, because we don't intend to support multiple skins for the Simulator, which is the only reason to have a skin/ directory instead of putting style files into the content/ directory.

So we should move all the style files to prosthesis/content/ and remove the prosthesis/skin/ directory.

mykmelez:
Fixed by #456.

PR #402, opened 2013-03-30
FIX: restore b2g agent sheet scrollbars and other mobile tweaks on b2g windows

### LINKED ISSUE #379: scrollbar appears in various apps
Scroll bar is visible on Windows emulator.
Just run it, and get top panel out.
And all applications that dont fit vertically. Even worse when app need whole width, and it will show horizontal scroll bar as well.

Maybe it is time for Mozilla to think about transparent scroll bar like on Mac OS X in Chrome?

rpl:
- new DOM window detected by notification observers
- inject full original b2g agent sheet css
  (e.g. to style buttons as in the FirefoxOS device)

should fix #379 

NOTE: there're some remaining issues related to shell styling, 
because currently any try to inject b2g agent sheet into the homescreen
iframe generates some weird side-effect (homescreen change rotation
instead to lock to portrait) but probably it's more important to restore the style for apps now, and then in a second time try to restore it on the homescreen as well.

### DIFF
--- a/prosthesis/content/inject-b2g-agentsheet.js
+++ b/prosthesis/content/inject-b2g-agentsheet.js
@@ -1,0 +1,15 @@
+const B2G_AGENTSHEET_URL = Services.io.newURI("chrome://prosthesis/content/b2g.css",
+                                              null, null);
+
+// inject b2g mobile agent stylesheet into
+// all next content DOM windows created
+Services.obs.addObserver(
+  function _injectStylesheet(subject, topic, data) {
+    debug("injectStylesheet:", data);
+    let winUtils = subject.QueryInterface(Ci.nsIInterfaceRequestor).
+      getInterface(Ci.nsIDOMWindowUtils);
+    winUtils.loadSheet(B2G_AGENTSHEET_URL, winUtils.AGENT_SHEET);
+  },
+  "content-document-global-created",
+  false
+);
--- a/prosthesis/content/window-manager.js
+++ b/prosthesis/content/window-manager.js
@@ -9,6 +9,9 @@ Cu.import("resource://prosthesis/modules/GlobalSimulatorScreen.jsm");
 navigator.mozSettings
   .createLock().set({'homescreen.ready': false});
 
+const B2G_AGENTSHEET_URL = Services.io.newURI("chrome://prosthesis/content/b2g.css",
+                                              null, null);
+
 let SimulatorWindowManager = {
   init: function() {
     // inject the b2g agent stylesheet into the firefoxos iframe container
@@ -68,11 +71,9 @@ let SimulatorWindowManager = {
     );
   },
   _injectStylesheet: function(win) {
-    let URL = Services.io.newURI("chrome://prosthesis/content/b2g.css",
-                                 null, null);
     let winUtils = win.QueryInterface(Ci.nsIInterfaceRequestor).
-      getInterface(Ci.nsIDOMWindowUtils);
-    winUtils.loadSheet(URL, winUtils.AGENT_SHEET);
+                   getInterface(Ci.nsIDOMWindowUtils);
+    winUtils.loadSheet(B2G_AGENTSHEET_URL, winUtils.AGENT_SHEET);
   },
   _initKeepWindowSize: function() {
     // WORKAROUND: keep the simulator window size
--- a/prosthesis/content/window-manager.js
+++ b/prosthesis/content/window-manager.js
@@ -11,6 +11,9 @@ navigator.mozSettings
 
 let SimulatorWindowManager = {
   init: function() {
+    // inject the b2g agent stylesheet into the firefoxos iframe container
+    this._injectStylesheet(this._homescreen);
+
     this._initObservers();
     this._initKeepWindowSize();
     this._initMutationObservers();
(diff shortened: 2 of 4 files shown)


rpl:
Screenshots of a couple of apps with reinjected b2g stylesheet:

app_with_mobile_scrollbar
firefoxos_browser_mobile_stylesheet

rpl:
Screenshots of a remaining issues on homescreen stylesheet:

style_issue_on_b2g_shell_style1
style_issue_on_b2g_shell_style2

mykmelez:
_FloatingScrollbars.jsm_ looks like it is designed to enable a stylesheet to be added and removed on demand during the lifetime of a content frame. In this case, however, we only need to add a stylesheet once, when a content frame is first created; and we don't need to remove it.

So we should be able to do something simpler, like observing content window creation and adding a stylesheet to each content window that gets created. Here's a self-contained example that works in my tests, including for the homescreen:

```
Services.obs.addObserver(
  function injectStylesheet(subject, topic, data) {
    debug("injectStylesheet: " + data);
    let URL = Services.io.newURI("chrome://prosthesis/skin/b2g-content.css",
                                 null, null);
    let winUtils = subject.QueryInterface(Ci.nsIInterfaceRequestor).
                   getInterface(Ci.nsIDOMWindowUtils);
    winUtils.loadSheet(URL, winUtils.AGENT_SHEET);
  },
  "content-document-global-created",
  false
);
```

It should be straightforward to integrate that into _SimulatorWindowManager_, so we shouldn't need a separate _B2GAgentSheet.jsm_ module.

Also, I think _content-document-global-created_ will suit our purpose just fine, but note the others listed on [Observer Notifications](https://developer.mozilla.org/en-US/docs/Observer_Notifications). If we run into problems adding the stylesheet to the right set of windows at the right time, then there are other options.

prosthesis/skin/b2g-content.css
```diff
@@ -0,0 +1,289 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mykmelez: Instead of putting this into the _skin/_ directory and calling it _b2g-content.css_, put it into the _content/_ directory and call it simply _b2g.css_. So its URL will be _chrome://prosthesis/content/b2g.css_. Over in issue #343 I plan to remove the _skin/_ directory; in the meantime, we shouldn't add anything to it anymore.

rpl:
> So we should be able to do something simpler, like observing content window creation and adding a stylesheet to each content window that gets created.

@mykmelez awesome! it's much more simple and clean now.

I've just rebased this pull request on the current master and updated this pullrequest.

I wasn't aware of this notification, in my tests it's working as well as the previous prototype (using less code :-P),
it doesn't seem to be able to solve the remain corner cases. 

In 4e18ea9 I've tweaked the suggested fix, injecting b2g.css stylesheet into the main FirefoxOS iframe 
to fix the remain style issues (e.g. app switcher and mobile select)

prosthesis/content/window-manager.js
```diff
@@ -52,6 +55,24 @@ let SimulatorWindowManager = {
...
+        debug("injectStylesheet: " + data);
+        this._injectStylesheet(subject);
+      }).bind(this),
+      "content-document-global-created",
+      false
+    );
+  },
+  _injectStylesheet: function(win) {
+    let URL = Services.io.newURI("chrome://prosthesis/content/b2g.css",
+                                 null, null);
```
mykmelez: Nit: as a variable, this name should be lower-case, i.e. _url_ (even though I mistakenly used upper-case in the original example!). Alternately, you could make this a global const like _B2G_STYLESHEET_URL_, since it never changes.

prosthesis/content/window-manager.js
```diff
@@ -52,6 +55,24 @@ let SimulatorWindowManager = {
...
+    let winUtils = win.QueryInterface(Ci.nsIInterfaceRequestor).
+      getInterface(Ci.nsIDOMWindowUtils);
```
mykmelez: Nit: indent this as:

```
    let winUtils = win.QueryInterface(Ci.nsIInterfaceRequestor).
                   getInterface(Ci.nsIDOMWindowUtils);
```

mykmelez:
> I wasn't aware of this notification, in my tests it's working as well as the previous prototype (using less code :-P),
> it doesn't seem to be able to solve the remain corner cases. 

Which corner cases remain? It looks like you've fixed the homescreen issues. Or are there still some issues with the homescreen?

rpl:
> Which corner cases remain? It looks like you've fixed the homescreen issues. Or are there still some issues with the homescreen?

@mykmelez my previous try (to fix the stylesheet issues in app switcher or mobile select) introduced a random 
rendering/layouting bug when we run an app on a starting b2g instance, e.g. pressing the run button when
b2g desktop is not already running)

I fixed the bug and removed the side-effect in the last commit (91d85de):
- move b2g stylesheet injection code into its own file (inject-b2g-agentsheet.js)
- observe "content-document-global-created" notification before ContentStart to
  inject into "iframe#homescreen" created inside shell.xul
  (prevents layouting bug on running app during b2g startup)

now it should work correctly in all the cases (at least in the known cases)

prosthesis/content/inject-b2g-agentsheet.js
```diff
@@ -0,0 +1,15 @@
+const B2G_AGENTSHEET_URL = Services.io.newURI("chrome://prosthesis/content/b2g.css",
```
mykmelez, line 1: This line exceeds 80 characters, so it would be better to format the statement as:

```
const B2G_AGENTSHEET_URL =
  Services.io.newURI("chrome://prosthesis/content/b2g.css", null, null);
```

Also, this file needs a license header.

mykmelez:
Looks good! (modulo the minor issues in the previous comment)

Ideally, we would not create a separate JavaScript file for this small amount of code, but there isn't currently another file that could incorporate it sensibly. It would make sense to put it into shell.js, but we load that on ContentStart at the moment, because months ago we had trouble loading it early. But we should try that again, since the bug I encountered, which I didn't have time to investigate further back then, might well have been fixed by now.

mykmelez:
I'm going to fix the last two issues and then merge this so I can get it into the preview builds I'm about to spin.

PR #415, opened 2013-04-07
FEATURE: remote scratchpad 

rpl:
This prototype adds a new DevTools command ("firefoxos scratchpad")
to open a tweaked scratchpad window connected to the controlled 
b2g-desktop instance using a new custom "scratchpad actor", 
it supports 3 switchable contexts:
- current displayed FirefoxOS app context ("app")
- current FirefoxOS browser tab ("browser-tab")
- FirefoxOS Chrome shell.xul ("chrome")

Supported scratchpad features:
- run selection/buffer
- display selection/buffer
- load / save to file

Not supported scratchpad features:
- inspect (needs more work)
- reload and run
- reset variables

DEMO SCREENCAST: http://youtu.be/M8jW8F7OCJU

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -183,6 +183,15 @@ Gcli.addCommand({
   },
 });
 
+Gcli.addCommand({
+  name: "firefoxos scratchpad",
+  description: "...",
+  params: [],
+  exec: function(args, context) {
+    Simulator.remoteSimulator.openScratchpad();
+  },
+});
+
 let PermissionSettings;
 try {
   PermissionSettings =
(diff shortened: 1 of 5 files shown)


mykmelez:
@rpl I'm going to close this for now, as it seems to have stagnated, and the open pull request is deceptive (it makes it look like there's currently something to review and merge). Nevertheless, it's a good idea! It also seems like the kind of thing we might get for free with the work that the DevTools group is doing to make all devtools remoteable. However, I'm not sure if the Scratchpad is actually included in that work.

PR #417, opened 2013-04-09
close Dashboard when user disables/updates addon (fixes #416)

### LINKED ISSUE #416: Dashboard remains open after addon disabled
If the Dashboard page is open when the addon is disabled, it remains open. It should close, since it's part of the addon's interface, and users disable addons in order to remove their interfaces (and other resources) from Firefox.

Leaving the Dashboard open also causes user confusion when the addon is reenabled (which commonly occurs after the addon is disabled during an update), because the Dashboard will no longer be connected to the addon's main module, and it will appear to be broken until you reload it. I suspect this is the cause of @jincreator's problem in [issue #385, comment #15666572](https://github.com/mozilla/r2d2b2g/issues/385#issuecomment-15666572).

mykmelez:


mykmelez:
@rpl: how does this look to you?

rpl:
it looks good!
merged.

PR #418, opened 2013-04-09
suppress duplicate error notification when adding app (fixes #358)

### LINKED ISSUE #358: Doorhanger for errors doesn't go away after fixing manifest validation errors
After an invalid manifest fails validation (this saved my @$$, awesome!), the user/developer is notified by the red border around the app AND the yellow door hanger.  If I fix the app and update, the border is removed, but I have to acknowledge the doorhanger for it to go away.  Seems like error information is being presented twice.  Might be nice to pick one.

mykmelez:
This change suppresses the error notification bar when you add an app, since the error also shows up in the validation results.

It doesn't suppress the error notification bar if you ignore the validation results and try to run an invalid app, however. In that case, the bar seems useful to remind you that the app is invalid and explain why your attempt to run it has failed.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -134,12 +134,11 @@ let simulator = module.exports = {
       console.log("Registered App " + JSON.stringify(apps[manifestFile]));
 
       this.updateApp(manifestFile, function next(error, app) {
-        // app reinstall completed
-        // success/error detection and report to the user
+        // Update the Dashboard to reflect changes to the record and run the app
+        // if the update succeeded.  Otherwise, it isn't necessary to notify
+        // the user about the error, as it'll show up in the validation results.
         simulator.sendListApps();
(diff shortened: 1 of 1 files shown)


mykmelez:
@nickdesaulniers: how does this look?

nickdesaulniers:
Looks good!
I don't like that simulator.error appends errors to the notification box.  If you keep getting a bunch of errors, they stack, and continue to stay stacked even if you refresh the page.  What are your thoughts on this?  It might be better to move discussion into its own bug?

mykmelez:
> I don't like that simulator.error appends errors to the notification box. If you keep getting a bunch of errors, they stack, and continue to stay stacked even if you refresh the page. What are your thoughts on this?

Hmm, indeed, that does sound suboptimal.

> It might be better to move discussion into its own bug?

Good idea!

PR #420, opened 2013-04-09
automatically configure prosthesis logging (fixes #378)

### LINKED ISSUE #378: automatically configure prosthesis logging
Configuring prosthesis console logging is cumbersome, as it relies on hardcoded flags in each individual script. Prosthesis should automatically enable logging when run via `make run` and in preview builds, but disable it in release builds.

One way to do this would be to define a pref, r2d2b2g.console.enabled, that scripts use to determine whether or not to log. Then create a prefs file, prosthesis/defaults/preferences/dev.js, that sets it to _true_; and include that file in the XPI when building prosthesis; but ignore it if the Make variable RELEASE is defined, so I can create release packages that don't log via `RELEASE=1 make && make package`.

Another option would be to send messages to the Firefox process instead of dumping them in the Simulator process. Then the Simulator addon can console.log them, which will automatically do the right thing (since the Simulator addon uses the Add-on SDK, which adjusts the log level for its console object based on how the addon was installed).

mykmelez:
This changeset automatically configures prosthesis logging, so prosthesis logs when run on a developer machine (via `make run`, which invokes `cfx run`) but doesn't log when run on a user machine (via installation of a package created by `cfx xpi`).

Primarily, it changes the function used to log B2G stdout from _dump()_ to _console.log()_, which then handles the decision about whether or not to log. But the changeset makes several other changes along the way.
1. Unconditionally enables logging in the simulator actor, so its logging behaves consistently.
2. Removes extra info from prosthesis log messages, like filename and line number. I understand that these can be useful, but they're inconsistent with the messages generated by all other code that logs, including core B2G code, the debugger server (and its actors), and all the modules, including the core SDK modules as well as the main addon's modules. (If it's important to record this information for a particular message, we can still make it do so by using _Cu.reportError()_ or the equivalent.)
3. Moves the logging from Simulator to RemoteSimulatorClient, as this seems more like an internal function of RemoteSimulatorClient.
4. Makes logging in adb.js be consistent with logging elsewhere.
5. Fixes a couple of minor style nits.
6. removes the unused _logStdout_ request type.

### DIFF
--- a/addon/lib/adb.js
+++ b/addon/lib/adb.js
@@ -49,7 +49,11 @@ if (!COMMONJS) {
 }
 
 function debug(aStr) {
-  dump("--*-- ADB.jsm: " + aStr + "\n");
+  if (COMMONJS) {
+    console.log("adb: " + aStr);
+  } else {
+    dump("--*-- ADB.jsm: " + aStr + "\n");
+  }
 }
 
 let ready = false;
@@ -273,7 +277,7 @@ this.ADB = {
       let data = aEvent.data;
       debug("length=" + data.length);
       let dec = new TextDecoder();
-      debug(dec.decode(data));
+      debug(dec.decode(data).trim());
 
       // check the OKAY or FAIL on first packet.
       if (waitForFirst) {
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -46,7 +46,7 @@ const RemoteSimulatorClient = Class({
     //       will be called only once)
     //off(this);
 
-    // on pinbackTimeout, emit an high level "timeout" event
+    // on pingbackTimeout, emit a high level "timeout" event
     // and kill the stalled instance
     this.once("pingbackTimeout", function() {
       this._pingbackCompleted = false;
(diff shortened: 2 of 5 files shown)


mykmelez:
@rpl: how does this look to you? I know the second change might be disappointing, but it makes these messages consistent with all messages logged by all other Mozilla code everywhere, and I think it's worth being consistent in this case.

rpl:
@mykmelez it looks good to me
don't worry for the "removing extra logging info", we know how to add that extra info if necessary ;-)

I found just a small nit:
- we can change "JSON.stringify(value, null, 2)" to JSON.stringify(value) because when we print it
  using console.log it will be printed into a single line of log:
  - https://github.com/mozilla/r2d2b2g/blob/master/prosthesis/content/window-manager.js#L119
  - https://github.com/mozilla/r2d2b2g/blob/master/prosthesis/content/window-manager.js#L177

mykmelez:
> I found just a small nit:
> 
>    we can change "JSON.stringify(value, null, 2)" to JSON.stringify(value) because when we print it using console.log it will be printed into a single line of log:
>        https://github.com/mozilla/r2d2b2g/blob/master/prosthesis/content/window-manager.js#L119
>        https://github.com/mozilla/r2d2b2g/blob/master/prosthesis/content/window-manager.js#L177

Actually, console.log() respects the whitespace in the message (except for leading and trailing space, which it trims in order to avoid a blank line in the output from a trailing newline). Here's what one of those messages looks like in the Error Console and on the command line:

```
info: r2d2b2g: appWindowObserver [
  {
    "target": {
      "location": {}
    },
    "type": "summary",
    "added": [
      {
        "__mutation_summary_node_map_id__": 6799,
        "reallyIsBrowserOrApp": true,
        "reallyIsApp": true,
        "isExpectingSystemMessage": false,
        "appManifestURL": "app://costcontrol.gaiamobile.org/manifest.webapp"
      }
    ],
    "removed": [],
    "reparented": []
  }
]
```

So I'll leave it in when landing the changeset. Thanks for the review!

PR #425, opened 2013-04-11
FIX: debug helper missing in runApp runnable

rpl:
during debugging helpers cleanup we broke runApp request handler:
runApp command handler spawns a runnable and it silently fails
because this.debug is missing.


mykmelez:
No need to apologize, I'm the one at fault!

In any case, thanks for finding and fixing it quickly. This looks fine, although I might prefer to bind each function that uses _this.debug_ instead of binding _this.debug_ itself.

Issue #422, opened 2013-04-10
reload app in single step from within Simulator


mykmelez:
It should be possible to reload an app in a single step from within the Simulator after making some changes to it, just like it's possible to reload a web page in the browser via View > Reload and the Ctrl/Cmd-R keyboard shortcut.

For a hosted app, this would just reinstall and then restart the app. For a packaged app, it would first repackage the app before reinstalling and restarting it.

The interface to the feature can be identical to the one in the browser: a View > Reload menuitem paired with a Ctrl/Cmd-R keyboard shortcut. And the implementation can use the remote debugging channel to trigger the reload, sending a message to the client (i.e. the Firefox addon) requesting that it do the same thing it does when the developer presses the Update button for an app in the Dashboard.

Then the developer will be able to go directly back and forth between their code editor and the Simulator without having to visit the Dashboard each time they make a change.

nickdesaulniers:
+1

rpl:
+1 +prototype :-)

nickdesaulniers:
Hey, awesome to see this land in 3.0pre6!  What a nice feature!

One of the things I'm currently working on is using this [new tool](https://marketplace-dev.allizom.org/developers/test/receipts/) to install a hosted app with a valid verifiable receipt.  This requires me to host my app I'm developing behind a domain.  While I develop, I have to push my changes to my remote machine.  Then I have to close down the app and restart it.  The command+r hotkey doesn't work because hosted apps installed from the browser in the sim aren't registered in the dashboard.

I figured I'd share what I was trying to do, so we could together discuss possible solutions.

mykmelez:
@nickdesaulniers I filed issue #443 on a solution to your problem. Please comment there about whether or not it would solve your problem!

PR #428, opened 2013-04-12
FEATURE: prosthesis application update shortcut as described by #422

rpl:
First prototype of application update shortcut from b2g-desktop instance
as described by #422.

Next steps:
- menu and menu item "View -> Reload"
- tweaks, cleanup and test on hosted apps

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -1041,6 +1041,24 @@ let simulator = module.exports = {
 
     let simulator = this;
     remoteSimulator = new RemoteSimulatorClient({
+      appUpdateHandler: function(appId) {
+        console.log("DO UPDATE", appId);
+        let foundAppKey = null;
+        Object.keys(simulator.apps).forEach(function (key) {
+          if (simulator.apps[key].xkey === appId) {
+            foundAppKey = key;
+          }
+        });
+        simulator.updateApp(foundAppKey, function next(error, app) {
+          simulator.sendListApps();
+          // success/error detection and report to the user
+          if (error) {
+            simulator.error(error);
+          } else {
+            simulator.runApp(app);
+          }
+        });
+      },
       onReady: function () {
         simulator.postIsRunning();
       },
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -1042,19 +1042,34 @@ let simulator = module.exports = {
     let simulator = this;
     remoteSimulator = new RemoteSimulatorClient({
       appUpdateHandler: function(appId) {
-        console.log("DO UPDATE", appId);
+        console.log("handle requested appUpdateRequest", appId);
         let foundAppKey = null;
         Object.keys(simulator.apps).forEach(function (key) {
           if (simulator.apps[key].xkey === appId) {
             foundAppKey = key;
           }
         });
+
+        if (!foundAppKey) {
+          simulator.remoteSimulator.showNotification("App not updated (not found)");
+        }
+
         simulator.updateApp(foundAppKey, function next(error, app) {
           simulator.sendListApps();
           // success/error detection and report to the user
           if (error) {
-            simulator.error(error);
+            simulator.remoteSimulator.showNotification(error);
           } else {
+            simulator.remoteSimulator.showNotification("App updated: " + app.name);
+            let validationMessage = "validation result: ";
+            if (app.validation.errors.length > 0) {
+              validationMessage += "INVALID";
+            } else if (app.validation.warnings.length > 0) {
+              validationMessage += "WARNINGS";
+            } else {
+                validationMessage += "OK";
+            }
+            simulator.remoteSimulator.showNotification(validationMessage);
             simulator.runApp(app);
           }
         });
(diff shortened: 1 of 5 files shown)


mykmelez:
Oooh, this is awesome!

addon/lib/simulator.js
```diff
@@ -1041,6 +1041,24 @@ let simulator = module.exports = {
...
     remoteSimulator = new RemoteSimulatorClient({
+      appUpdateHandler: function(appId) {
+        console.log("DO UPDATE", appId);
+        let foundAppKey = null;
+        Object.keys(simulator.apps).forEach(function (key) {
+          if (simulator.apps[key].xkey === appId) {
+            foundAppKey = key;
+          }
+        });
+        simulator.updateApp(foundAppKey, function next(error, app) {
```
mykmelez, line 1057: Does this intentionally call _simulator.updateApp()_ with a _null_ key if the app isn't found in order to display an error message to the developer? If so, that seems like a good thing, but it'd be worth noting it in a comment, since it isn't obvious from looking at the code. (My first thought, when I read this, was to ask why the code never checks to see if it actually found the key.) Also, I don't actually see any error in that case.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -17,6 +17,12 @@ let SimulatorActor = function SimulatorActor(aConnection) {
   this._listeners = {};
   this.clientReady = false;
 
+  this._observers = {};
+
+  let obs = Services.obs.addObserver(this.appUpdateObserver.bind(this),
+                                     "r2d2b2g-app-update", false);
+  this._observers["r2d2b2g-app-update"] = obs;
```
mykmelez: Instead of adding individual functions as observers and then having to cache them, you can simply add the _this_ object as the observer. Then, since it's an object rather than a function, the observer service will treat it as an [nsIObserver](http://mxr.mozilla.org/mozilla-central/source/xpcom/ds/nsIObserver.idl) and call its _observe()_ method, which you can then implement as a function that itself calls _this.appUpdateObserver()_.

That's a common pattern in Mozilla code. Here's an example:

https://github.com/mozilla/mozilla-central/blob/cfe5af8799af86659d5f9926fac5572f1c083c3d/browser/metro/base/content/browser-ui.js#L558-L580

prosthesis/content/shell.xul
```diff
@@ -51,4 +51,8 @@
     </toolbar>
   </toolbox>
 
+  <keyset>
+    <key id="app-update" modifiers="control" key="R"
```
mykmelez: Specify the _accel_ modifier instead of _control_, as it binds the appropriate platform-specific accelerator (_control_ on Linux/Windows, _command_ on Mac).

rpl:
Applied changes:
- change modifier to accel
- show notification on b2g-desktop for "app not found" and "app updated" and "validation result"
- added observe method to the SimulatorActor

Missing pieces from #422 
- menu and menu item "View -> Reload"

@mykmelez how it looks now?

I think we could add the menu bar in another pull request to fix some additional problems:
- menu styling (e.g. in my gnome desktop it uses grey text on black background which is totally unreadable)  
- recalculate window height (it will contain toolbox and menubar heights)

mykmelez:
> Applied changes:
> 
> ```
> change modifier to accel
> show notification on b2g-desktop for "app not found" and "app updated" and "validation result"
> added observe method to the SimulatorActor
> ```
> 
> Missing pieces from #422
> 
> ```
> menu and menu item "View -> Reload"
> ```
> 
> @mykmelez how it looks now?

Overall, it looks good, and the shortcut is now Cmd-R on Mac! But the "app updated" notification is redundant, because Gaia already pops up a notification that the app was installed. And the validation notification, while useful, seems like it'll become annoying over time, since it appears every time you reload the app, which you frequently do while developing it.

(A minor issue is that I'd prefer to avoid using Gaia's notifications tray for Simulator-related notifications, as it conflates Gaia notifications, which are part of the simulation, with Simulator notifications, which are external to the simulation, i.e. not something you would ever see on a real device.)

So I suggest we only ever display a validation notification when there was an error, i.e. a problem that prevented us from reloading the app. Then, in the future, we might find a less obtrusive way to display warnings.

> I think we could add the menu bar in another pull request to fix some additional problems:
> 
> ```
> menu styling (e.g. in my gnome desktop it uses grey text on black background which is totally unreadable)
> recalculate window height (it will contain toolbox and menubar heights)
> ```

Ok. Regarding the style, I notice that the toolbar has the same problem.

rpl:
@mykmelez it seems reasonable (e.g. we might notify only validation result changes)

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -51,8 +53,30 @@ let SimulatorActor = function SimulatorActor(aConnection) {
...
+      appId: message.wrappedJSObject.appId
+    });
+  },
+
   disconnect: function() {
     this.debug("simulator actor connection closed");
+    Object.keys(this._observers).forEach(function(topic) {
+      let obs = this._observers[topic];
+      Services.obs.removeObserver(obs, topic);
+    });
```
mykmelez, line 79: Nit: this is no longer needed.

mykmelez:
Pushed with the nit fixed! (Along with two others I noticed while fixing that one.)

PR #434, opened 2013-04-17
add a menu item for new "update app" shortcut

mykmelez:
@rpl: Here's a mostly complete implementation of a basic menubar that adds an item for the new "update app" shortcut. It still lacks style on Linux, and I haven't fully tested with orientation changes, which might need to take the size of the menubar into account.

### DIFF
--- a/prosthesis/chrome.manifest
+++ b/prosthesis/chrome.manifest
@@ -12,5 +12,9 @@ skin prosthesis skin skin/
 # our own style (including style from the B2G sheet that we end up wanting).
 override chrome://browser/content/content.css chrome://prosthesis/content/global.css
 
+overlay chrome://prosthesis/content/shell.xul chrome://prosthesis/content/shell-mac.xul os=Darwin
+overlay chrome://prosthesis/content/shell.xul chrome://prosthesis/content/shell-windows.xul os=WINNT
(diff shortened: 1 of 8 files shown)


rpl:
@mykmelez it looks fine and works correctly, but I've to test window height issue (if any) more deeply and find a reason/workaround for the styling issues on linux :-( 
I'm attaching a screenshot as reference:
rendering_issue1

mykmelez:
@rpl I figured out the style issue: B2G disables native theming via the pref _mozilla.widget.disable-native-theme_. Reenabling it got things working again.

And as for window height, I'm not noticing any problems, at least not anymore. I do see a bit of scrunching of the button icons on Linux. I suspect that's because we apply some style to them in shell.css that they probably don't need anymore now that we've enabled native styling.

rpl:
@mykmelez awesome! it works and looks good now

mykmelez:
After a bit of tweaking, the buttons are no longer scrunched on Linux, and they look good on Windows and Mac as well.

The only issue I still see is that the color of the menuitem labels is incorrect on the Linux distribution I've been using for testing. It's black, on a dark background, whereas it should be white. The color of the menuitem labels in Firefox itself is correct, so it doesn't appear to be a bug in the platform itself (unless it's specific to the 18/b2g18 branch).

But that's a relatively minor issue, and I'm not going to hold up landing this change for it. Thanks for the review and testing!

Issue #356, opened 2013-03-07
Geolocation: Default to custom, not current, coordinates


nickdesaulniers:
This will help the addon pass review, the user/developer has to opt in to sharing their geolocation coordinates with the addon.

mykmelez:
Did some thinking on the way home, and here are my thoughts on implementation (for both this and issue #357):

Instead of passing it back to the shell, the geolocation dialog persists the type of geolocation to provide (default, current, or custom) in a preference, r2d2b2g.geolocation.type. It also stores a custom geolocation in two preferences, r2d2b2g.geolocation.custom.latitude and r2d2b2g.geolocation.custom.longitude (note: prefs can't contain floats, so the values must be stored as strings). The [&lt;prefwindow> XUL tag](https://developer.mozilla.org/en-US/docs/XUL/prefwindow), and its associated machinery for reflecting XUL form controls into prefs, is probably the best way to do this.

When FakeGeoPositionProvider is instantiated, it checks the value of r2d2b2g.geolocation.type. If the value is "default" or "custom", it does nothing (besides setting a pref observer; see below); but when its watch() method is called, it updates the callback with the default or custom geolocation. If the value is "current", however, it notifies "r2d2b2g-geolocation-observe".

Meanwhile, SimulatorActor has registered a listener for "r2d2b2g-geolocation-observe", and when it observes the notification, it requests the current geolocation from Firefox. Firefox then uses watchPosition to watch the geolocation and notifies SimulatorActor each time it changes; whereupon SimulatorActor notifies "r2d2b2g-geolocation-change"; which FakeGeoPositionProvider observes in order to update its callback.

Finally, when FakeGeoPositionProvider is instantiated, in addition to checking the value of r2d2b2g.geolocation.type, it sets a pref observer on the r2d2b2g.geolocation.\* branch. When one of the prefs change, it updates its callback with the new geolocation. If r2d2b2g.geolocation.type changes to "current", then it notifies "r2d2b2g-geolocation-observe" to start observing current geolocation changes. If, however, it changes _from_ "current" to "default" or "custom", then it notifies "r2d2b2g-geolocation-ignore", which SimulatorActor observes in order to tell Firefox to stop watching the geolocation.

So Firefox only watches the geolocation while the pref is set to "current" _and_ the fake provider is instantiated.

As a beneficial side-effect, the value of the prefs persist across sessions, so users who set a custom geolocation don't have to reset it after restarting the Simulator.

[...]

PR #436, opened 2013-04-17
FIX: geolocation setup sequence

rpl:
Currently we've 3 sub-issues:
- on nightly, 23.0a1 (2013-04-16), Ci.nsIDOMGeoGeolocation doesn't exist
  (I'm not sure if it could be a temporary problem or a permanent change for
  the future releases)
- if firefox doesn't send a geolocation response, click the button doesn't open
  the geolocation dialog and we can't set a custom location
- our "muted actor due to unsolicited event" workaround doesn't work
  (packets blocked in queue until another unsolicited event will be sent, 
  e.g. appUpdateRequest)

This pull request introduce small changes needed to get simulated geolocation
to work again:
- register geolocation button click listener on start
  to be able to use custom location even if current location response
  will never be received
- remove unsolicited event workaround
- new geolocation setup workflow
  geolocationReady (FIREFOX CLIENT -> B2G ACTOR)
  r2d2b2g:geolocation-ready (B2G ACTOR -> B2G SHELL)
  r2d2b2g:geolocation-update (B2G SHELL -> B2G ACTOR)
  geolocationRequest (B2G ACTOR -> FIREFOX CLIENT)
  geolocationResponse (FIREFOX CLIENT -> B2G ACTOR)

### DIFF
--- a/prosthesis/components/FakeGeolocationProvider.js
+++ b/prosthesis/components/FakeGeolocationProvider.js
@@ -50,7 +50,7 @@ function FakeGeoPositionProvider() {
       this.callback.update(new FakeGeoPositionObject(
         message.wrappedJSObject.lat, message.wrappedJSObject.lon));
     }
-  }).bind(this), "r2d2b2g-geolocation-response", false);
+  }).bind(this), "r2d2b2g:geolocation-response", false);
 }
 
 FakeGeoPositionProvider.prototype = {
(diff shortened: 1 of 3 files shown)


rpl:
@mykmelez @nickdesaulniers I noticed that the workaround that we have introduced doesn't really work :-( 
how it looks with the tweaks proposed in this pull request?

mykmelez:
Great fix, thanks @rpl!

Note that I described a new workflow in issue #356. I'm not sure if that's still relevant, given these changes.

In a separate issue, we should also figure out where nsIDOMGeoGeolocation has gone.

nickdesaulniers:
Is the first issue listed still a problem?  Someone was touching the gelocation code earlier this month, but it [still seems to be using](http://mxr.mozilla.org/mozilla-central/source/dom/src/geolocation/nsGeolocation.h#183) that interface.

mykmelez:
It's still a problem, but I fixed it in #438. Sorry, I should have mentioned that here!

PR #437, opened 2013-04-17
wait until Simulator process dead before killing ADB (fixes #435)

### LINKED ISSUE #435: "firefox quit unexpectedly." or pegs CPU on Firefox quit while Simulator running
When I quit FirefoxNightly while the Simulator is running on my Mac, one of two things seems to be happening fairly consistently:
- the OS informs me that "firefox quit unexpectedly";
- a "firefox" process starts pegging a CPU.

I wonder if this is related to the way we use the _subprocess_ module to kill the Simulator process on FirefoxNightly shutdown. Although I only started seeing it recently. Perhaps it has something to do with my recent change to start using _subprocess_ to start the ADB process.

@ochameau: do you have any thoughts?

mykmelez:


mykmelez:
@ochameau How does this look to you?

mykmelez:
I'm going to merge this to get it into today's preview build. But I'm still very interested in knowing your thoughts!

ochameau:
I tend to think that ADB.kill won't be called, onless Simulator.kill is synchronous.
When addon are unloaded on shutdown, we notify all of them about the shutdown,
but they will basically be able to just run of sequence of code in this event loop.
On the next loop iteration, if no platform code used enterLastWindowClosingSurvivalArea, Firefox may already be exited.

There is no API in jetpack to control enterLastWindowClosingSurvivalArea, as it is a bit scary, because addons would easily be able to prevent firefox closing.
But we can give it a try here?

See this code for usage:
http://mxr.mozilla.org/mozilla-central/source/toolkit/content/contentAreaUtils.js#1102

mykmelez:
> I tend to think that ADB.kill won't be called, onless Simulator.kill is synchronous.

I thought that too. But then I tested it and found that ADB is consistently killed after I quit Firefox, even though _simulator.kill()_ is asynchronous.

And if I instrument the code, I find that _simulator.kill()_ does get called at least sometimes. Here's the simple instrumentation I did:

```
console.log("before asking to kill simulator process");
this.kill(function() {
  dump("simulator process killed\n");

  // We wait until the Simulator process is dead before starting the process
  // that kills the ADB server because Firefox was crashing or hanging when
  // we did those two things simultaneously (issue #435).
  ADB.kill(Runtime.OS == "WINNT" ? true : false /* sync */);
});
console.log("after asking to kill simulator process");
```

And here are the two varieties of logs I get (sometimes I see one, sometimes the other):

```
info: r2d2b2g: before asking to kill simulator process
info: r2d2b2g: after asking to kill simulator process
simulator process killed
Total time: 10.892656 seconds
Program terminated successfully.

info: r2d2b2g: before asking to kill simulator process
info: r2d2b2g: after asking to kill simulator process
Total time: 10.356187 seconds
Program terminated successfully.
```

In the second log, it looks like the callback never got called, since "simulator process killed" doesn't appear. Nevertheless, ADB does get killed! And it doesn't get killed if I comment out the call to _ADB.kill()_. I'm not sure how to interpret that, but it looks like the code works as intended, even if it doesn't work as expected. :-/

Issue #166, opened 2012-11-30
Translate mouse into touch events


digitarald:
B2G doesn't fire touch events currently.

Mouse interaction should fire touch events (only single-touch probably).

Optionally mouse events should not reach B2G (default: prevent events), to properly simulate phone behavior (the spec allows to fire both, though all implementations I saw never fire mouse events for touch).

Feedback?

mykmelez:
Erm, I lost track of this, but your suggestion makes perfect sense to me, as it would much more closely simulate a phone environment!

nhoizey:
Is there any progress on this? It's not possible currently to test touch events without replacing them all by mouse events.

Modernizr sees it as touch enabled.

digitarald:
I took a first stab at it @ https://github.com/digitarald/r2d2b2g/tree/feature-touchme

endemic:
This bug bit me when submitting an app to the Firefox Marketplace. Just leaving a note here instead of opening a new issue.

peregrinogris:
What's the status of the patch for this? I faced this issue when trying to use a library that supports both mobile and desktop, and it fails when detecting if running on mobile or desktop.
I tried to monkey-patch the library to get it working at least on my app, but I couldn't find a way to know if it's running on the FirefoxOS Simulator or the real thing. Is there a way to detect when running on the Simulator?

elgerlambert:
Just wanted to drop a line to say this is an issue for me also. I've built a mobile web app which uses Zepto to bind to touch events (I don't bind to click events at all).

Since this is my first comment related to Firefox OS, just wanted to say I love the project and wish it luck aswell!

Kind regards

mykmelez:
[Bug 751147](https://bugzilla.mozilla.org/show_bug.cgi?id=751147) removed something similar from B2G a while back. Perhaps we can reuse that code.

mykmelez:
Given the issues in the pull request, it feels scary to take this at the end of a development cycle. Even though it's very important, I'd rather land it at the beginning of the next cycle. So I've removed its milestone/priority settings pending reprioritization for the next milestone.

mykmelez:
Fixed by #510!

PR #439, opened 2013-04-18
Enabling Touch Events

digitarald:
1st take on fixing #166

touch.js from http://mxr.mozilla.org/mozilla-esr17/source/b2g/chrome/content/touch.js?rev=df7835caf208

Testable on
- http://paulirish.com/demo/multi
- http://openlayers.org/dev/examples/multitouch.html
- http://phoboslab.org/crap/mt.html
- google "test multi touch"

### DIFF
--- a/prosthesis/content/touch.js
+++ b/prosthesis/content/touch.js
@@ -1,0 +1,202 @@
+
+(function touchEventHandler() {
+  let debugging = false;
+  function debug(str) {
+    if (debugging)
+      dump(str + '\n');
+  };
+
+  let contextMenuTimeout = 0;
+
+  // This guard is used to not re-enter the events processing loop for
+  // self dispatched events
+  let ignoreEvents = false;
+
+  // During a 'touchstart' and the first 'touchmove' mouse events can be
+  // prevented for the current touch sequence.
+  let canPreventMouseEvents = false;
+
+  // Used to track the first mousemove and to cancel click dispatc if it's not
+  // true.
+  let isNewTouchAction = false;
+
+  // If this is set to true all mouse events will be cancelled by calling
+  // both evt.preventDefault() and evt.stopPropagation().
+  // This will not prevent a contextmenu event to be fired.
+  // This can be turned on if canPreventMouseEvents is true and the consumer
+  // application call evt.preventDefault();
+  let preventMouseEvents = false;
+
+  let TouchEventHandler = {
+    events: ['mousedown', 'mousemove', 'mouseup', 'click', 'unload'],
+    start: function teh_start() {
+      this.events.forEach((function(evt) {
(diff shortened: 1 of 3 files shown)


prosthesis/content/touch.js
```diff
@@ -0,0 +1,202 @@
+
```
mykmelez, line 1: Nit: this needs a license header.

prosthesis/content/touch.js
```diff
@@ -0,0 +1,202 @@
...
+  // If this is set to true all mouse events will be cancelled by calling
+  // both evt.preventDefault() and evt.stopPropagation().
+  // This will not prevent a contextmenu event to be fired.
+  // This can be turned on if canPreventMouseEvents is true and the consumer
+  // application call evt.preventDefault();
+  let preventMouseEvents = false;
+
+  let TouchEventHandler = {
+    events: ['mousedown', 'mousemove', 'mouseup', 'click', 'unload'],
+    start: function teh_start() {
```
mykmelez, line 32: I have a hard time not reading this as lolspeak! ;-)

prosthesis/content/touch.js
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            debug('click: fire');
+          }
+          return;
+      }
+
+      let target = eventTarget || this.target;
+      if (target && type) {
+        console.log(type);
```
mykmelez, line 131: This _console.log()_ call and the one on line 179 should be debug calls instead, as they overwhelm the Error Console with messages. The one on line 135, on the other hand, should be a Cu.reportError().

mykmelez:
> Testable on
> 
> ```
> http://paulirish.com/demo/multi
> http://openlayers.org/dev/examples/multitouch.html
> http://phoboslab.org/crap/mt.html
> ```

They work!

And I see some improvement in Settings. If I mousedown and then drag really slowly for some short period of time, I get into a mode in which I can drag ad infinitum, including quickly.

But if I mousedown and then drag quickly (or even at just a moderate pace), then I see the same problem: the app (or the platform?) starts interpreting the drag as a mouse drag-drop on an image or selection.

And there's something weird going on in Windows 8. There, mousedown and drag on the homescreen accelerates rapidly at some point, no matter how slowly I drag. So I'll drag part of the next page in view and then whoosh! The page scoots the rest of the way in a moment.

So there's some funkiness here. Perhaps the simulation doesn't capture and suppress all mouse events before the platform takes a crack at them, which is why something interprets a rapid movement in the Settings app as a click-drag? And maybe Windows 8 messes with whatever calculation the simulation does to determine how fast to say the finger is moving? (I'm obviously speculating here.)

ochameau:
PR #510 should fix this issue by sharing touch event shim shipped in gaia.

mykmelez:
Fixed by #510!

Issue #357, opened 2013-03-07
Geolocation: Do not fetch the current location until explicitly shared by user


nickdesaulniers:

nickdesaulniers:
Fixed in #446

PR #446, opened 2013-04-18
default to custom, not current coordinaates Issue #356

nickdesaulniers:
Myk, I wanted to bang out Issue #356.  This patch will default to custom coordinates (0, 0) instead of current coordinates.  I want to finish this up in Issue #357, by not actually getting the current position until it's requested from the Firefox process.  If you agree and this patch looks good, please close Issue #356 after merging this.

### DIFF
--- a/prosthesis/content/geolocation.js
+++ b/prosthesis/content/geolocation.js
@@ -8,26 +8,25 @@ window.addEventListener("DOMContentLoaded", function() {
       custom = $("custom"),
       latitudeEle = $("latitude"),
       longitudeEle = $("longitude"),
-      inputOutput = window.arguments[0].wrappedJSObject,
-      input = inputOutput.input,
-      output = inputOutput.output,
-      latitude = input.lat,
-      longitude = input.lon,
-      tempUseCurrent = input.useCurrent,
+      windowParams = window.arguments[0].wrappedJSObject,
+      latitude = windowParams.lat,
+      longitude = windowParams.lon,
+      tempUseCurrent = windowParams.useCurrent,
       setTextBoxes = function(disabled) {
         latitudeEle.disabled = disabled;
         longitudeEle.disabled = disabled;
         tempUseCurrent = disabled;
       },
       accept = function () {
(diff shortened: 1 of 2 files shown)


mykmelez:
> Myk, I wanted to bang out Issue #356. This patch will default to custom coordinates (0, 0) instead of current coordinates. I want to finish this up in Issue #357, by not actually getting the current position until it's requested from the Firefox process. If you agree and this patch looks good, please close Issue #356 after merging this.

That's fine, I'll just add #357 to v3 then.

The first time I tested this patch, I saw some weird behavior: when I pressed the Ok or Cancel button in the dialog, it reappeared, and I had to press the button again to dismiss the dialog. Also, the second time the dialog appeared, the custom coordinates had been reset to 0x0.

The second time I tested, I didn't see this behavior. However, both times I saw some other weird behavior in the Mykzilla app I use to test (http://www.mykzilla.org/app/manifest.webapp). Its "geolocate" button implementation is very simple:

```
function geolocate() {
  navigator.geolocation.getCurrentPosition(function(position) {
    alert("Your position is " + position.coords.latitude + " x " + position.coords.longitude + ".");
  });
}
```

When I press the button after changing the location, however, it first alerts me with the old location. Then, a moment later, and before I have dismissed the first alert, it alerts me with the new location. I.e. I get two alerts, one after the other, in quick succession. Of which only the second one is correct.

nickdesaulniers:
I am unable to reproduce the first error listed.

I was able to reproduce the second error with my patch applied, and with 3.0pre6 (no patch applied).  This is a new bug, not introduced by my patch.

I'd be happy to look into it as part of #357.

mykmelez:
> I was able to reproduce the second error with my patch applied, and with 3.0pre6 (no patch applied). This is a new bug, not introduced by my patch.

Ah, ok. I wonder if @rpl's geolocation improvements might have introduced a regression. Or perhaps it's something that has been with us the whole time.

> I'd be happy to look into it as part of #357.

Ok, cool! Merging…

Issue #442, opened 2013-04-18
add make clean target


mykmelez:
We should have a `make clean` target that removes downloaded and generated files, including:
- B2G package and bindir (addon/data/PLATFORM/b2g/)
- ADB package and bindir (addon/data/PLATFORM/adb/)
- Gaia profile and associated files (addon/template)
- gaia/build/custom-prefs.js
- perhaps also a `make clean` in the gaia/ subdirectory

nickdesaulniers:
Note: B2G package and bindir (`addon/data/PLATFORM/b2g/`) on OSX is `addon/data/PLATFORM/B2G.app/`

nickdesaulniers:
Fixed in #447

PR #447, opened 2013-04-19
add make clean Issue #442

nickdesaulniers:
I did not try `perhaps also a make clean in the gaia/ subdirectory` because I don't know how to invoke make recursively.  Is it as simple as:

```
cd gaia
make clean
cd -
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,3 +182,8 @@ help:
 	@echo '  run: start Firefox with the addon installed into a new profile'
 	@echo '  package: package the addon into a XPI'
 	@echo '  help: show this message'
+
+clean:
+	rm -rf addon/data/$(B2G_PLATFORM)/*
+	rm -rf addon/template/*
+	rm gaia/build/custom-prefs.js
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build profile prosthesis b2g adb locales run package help
+.PHONY: build profile prosthesis b2g adb locales run package clean help
 
 -include local.mk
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -182,3 +182,8 @@ help:
 	@echo '  run: start Firefox with the addon installed into a new profile'
 	@echo '  package: package the addon into a XPI'
 	@echo '  help: show this message'
+
+clean:
+	rm -rf addon/data/$(B2G_PLATFORM)/*
+	rm -rf addon/template/*
```
mykmelez: Make these simply:

```
rm -rf addon/data/$B2G_PLATFORM
rm -rf addon/template
```

Then you'll also delete any hidden files (i.e. those beginning with a dot) in those directories as well as the directories themselves.

mykmelez:
> cd gaia
> make clean
> cd -

Each line in a make recipe is invoked from cwd, so this will re-run make in the top-level directory, creating an infinite loop!

Since a trailing backslash escapes a newline, you could do:

```
    cd gaia && \
    make clean
```

(No need to change back to the cwd for the next line; make will do that for you! But note that `cd -` will change to $HOME, which is not what you'd want anyway. You'd want to do `cd ..` if make didn't do it for you.)

But this sequence of commands is so short that I would instead put it all on one line:

```
cd gaia && make clean
```

However, in this case you can simplify further via the make -C flag, which changes to the specified directory before making:

```
make -C gaia clean
```

Two more things to remove:
- $(ADB_PACKAGE)
- the B2G package; but only make-b2g.py knows its name, so we'd need to make it do this (f.e. given a --clean flag); we can make this a followup :-)

Also, make sure to add `clean` to the list of .PHONY targets at the top of the file and a description of it to the `help` recipe at the bottom of the file. Finally, a minor nit: this target would be best positioned either below the `build` target or above the `help` target.

mykmelez:
Looks great! My only nittiest of nits is that I'd like to keep the phony target list ordered in the same order as the targets appear in the file. I'll fix that nit while merging this change. Thanks for the fix! It'll be very useful, especially the next time I make preview and release builds! :-)

Issue #238, opened 2012-12-27
improve Dashboard usability


mykmelez:
@digitarald has mocked up a number of reasonable changes to the Dashboard that improve its usability (including but not limited to #187). I don't have the mockups at the moment, but @digitarald should be able to provide them, and we should implement them.

potch:
I'm going to do some dashboard front-end enhancements that will hopefully fall under the purview of this issue.

digitarald:
The first page contains ideas for splitting "add app" into the various app typed and mental models with better guidance: https://dl.dropboxusercontent.com/u/1864007/Mozilla/17-10-12%20-%20Simulator.pdf

PR #449, opened 2013-04-19
Use templates for dashboard

potch:
Initial ground work for #238.

In this PR:
- A new JS file, applist.js that contains the app list/management code
- Use event delegation for App management
- Moved jQuery DOM code to nunjucks templating
- Use CSS transitions for device-dependent UI
- Removed unused modernizr


addon/data/content/index.html
```diff
@@ -123,8 +122,78 @@ <h5 id="device-status" style="opacity: 0">
...
+                    {% set errors = validation.errors %}
+                    {% set warning = validation.warnings %}
+                    <div class="app-validation">
+                        <span class="app-validation-result">Validation Result:
+                            {% if validation.running %}
+                                RUNNING
+                            {% else %}
+                                {% if errors.length %}
+                                    INVALID
+                                {% else %}
```
tofumatt: No else if in this template parser I'm guessing? :smile:

tofumatt: Oh wait, it's nunjucks. This could be less nested if you used `else if`!

addon/data/content/js/applist.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+    });
+
+
+    return {
+        'update': update
+    };
+
+})();
```
tofumatt: `\n` EOF?

addon/data/content/index.html
potch: Yeah, just realized nunjucks has elif! Will update.

addon/data/content/js/applist.js
potch: blergh

addon/data/content/index.html
```diff
@@ -123,8 +122,74 @@ <h5 id="device-status" style="opacity: 0">
...
+                </h4>
+                <p>
+                    <a title="{{ id }}" href="#">Open Location</a> <span>({{ id }})</span>
+                </p>
+                <p>
+                    Last updated: {{ prettyLastUpdate or 'never' }}
+                </p>
+                {% if validation and not removed %}
+                    {% set errors = validation.errors %}
+                    {% set warning = validation.warnings %}
```
mykmelez: This should be `{% set warnings = validation.warnings %}`.

addon/data/content/js/applist.js
```diff
@@ -0,0 +1,102 @@
...
+    if (d < 120) return 'a minute ago';
+    if (d < 3600) return ~~(d/60) + ' minutes ago';
+    d = ~~(d/3600);
+    if (d < 2) return 'an hour ago';
+    if (d < 24) return d + ' hours ago';
+    d = ~~(d/24);
+    if (d < 2) return 'a day ago';
+    if (d < 365) return d + ' days ago';
+    if (d < 365*2) return 'a year ago';
+    return ~~(d / 365) + ' years ago';
```
mykmelez, line 16: It seems like it'd be useful to express months as an intermediate unit between days and years for changes that are more than a month old (for which days > 30 is probably a good enough approximation).

mykmelez:
This should also remove addon/data/content/js/vendor/jquery-1.8.3-min.js, since index.html is the only file that used it!

mykmelez:
Other than those few issues I noted, this looks great and seems like it'll be much more maintainable and enhanceable!

Aside: in general, there's no need to squash commits after you've submitted a pull request, and it's often better not to do so; both for you, as the developer, who doesn't have to do the extra work; and for me, as the reviewer, who can then match individual commits to review comments in a followup review (while always being able to see the changeset as a whole via GitHub's Files Changed view).

Before you've submitted a pull request, on the other hand, you are welcome to squash away at will and your discretion!

addon/data/content/index.html
potch: good catch!

addon/data/content/js/applist.js
potch, line 16: I'll play dumb and say a month is 30 days :)

mykmelez:
Thanks, @potch; merged!

Issue #448, opened 2013-04-19
Error when loading new url into simulator page


jeffgca:
If I open the app manager, then paste a new url into the location bar I see this traceback:

Timestamp: 2013-04-19 8:27:45 AM
Error: r2d2b2g: An exception occurred.
Traceback (most recent call last):
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/tabs/tab-firefox.js", line 69, in _onReady
this._emit(EVENTS.ready.name, this._public);
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/events.js", line 123, in _emit
return this._emitOnObject.apply(this, args);
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/events.js", line 153, in _emitOnObject
listener.apply(targetObj, params);
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/windows/tabs-firefox.js", line 110, in _emitEvent
tabs._emit(type, tab);
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/events.js", line 123, in _emit
return this._emitOnObject.apply(this, args);
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/deprecated/events.js", line 153, in _emitOnObject
listener.apply(targetObj, params);
File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/main.js", line 924, in
simulator.sendListTabs();
File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/main.js", line 587, in simulator.sendListTabs
list: tabs
File "resource://r2d2b2g-at-mozilla-dot-org/addon-sdk/lib/sdk/content/worker.js", line 351, in postMessage
throw new Error(ERR_FROZEN);
Error: The page is currently hidden and can no longer be used until it is visible again.

mykmelez:
Oof, I see that too. We experimented with using _addon-page_ to hide the top chrome, including the location bar, to make the Dashboard look more like part of the application's interface than a web page (à la Add-ons Manager), which would reduce the risk of this happening.

[...]

PR #454, opened 2013-04-22
Issue #448: Disable worker when the Simulator is hidden in history and ensure that we load only one Simulator page.

ochameau:
Followup of #448. I'm not sure how much nullication of worker when the simulator disappear in history is going to help, but I tried to avoid cases where we open simulator page multiple times. It should definitely kill some weirdness if a user was opening mutliples...


mykmelez:
Looks good, and it addresses #448 along with perhaps some other similar issues lurking in the background.

PR #455, opened 2013-04-22
Use San Francisco office position by default instead of 0,0 (in middle of Atlantic ocean!)

ochameau:
Very simple modification. It isn't very handy to end up in middle of an ocean to test a map application ;-)


mykmelez:
Good idea!

Issue #196, opened 2012-12-12
Window shouldn't be resizeable


waldyrious:
Resizing the window results in black background showing up, and there's no way to restore the original size.

OS: Xubuntu 12.04
Firefox 17.0.1
Simulator Version 1.0

However, window resizing should be possible (automatically) in the event of a rotation between portrait and landscape modes. See #195.

mykmelez:
I've seen this too, and it's quite annoying. It should also be fairly easy to fix, although we might have to fix it in the B2G code.

mykmelez:
Adding this to prosthesis/defaults/preferences/prefs.js should fix this, but it doesn't:

```
user_pref("toolkit.defaultChromeFeatures", "chrome,dialog=no,resizable=no,all");
user_pref("dom.disable_window_open_feature.resizable", false);
```

I'm not sure why that is.

mykmelez:
#457 fixed horizontal resizing. Unfortunately, it looks like we can't fix vertical resizing, or at least we haven't found a way to do so. So I'm going to close the issue, as we've gotten as close as it looks like we can get.

PR #457, opened 2013-04-22
Issue #196: Prevent horizontal resize and prevent smaller vertical one.

ochameau:
I also tried toolkit.defaultChromeFeatures and various other things, but there almost nothing works to prevent resize :/
I wasn't able to prevent all kind of resize, but with this simple patch, only one kind of resize is still possible: you can get a bigger vertical size.


mykmelez:
Thanks, that's an improvement!

PR #468, opened 2013-04-25
Don't share current location until current location selected Issue #357

nickdesaulniers:


prosthesis/components/FakeGeolocationProvider.js
```diff
@@ -2,6 +2,19 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* The FakeGeolocationProvider relies on observing r2d2b2g:geolocation-updates
+ * notifications from content/shell.js, when the user changes their custom
+ * coordinates, AND from content/dbg-simulator-actors.js, when the user wants
+ * to use their current coordinates.  The current coordinates are never fetched
+ * until the user has explicitly selected that they want to share them.
+ * shell.js will send a r2d2b2g:geolocation-start notification that is observed
+ * by dbg-simulator-actors.js.  dbg-simulator-actors.js uses "unsolicated"
```
mykmelez: Nit: geolocation-updates -> geolocation-update

Nit: unsolicated -> unsolicited

addon/lib/remote-simulator-client.js
```diff
@@ -91,9 +91,6 @@ const RemoteSimulatorClient = Class({
     this.on("clientReady", function (remote) {
       console.debug("rsc.onClientReady");
       this._remote = remote;
-      this._sendGeolocationReady(function(packet) {
-        console.debug("GEOLOCATION READY REPLY:: "+JSON.stringify(packet, null, 2));
```
mykmelez, line 95: I'm not a fan of SHOUTING in console messages, and there may be too many messages about geolocation in the codebase, but it looks like you've removed them all, which makes it hard to see the code in action while testing the patch.

I would prefer to have at least a couple messages written to the console, perhaps one on the Simulator side on geolocationStart (i.e. when the user selects "current geolocation") and another on the Firefox side on successful determination of the geolocation (although we should not write the actual coordinates to the console, as it isn't necessary, and there might be privacy considerations).

mykmelez:
@nickdesaulniers Just two minor issues, the first of which you've already fixed. Otherwise this looks great!

mykmelez:
Thanks, @nickdesaulniers, looks great!

nickdesaulniers:
http://www.youtube.com/watch?v=QEmbOL3AAEs

PR #471, opened 2013-04-25
match hashes at the end of our contentPage.

potch:
&hellip;otherwise, restored simulator pages may not attach to the worker properly.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -127,7 +127,7 @@ let simulator = module.exports = {
     }
   },
 
-  get contentPage() Self.data.url("content/index.html"),
+  get contentPage() Self.data.url("content/index.html*"),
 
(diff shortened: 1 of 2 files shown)


addon/lib/simulator.js
```diff
@@ -127,7 +127,7 @@ let simulator = module.exports = {
     }
   },
 
-  get contentPage() Self.data.url("content/index.html"),
+  get contentPage() Self.data.url("content/index.html*"),
```
mykmelez: It isn't madness! But _contentPage_ is also used in a Tabs.open call in simulator.js, for which it needs to be the actual URL. So this wildcard character should get added to the URL when it is used in main.js.

mykmelez:
Looks great, thanks @potch!

Issue #312, opened 2013-02-13
Enhance orientation events simulation


past:
Issue #6 is a great first step, but it would be great if we could incorporate into the simulator something like this, as a more rich way to emulate various orientation events:
http://louisremi.github.com/orientation-devtool/

PR #482, opened 2013-04-30
Fix rotate button background when active/inactive.

ochameau:
I thought that rotate button was just broken and it took me some time to understand that rotation is possible only when being in some apps.
Currently the rotate button has inversed colors. Gray for active and black for inactive.
Fixing the color already helps comprehesion, but it may be usefull to also add a tooltip or a warning.

### DIFF
--- a/prosthesis/content/shell.css
+++ b/prosthesis/content/shell.css
@@ -25,11 +25,11 @@ toolbarbutton > .toolbarbutton-text {
 
 #rotateButton {
   list-style-image: url("chrome://prosthesis/content/icons/rotate.png");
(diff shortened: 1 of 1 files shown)


mykmelez:
Nice fix!

>  Fixing the color already helps comprehesion, but it may be usefull to also add a tooltip or a warning.

Yes, that would also be useful. And in the long run we need to make the button simulate actual rotation instead of just orientation changes. So you can rotate the simulation even in an app that doesn't support other orientations and thus test the usability of rotating such apps.

More thoughts on orientation enhancements are in issue #312.

PR #483, opened 2013-04-30
make tests work again and simplify/conventionalize them

mykmelez:
This makes tests work again and also simplifies them to the extent possible and makes them use a conventional structure and style like the more recent tests in the Add-on SDK.


mykmelez:

ochameau:
It looks good!

PR #484, opened 2013-04-30
remove unnecessary getters and imports

mykmelez:
@ochameau this is a simple cleanup fix that removes unnecessary lazy service getters and unused/redundant _Cu.import_ calls.


ochameau:
r+

PR #487, opened 2013-05-02
Use gaia's builtin custom settings and disable lockscreen via a custom setting

ochameau:
I think you started the project before gaia had the custom setting support.
Now, it would be easier to maintain by using it.

Also, we shouldn't call any gaia javascript code.
I removed the code calling homescreen unlock in favor of a custom setting.
By doing so, the user can re-enable the lockscreen and may be break some other features.
I'm open to get lockscreen.js back as soon as we do not call gaia code.

I tried this patch and the lockscreen doesn't show up, as idle API looks broken:

```
Timestamp: 02/05/2013 16:18:03
Error: [Exception... "'JavaScript component does not have a method named: "onactive"' when calling method: [nsIIdleObserver::onactive]"  nsresult: "0x80570030 (NS_ERROR_XPC_JSOBJECT_HAS_NO_FUNCTION_NAMED)"  location: "native frame :: <unknown filename> :: <TOP_LEVEL> :: line 0"  data: no]
```

We can modify it like this:

```
$ git diff lockscreen.js
diff --git a/prosthesis/content/lockscreen.js b/prosthesis/content/lockscreen.js
index 57420ba..90d4447 100644
--- a/prosthesis/content/lockscreen.js
+++ b/prosthesis/content/lockscreen.js
@@ -10,22 +10,6 @@ SettingsListener.observe("lockscreen.enabled", false, functio
     return;
   }

-  let homescreen = document.getElementById("homescreen").contentWindow.wrappedJ
-
-  debug("wait for lockscreen.locked");
-  // keep the lockscreen unlocked
-  SettingsListener.observe("lockscreen.locked", true, function(value) {
-    debug("LOCKSCREEN LOCKED: "+value);
-    try {
-      if (value) {
-        homescreen.LockScreen.unlock(true);
-      }
-    } catch(e) {
-      // keep trying to unlock (preserving lockscreen.locked=true)
-      setTimeout(function () {
-        navigator.mozSettings
-          .createLock().set({'lockscreen.locked': true});
-      }, 200);
-    }
-  });
+  navigator.mozSettings
+          .createLock().set({'lockscreen.enabled': false});
 });
```

It will immediatly toggle lockscreen off as soon as the user tries to switch it on.
I would have liked to print an alert popup, but alert() doesn't seem to work??
But at the end, I thought it would be better to let the user enable lockscreen and play with it if he explicitely requests it.


mykmelez:
>  But at the end, I thought it would be better to let the user enable lockscreen and play with it if he explicitely requests it.

I agree, this makes sense.

> I think you started the project before gaia had the custom setting support.

Gaia might have had custom-settings.json, but that file doesn't let you remove a setting, and I needed to remove ftu.manifestURL, since NOFTU didn't exist. Hopefully we won't have to remove another setting!

PR #488, opened 2013-05-02
Handle connection failure and retry connecting instead of having to maitain a pingback client/server.

ochameau:
I'm aiming to upstream a bunch of simulator code to devtools/firefox.
The pingback is one simulator part that, I think, won't be easy to upstream.
I'm not sure it worth the cost, whereas I can easily imagine submitting a patch to offer a retry-until-timeout feature to DebuggerClient.

I'm not sure this patch will fix any reported issue, but it will definitely simplify b2g desktop startup workflow and eventually avoid some bugs with extremelly strict firewalls.


gaia
```diff
@@ -1 +1 @@
-Subproject commit d6be6b1fbda9c4a6b2719a963b513855f1dd86c0
+Subproject commit 598fcc7d1a94450b7f8d743204ddc37c7aea5467
```
mykmelez: I'm guessing this is accidental! It looks like you didn't do `git submodule update` after pulling upstream changes at some point.

mykmelez: Or perhaps you're testing with a local commit to Gaia? I see this when I try to `git submodule update`:

```
fatal: reference is not a tree: 598fcc7d1a94450b7f8d743204ddc37c7aea5467
Unable to checkout '598fcc7d1a94450b7f8d743204ddc37c7aea5467' in submodule path 'gaia'
```

mykmelez:
The validator reports the following error when I press the Update button for an app in the Dashboard with this change applied:

```
Error updating cached Manifest:

TypeError: this._remote is undefined
```

gaia
ochameau: Yes, it is. It doesn't depend on anything special on gaia side.
I removed it.

ochameau:
I do not see this error with hosted and packaged app... on windows.
I'll give it a try later on linux so see if I can try reproducing it.

mykmelez:
> I do not see this error with hosted and packaged app... on windows.
> I'll give it a try later on linux so see if I can try reproducing it.

Hmm, I did some more testing on my Mac, and I don't see it with a new profile, but I do see it with the existing profile I use for much of my testing.

ochameau:
Oh interesting, then I'll try with an old profile then and get back to you when I did.
Thanks for the extra testing!

gaia
timmi: bb

mykmelez:
I can also now reproduce on a new profile. My steps are:
1. `make run`;
2. Tools > Web Developer > Firefox OS Simulator;
3. press the Simulator button to start the Simulator;
4. press the Add Directory… button and select the manifest.webapp file for Gaia's Template app.

mykmelez:
I did some more testing and even managed to see things work on my old profile, although they broke again in the next session.

When it fails, it seems like the client.connect() callback never gets called, but clientClosed also isn't triggered. Perhaps there's a race in the DebuggerClient code that causes this when the client connects to the server right as the server is starting up?

ochameau:
I tried really hard to reproduce this issue and I've seen it only once :(
I was wondering if I reached some number of maximum pending sockets. but `$ ss -s` tend to say that's not the case. The current code is quite extreme, as it loops without any delay between each try, so it tries to connect **A LOT**. That's something to fix, and may fix the issue your are seeing, but I'd really prefer to know why it fails.

Can you give it a try?
https://github.com/ochameau/r2d2b2g/blob/bbc22a5acd0f3d9b15f7840b99840ce51e1de177/addon/lib/remote-simulator-client.js#L101
`this.connectDebuggerClient();`
->
`setTimeout(this.connectDebuggerClient.bind(this), 250)`;

I've also seen a state where b2g process seems to act like a daemon and gets its `done` subprocess method called immediatly, while the b2g window stay black. And if I ask for the console, only the console appear.
Same thing here, I'm no longer able to reproduce.

ochameau:
Ok, I found a simple race and pushed a new changeset to fix it.
Your scenario reproduce this race only on slow hardware and when you click very quickly on the UI:

> I can also now reproduce on a new profile. My steps are:
> 1. `make run`;
> 2. Tools > Web Developer > Firefox OS Simulator;
> 3. press the Simulator button to start the Simulator;
> 4. press the Add Directory… button and select the manifest.webapp file for Gaia's Template app.

You especially need to do 4th step quickly, in order to try to send the validateManifest request, while 3rd step is still in process and isRunning=true whereas the remote debugger client isn't ready yet.

ochameau:
If that fixes all issues, I'd still like to do this modification before landing:

```
https://github.com/ochameau/r2d2b2g/blob/bbc22a5acd0f3d9b15f7840b99840ce51e1de177/addon/lib/remote-simulator-client.js#L101
this.connectDebuggerClient();
->
setTimeout(this.connectDebuggerClient.bind(this), 250);
```

I'd like to ensure it works even with the current behavior that can easily help us identify eventual races.

mykmelez:
With the latest changes, I see a different issue: when I Update an existing app after restarting Firefox, it gets stuck in the RUNNING state of validation.

Steps To Reproduce:
1. `mkdir ~/test-profile`
2. `make run PROFILE=~/test-profile`;
3. Tools > Web Developer > Firefox OS Simulator;
4. press the Add Directory… button and select the manifest.webapp file for Gaia's Template app;
5. quit Firefox;
6. `make run PROFILE=~/test-profile`;
7. Tools > Web Developer > Firefox OS Simulator;
8. press the Update button for the Template app.

Expected Results: Simulator starts, validation result becomes OK, and app is installed in Simulator.

Actual Results: Simulator starts, but validation result gets stuck in RUNNING, and app isn't installed in Simulator.

ochameau:
I tried hard but I wasn't able to reproduce it on windows, nor linux...
Is it an intermittent issue?

mykmelez:
> I tried hard but I wasn't able to reproduce it on windows, nor linux...
> Is it an intermittent issue?

No, I could reproduce it consistently. Hmm, but maybe it's Mac-specific. Or a timing issue that manifests on my computer because it's slower or faster than yours. I'll dig into it tomorrow to figure out what's going on.

mykmelez:
I dug into this yesterday, and I still haven't figured out what's going on, although I've made some progress. It doesn't seem to be a race in the Simulator itself, although this change exposes it. Rather, it looks to be somewhere between the client, the transport, and the server. Or perhaps even deeper down the stack, in the nsIServerSocket or nsISocketTransportService implementations (or, although I dearly hope this isn't the case, even lower, at the OS level).

Pure speculation: perhaps the client somehow manages to connect to the server after it initiates listening on the port but before it is prepared to send the hello packet, so it never sends that packet, and the client never receives it, and therefore the client never thinks the connection was successful (but also doesn't close the connection, because the underlying stream doesn't get closed).

Note that _DebuggerServer.openListener_ calls _nsIServerSocket.asyncListen_, which (presumably) asynchronously calls _DebuggerServer.onSocketAccepted_, which then synchronously calls _DebuggerServer._onConnection_ to send the hello packet. So maybe the asynchronicity of _nsIServerSocket.asyncListen_ is the locus of this race.

I wonder if @past would have any insight here!

mykmelez:
A bit more info, with the caveat that I'm a novice at this:

I used tcpdump to observe traffic, and this is what it looks like when the connection succeeds (with the server port hardcoded to 6000 to make it easier to observe, and success obtained by introducing a timeout between multiple connection attempts so the client doesn't try to reconnect too frequently):

```
05-15 10:34 > sudo tcpdump -i lo0 -v -A 'tcp port 6000'
tcpdump: listening on lo0, link-type NULL (BSD loopback), capture size 65535 bytes
…
    localhost.50290 > localhost.6000: Flags [S], cksum 0xfe34 (incorrect -> 0x54d7), seq 2994017685, win 65535, options [mss 16344,nop,wscale 4,nop,nop,TS val 317125420 ecr 0,sackOK,eol], length 0
E..@..@.@............r.p.u...........4....?........
...,........
10:34:36.082828 IP (tos 0x0, ttl 64, id 63671, offset 0, flags [DF], proto TCP (6), length 40, bad cksum 0 (->4416)!)
    localhost.6000 > localhost.50290: Flags [R.], cksum 0xfe1c (incorrect -> 0x0de0), seq 0, ack 2994017686, win 0, length 0
E..(..@.@............p.r.....u..P.......
10:34:36.585453 IP (tos 0x0, ttl 64, id 7134, offset 0, flags [DF], proto TCP (6), length 64, bad cksum 0 (->20d8)!)
    localhost.50291 > localhost.6000: Flags [S], cksum 0xfe34 (incorrect -> 0xf945), seq 2411106803, win 65535, options [mss 16344,nop,wscale 4,nop,nop,TS val 317125917 ecr 0,sackOK,eol], length 0
E..@..@.@............s.p.............4....?........
............
10:34:36.585509 IP (tos 0x0, ttl 64, id 60594, offset 0, flags [DF], proto TCP (6), length 64, bad cksum 0 (->5003)!)
    localhost.6000 > localhost.50291: Flags [S.], cksum 0xfe34 (incorrect -> 0x7667), seq 1452680243, ack 2411106804, win 65535, options [mss 16344,nop,wscale 4,nop,nop,TS val 317125917 ecr 317125917,sackOK,eol], length 0
E..@..@.@............p.sV.$3.........4....?........
............
10:34:36.585520 IP (tos 0x0, ttl 64, id 30686, offset 0, flags [DF], proto TCP (6), length 52, bad cksum 0 (->c4e3)!)
    localhost.50291 > localhost.6000: Flags [.], cksum 0xfe28 (incorrect -> 0xcc77), ack 1, win 9186, options [nop,nop,TS val 317125917 ecr 317125917], length 0
E..4w.@.@............s.p....V.$4..#..(.....
........
10:34:36.585529 IP (tos 0x0, ttl 64, id 40324, offset 0, flags [DF], proto TCP (6), length 52, bad cksum 0 (->9f3d)!)
    localhost.6000 > localhost.50291: Flags [.], cksum 0xfe28 (incorrect -> 0xcc77), ack 1, win 9186, options [nop,nop,TS val 317125917 ecr 317125917], length 0
E..4..@.@............p.sV.$4......#..(.....
........
10:34:36.586255 IP (tos 0x0, ttl 64, id 54258, offset 0, flags [DF], proto TCP (6), length 123, bad cksum 0 (->6888)!)
    localhost.6000 > localhost.50291: Flags [P.], cksum 0xfe6f (incorrect -> 0x1844), seq 1:72, ack 1, win 9186, options [nop,nop,TS val 317125917 ecr 317125917], length 71
E..{..@.@............p.sV.$4......#..o.....
........68:{
  "from": "root",
  "applicationType": "browser",
  "traits": {}
}
```

It shows port 50290 (the client) trying to connect to 6000 (the server) and failing. Then the next new client instance at 50291 tries to connect to the server and succeeds, with the server sending back something that I presume is the hello packet.

When the connection fails, on the other hand, the traffic looks similar except that no connection attempt ever succeeds, so the traffic consists of a sequence of request/response pairs like the one between 50290 and 6000, with the client port increasing incrementally each time the client process creates a new client instance to try again.

ochameau:
Are you able to reproduce it on Linux/Windows?
In case you are using slow hardware, wouldn't it be because setTimeout(0) stress the firefox process so much that b2g one isn't able to startup the debugger server in 15 seconds?

Otherwise, depending on the OS, zomby sockets may lay around and reach the maximum number of sockets and disallow creating any more new socket. So if you are running it a lot, it can eventually start failing.

Can you try enabling debugger logs by setting `devtools.debugger.log` to true on b2g side?

mykmelez:
> Are you able to reproduce it on Linux/Windows?

I haven't tried this yet.

> In case you are using slow hardware, wouldn't it be because setTimeout(0) stress the firefox process so much that b2g one isn't able to startup the debugger server in 15 seconds?

No, as I'm running on fast hardware, and the Simulator process starts up quickly.

> Otherwise, depending on the OS, zomby sockets may lay around and reach the maximum number of sockets and disallow creating any more new socket. So if you are running it a lot, it can eventually start failing.

I don't think this is the issue for several reasons:
1. I can reproduce after restarting the OS;
2. I have occasionally reproduced even when delaying reconnect attempts with a significant timeout like 250ms;
3. reconnect attempts consistently stall right before the console shows that the Simulator has started ("patch RemoteDebugger.start", "processing -dbgport command line option", etc.), no matter how long it takes to start (i.e. no matter how many times the client tries to reconnect), which suggests that Simulator startup is what stalls the last attempt.

> Can you try enabling debugger logs by setting devtools.debugger.log to true on b2g side?

I did that in my instrumented branch, and no DBG-SERVER messages appear in the log when the problem occurs (whereas they do appear when the connection succeeds, i.e. "info: r2d2b2g: DBG-SERVER: New debugging connection on 127.0.0.1:49519").
- My fork of your branch (w/master merge): https://github.com/mykmelez/r2d2b2g/tree/ochameau-pingback
- My instrumented fork of your branch: https://github.com/mykmelez/r2d2b2g/tree/ochameau-pingback-instrumented
- The difference: https://github.com/mykmelez/r2d2b2g/compare/ochameau-pingback...ochameau-pingback-instrumented

ochameau:
Just rebased against current master. (Sorry I realized too late that you forked this branch and I should have just done a merge...)

ochameau:
I was able to reproduce this issue on a Mac but wasn't able to figure out exactly what is going wrong without building custom builds.
I think that we are having a race when a socket start connecting to the nsIServerSocket component, in the following timeframe: between nsIServerSocket instanciation and the call to asyncListen:
http://mxr.mozilla.org/mozilla-central/source/toolkit/devtools/server/main.js#213

What actually happens is that the nsIStreamListener given to the nsIInputStreamPump on client side, seems to connect (inputstream.available() doesn't throw over time), but onStartRequest/onStopRequest/onDataAvailable never gets called:
http://mxr.mozilla.org/mozilla-release/source/toolkit/devtools/debugger/dbg-transport.js#111

Also, on server side onSocketAccepted passed to the nsIServerSocket doesn't get called:
http://mxr.mozilla.org/mozilla-release/source/toolkit/devtools/debugger/server/dbg-server.js#279

I tried to produce a minimal mochitest-like testcase without success. It looks like we really need two distinct processes with high level of concurrence to reproduce it.

Given that, I pushed a timeout workaround that allows to detect such scenario and force new connection that immediatly works.
Can you confirm that it also fixes the issues you were seeing?

Then, I still want to set a small timeout between each connection attempt that will avoid facing this race and soften CPU usage during connection:
https://github.com/ochameau/r2d2b2g/blob/bbc22a5acd0f3d9b15f7840b99840ce51e1de177/addon/lib/remote-simulator-client.js#L101

```
-  this.connectDebuggerClient();
+ setTimeout(this.connectDebuggerClient.bind(this), 250);
```

past:
I haven't actually found the time to look into this yet, but your description reminds me of bug 775924:

https://bugzilla.mozilla.org/show_bug.cgi?id=775924

That test occasionally seems to get stuck in a similar state that I haven't managed to fully decipher yet.

mykmelez:
> Given that, I pushed a timeout workaround that allows to detect such scenario and force new connection that immediatly works.
> Can you confirm that it also fixes the issues you were seeing?

Yes, and this is a reasonable workaround. But it never clears the timeouts, and it associates the value of _this._clientConnecting_ with the value of _client_ in the timeout callback, although _this._clientConnecting_ represents the status of the most recent _client_ instance, while _client_ is an instance specific to the callback (from the scope over which it closes).

So there's a race here: a call to _connectDebuggerClient_ creates a client and sets this._clientConnecting to `true`; then the client closes, setting this._clientConnecting to `false` and triggering another call to _connectDebuggerClient_, which sets it back to `true`; finally, the timeout for the first client occurs while the second client is still connecting, so the first timeout's callback tries to close its client, even though it's already closed.

Of course this will only happen if the server takes more than about three seconds to start, since otherwise a client will successfully connect before the first timeout occurs. But that's still one more race condition than I'd like to have!

Here's a modification that should eliminate the condition. It ignores the value of _this._clientConnecting_, unconditionally closes the client in the timeout callback, but cancels the timeout if the client connects or closes beforehand.

```
@@ -226,36 +225,36 @@ const RemoteSimulatorClient = Class({
     this._clientConnecting = true;

     let transport = debuggerSocketConnect("127.0.0.1", this.remoteDebuggerPort);
     // Unbrick some rare connection attempts that ends up facing a race.
     // In such scenario, the connection isn't refused but input-stream-pump
     // component used in DebuggerTransport never gets its onStartRequest/onStopRequest
     // methods called.
     // We cancel attempts taking more than 3seconds to achieve.
-    let timeout = setTimeout((function () {
-      if (this._clientConnecting) {
-        console.error("Killing stuck connection");
-        client.close();
-      }
-    }).bind(this), 3000);
+    let timeout = setTimeout(function onTimeout() {
+      console.log("Killing stuck connection");
+      client.close();
+    }, 3000);

     let client = new DebuggerClient(transport);

     client.addListener("closed", (function () {
+      clearTimeout(timeout);
       emit(this, "clientClosed", {client: client});
       this._stopGeolocation();
     }).bind(this));

     client.addListener("geolocationStart", this.onGeolocationStart.bind(this));
     client.addListener("geolocationStop", this.onGeolocationStop.bind(this));

     this._registerAppUpdateRequest(client);

     client.connect((function () {
+      clearTimeout(timeout);
       emit(this, "clientConnected", {client: client});
     }).bind(this));
   },

   _registerAppUpdateRequest: function(client) {
     client.addListener("appUpdateRequest", (function(type, pkt) {
       console.log("APP UPDATE REQUEST RECEIVED", JSON.stringify(arguments, null, 2));
       this._appUpdateHandler(pkt.appId);
```

A couple more notes:
- If the timeout callback fires, we should use _console.log_ instead of _console.error_ to log the problem. Otherwise regular users will see the message in their Error Console, and it will mislead them when they investigate and report other problems. (Alternately, we could use _console.warn_, which might make the messages more noticeable to Simulator developers while still flying under the radar of regular users.)
- I suspect we should drop the timeout to one second or even lower. In my tests, the client closes in 0-3ms most of the time when the server is unavailable and connects in 10-20ms after it becomes available. Sometimes it takes longer, and it'll take more time on a slower machine, so a lower timeout might sometimes close clients that aren't stuck; but that seems ok, since the addon will try reconnecting anyway. And both server and client are on the same machine, so the likelihood that a connection attempt will consistently take more than a second to complete seems low. Meanwhile, a three second timeout will make a noticeable difference to users who experience the race: it will add three seconds to the time it takes to start the Simulator! (If we wanted to get really fancy, we could set a short timeout that increases over time. But this probably isn't worth it right now.)
- _setTimeout_ and _clearTimeout_ are now being required in _remote-simulator-client_ on the master branch, so we need to remove the _require_ call in this branch, since both calls define them with _const_, which busts the module when you merge master into this branch.

addon/lib/remote-simulator-client.js
```diff
@@ -114,28 +99,45 @@ const RemoteSimulatorClient = Class({
...
-      this._clientConnected = false;
       this._clientConnecting = false;
-      this._remote = null;
-      emit(this, "disconnected", null);
+      if (!this._clientConnected && this.process) {
+        // If the connection was closed before connection succeed,
+        // and the process is still alive, the attempt was rejected or timed out.
+        // We should keep trying to connect until we reach our own timeout.
+        let timeout = this._timeout || 30000;
+        if (new Date().getTime() - this._startConnectingTime < timeout) {
```
mykmelez, line 108: (Going over this one more time just to make sure I didn't miss anything!)

Nit: `new Date().getTime()` -> `Date.now()`

addon/lib/remote-simulator-client.js
```diff
@@ -197,10 +194,14 @@ const RemoteSimulatorClient = Class({
...
         // NOTE: reset old allocated remoteDebuggerPort
         this.remoteDebuggerPort = null;
+        this._clientConnecting = false;
         this.shuttingDown = false;
         emit(this, "exit", result.exitCode);
       }).bind(this)
-    });    
+    });
+
+    this._startConnectingTime = new Date().getTime();
```
mykmelez, line 203: Nit: new Date().getTime() -> Date.now()

addon/lib/remote-simulator-client.js
```diff
@@ -225,10 +226,20 @@ const RemoteSimulatorClient = Class({
...
     let transport = debuggerSocketConnect("127.0.0.1", this.remoteDebuggerPort);
+    // Unbrick some rare connection attempts that ends up facing a race.
+    // In such scenario, the connection isn't refused but input-stream-pump
+    // component used in DebuggerTransport never gets its onStartRequest/onStopRequest
+    // methods called.
+    // We cancel attempts taking more than 3seconds to achieve.
+    let timeout = setTimeout((function () {
+      console.log("Killing stuck connection");
+      client.close();
+    }).bind(this), 1000);
```
mykmelez, line 237: Nit: there's no longer any need to `bind()` this function.

addon/lib/simulator.js
```diff
@@ -1011,7 +1011,10 @@ let simulator = module.exports = {
     // NOTE: if a b2g instance is already running send request
     //       or start a new instance and send request on ready
     if (this.isRunning) {
-      next();
+      if (this.remoteSimulator.isReady)
+        next();
+      else
+        this.remoteSimulator.once("ready", next);
```
mykmelez, line 1017: Nit: our convention is to put braces around all conditional blocks, even one-liners.

mykmelez:
Merged with review nits fixed and the reconnect timeout added!

PR #507, opened 2013-05-07
Ignore eventual previous b2g desktop zombie instances.

ochameau:
Fixes the issue described here: https://github.com/mozilla/r2d2b2g/pull/488#issuecomment-17507891

When something goes wrong on we leave a b2g instance alive, the next try to run b2g will fail. The b2g app window will stay black, or, when we ask for the console, only the console shows up.
That's because the new b2g instance will connect to the previous one and immediatly quit.

We should use `-no-remote` to prevent this.


mykmelez:
Nice fix! This makes B2G notify the user about why it won't start, which is better than simply opening a black window that doesn't work.

We should think about ways to make this experience even better, however. Perhaps we can detect that zombie process and force it to quit before we try to start a new one?

Issue #427, opened 2013-04-12
Allow for addition of default receipt for installed apps to test paid apps


nickdesaulniers:
If one wants to test receipt validation for a paid app, the appRecord returned by navigator.mozApps.getSelf must have a receipts property that is an array of receipt objects.  If there are no receipts, than it should be an empty array.  Currently, for apps installed through the simulator, the receipts property exists, but is null.  So if I want to test my paid packaged app, I must use the simulator to develop in (as most privileged APIs are b2g only, so far), but I can't actually test something like receipt verification for instance.

A receipt is bound to the appRecord during the call to navigator.mozApps.install as the second argument.

A receipt cannot be modified after installation but client side code can read and modify an in memory copy before sending it off to be validated, it just can't be written back to the appRecord from client side code.

I propose adding a button to the dashboard page for each installed application to add a default receipt.  The [validating a receipt MDN page](https://developer.mozilla.org/en-US/docs/Apps/Publishing/Validating_a_receipt) has more info on the fields within a receipt.  It may be nice to allow the developer to customize a few fields, especially the product  and verify fields.  But that can be a v2 improvement.  For now, maybe make all receipts have the product.storedata default to 1.

kumar303:
I don't really see how the simulator can be responsible for this. Creating a cryptographically valid receipt can only done by a signing authority such as the Marketplace (or any store).

It sounds like the Marketplace is in a better position to offer receipts for developer testing purposes. See these for progress on that:
https://bugzilla.mozilla.org/show_bug.cgi?id=855912
https://bugzilla.mozilla.org/show_bug.cgi?id=863015

wfwalker:
Can we wire up the simulator to somehow use https://marketplace.firefox.com/developers/test/receipts/ ?

wfwalker:
see https://bugzilla.mozilla.org/show_bug.cgi?id=865498

nickdesaulniers:
Note to self, don't forget to test the case where an app is:
- hosted
- packaged
- generated

nickdesaulniers:
https://bugzilla.mozilla.org/show_bug.cgi?id=867265

kumar303:
Currently you have to log in to the Marketplace Devhub to get a test receipt. However, we can maybe expose this with the [Marketplace API](http://firefox-marketplace-api.readthedocs.org/en/latest/). I'm half expecting @andymckay to tell me it's already there :)

andymckay:
[...]

Issue #497, opened 2013-05-04
Increase the duration of validity of the receipt.


nickdesaulniers:
Increase the [duration of validity](https://bugzilla.mozilla.org/show_bug.cgi?id=867662) of the receipt.

mykmelez:
@nickdesaulniers: do you plan to work on this in the near future? If so, let's assign this to you. Otherwise, let's move it out of the v4 milestone. Your pick!

mykmelez:
We need to keep valid receipts valid, so users don't have to manually deal with an expired receipt. Increasing their duration of validity might mitigate the problem, but it won't eliminate it (unless we can increase it to infinity!).

@nickdesaulniers Perhaps we can morph this issue into "keep valid receipts valid" and use it to track the work for that enhancement?

nickdesaulniers:
@andymckay Would it be possible to have test receipts have a far future expiration date, since developers should be screening for them anyways?  So maybe instead of adding the query string parameter to the request, and the additional UI to the simulator, just always fetch a long lived valid receipt?

andymckay:
You filed a bug about that I just made it very low priority because its pointless until we've got the rest of the infrastructure in place. We are blocked on the platform with https://bugzilla.mozilla.org/show_bug.cgi?id=757226, once thats cleared, we can start to worry about dealing with receipt expiry.

nickdesaulniers:
@andymckay , sure but now an app installed with a valid receipt in the simulator will be invalid after 24 hours.  Instead of having the dev have to reinstall the app every day, why can't the VALID test receipts have a far future expiration date?  Isn't the expiration time something that is changeable when it's generated?  Why is it currently set to 24 hours?  I'd be happy to close [bug 867662](https://bugzilla.mozilla.org/show_bug.cgi?id=867662) (having the API to choose the expiration) in exchange for just getting VALID receipts that default to something longer than 24 hours.  I agree having [bug 757226](https://bugzilla.mozilla.org/show_bug.cgi?id=757226) land will be helpful, but I don't think work on simulator receipt support should block on it.  Maybe there is another way we can solve the need-to-reinstall-every-24 hour problem?

andymckay:
The expiration time is changeable, however nothing checks it because you cannot replace a receipt. The simulator should not be checking the expiration date, just like apps should not be checking the expiration receipt.

nickdesaulniers:
[...]

Issue #496, opened 2013-05-04
Push receipts to device


nickdesaulniers:

mykmelez:
@nickdesaulniers: do you plan to work on this in the near future? If so, let's assign this to you. Otherwise, let's move it out of the v4 milestone. Your pick!

mykmelez:
@nickdesaulniers I'm going to move this back into the v4 milestone, as I think we need to do this in order to complete the feature!

nickdesaulniers:
The other half of this bug is [here](https://bugzilla.mozilla.org/show_bug.cgi?id=881499).

mykmelez:
Both halves of this have landed; fixed!

Issue #494, opened 2013-05-03
Add UI to dashboard to specify receipt type.


nickdesaulniers:
@potch is working on updating the dashboard UI.  With the addition of receipts, I'd like him to think about the UI for receipts.  @andymckay 's [tool](https://marketplace.firefox.com/developers/test/receipts/) features a dropdown with the receipt types of None, Valid, Expired, Invalid, Refunded, or None.  We may want to allow the dev to chose this on initial install _and_ subsequent updates.

mykmelez:
@nickdesaulniers: do you plan to work on this in the near future? If so, let's assign this to you. Otherwise, let's move it out of the v4 milestone. Your pick!

mykmelez:
@nickdesaulniers I'm going to move this back into the v4 milestone, as I think we need to do this in order to complete the feature!

@potch this is currently assigned to you, and I think @nickdesaulniers's thinking is that you would design the UX, then @nickdesaulniers would implement it.

@nickdesaulniers is that correct?

@potch are you available to do the design here?

nickdesaulniers:
I was thinking @potch would implement it along with the new UI, but if we're looking to implement something before the new UI lands, then I'd be happy to implement it with some UI decisions made for me.

potch:
Glad to work this into the new design, and happy to handle the UI stuff. Not sure what sauce is needed on the backend.

nickdesaulniers:
@potch , I'd like to implement the feature by milestone 4 which is due by 6/11/13.  I'm ready to go on this.  If you are too, I'd recommend if you created a separate branch that we both work out of, then could merge into master.  Thoughts?

nickdesaulniers:
The first step is to get listApps to include information on the installed receipt type, then have updateApp contain info on the newly selected receipt.  updateApp should probably be fired onchange of the receipt type UI element.

potch:
@nickdesaulniers, sounds easy enough :) is there a division of labor you want to do here? I have some UI stuff I can land in the coming week that could include this. Is there a particular format to the receipt info that the addon-land code would like? (integer, string, etc)

nickdesaulniers:
[...]

PR #509, opened 2013-05-07
Initial support for receipts Issue #427

nickdesaulniers:
review? @rpl @mykmelez

I created an [app](https://github.com/nickdesaulniers/simulatorTestReceipt) that uses our [receiptverifier](https://github.com/mozilla/receiptverifier) library (and found some bugs in it along the way).  Please see the [README](https://github.com/nickdesaulniers/simulatorTestReceipt/blob/master/README.md) for instructions on how to test.

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -181,10 +181,9 @@ var Simulator = {
     window.postMessage({ name: "openConnectDevtools" }, "*");
   },
 
-  // Once UI is added, this method should parse the receipt type
   // receipt type can be: ok, expired, invalid, refunded, none
   getReceiptType: function() {
-    return "ok";
+    return $('#receipt_type').val();
   },
 };
 
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -182,7 +182,7 @@ var Simulator = {
   },
 
   // Once UI is added, this method should parse the receipt type
-  // receipt type can be: ok, expired, invalid, or refunded
+  // receipt type can be: ok, expired, invalid, refunded, none
   getReceiptType: function() {
     return "ok";
   },
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -59,7 +59,7 @@ var Simulator = {
       window.postMessage({
         name: "addAppByTab",
         url: input.val().trim(),
-        receiptType: Simulator.getReceiptType()
+        receiptType: Simulator.getReceiptType(),
       }, "*");
       $("#form-add-app").get(0).reset();
     });
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -55,9 +55,11 @@ var Simulator = {
 
       var input = $('#add-app-url');
       var valid = input[0].checkValidity();
+      // hosted and generated apps
       window.postMessage({
         name: "addAppByTab",
-        url: input.val().trim()
+        url: input.val().trim(),
+        receiptType: Simulator.getReceiptType()
       }, "*");
       $("#form-add-app").get(0).reset();
     });
@@ -168,13 +170,22 @@ var Simulator = {
   },
 
   addAppByDirectory: function() {
-    window.postMessage({ name: "addAppByDirectory" }, "*");
+    // packaged apps
+    window.postMessage({
+      name: "addAppByDirectory",
+      receiptType: Simulator.getReceiptType()
+    }, "*");
   },
 
   openConnectDevtools: function() {
     window.postMessage({ name: "openConnectDevtools" }, "*");
-  }
+  },
 
+  // Once UI is added, this method should parse the receipt type
+  // receipt type can be: ok, expired, invalid, or refunded
+  getReceiptType: function() {
+    return "ok";
+  },
 };
 
 $(window).load(function() {
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -150,6 +150,7 @@ let simulator = module.exports = {
         type: "local",
         xkey: UUID.uuid().toString().slice(1, -1),
         receipt: null,
+        receiptType: receiptType,
       };
       console.log("Registered App " + JSON.stringify(apps[manifestFile]));
 
@@ -633,6 +634,7 @@ let simulator = module.exports = {
       installOrigin: origin,
       generated: true,
       receipt: null,
+      receiptType: receiptType,
     };
     if (receiptType && receiptType !== "none") {
       this.fetchReceipt(origin, receiptType, function (err, receipt) {
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -369,7 +369,9 @@ let simulator = module.exports = {
               };
               receiptType = null;
               simulator.fetchReceipt(receiptParams, function(err, receipt) {
-                if (err) return next(err, config);
+                if (err) {
+                  return next(err, config);
+                }
                 appInfo.appReceipt = receipt;
                 simulator.remoteSimulator.install(appInfo, onInstall);
               });
@@ -451,7 +453,9 @@ let simulator = module.exports = {
                   };
                   receiptType = null;
                   simulator.fetchReceipt(receiptParams, function(err, receipt) {
-                    if (err) return next(err, config);
+                    if (err) {
+                      return next(err, config);
+                    }
                     appInfo.appReceipt = receipt;
                     simulator.remoteSimulator.install(appInfo, onInstall);
                   });
@@ -477,7 +481,6 @@ let simulator = module.exports = {
       content: params,
       onComplete: function(response) {
         const INVALID_MESSAGE = "INVALID_RECEIPT";
-        console.log("request complete");
         if (response.status === 400 && "error_message" in response.json) {
           console.log("Bad request made to test receipt server: " +
             JSON.stringify(response.json.error_message));
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -163,9 +163,9 @@ let simulator = module.exports = {
         }
       };
       if (receiptType && receiptType !== "none") {
-        let manifest_url = "https://" + apps[manifestFile].xkey + ".simulator";
+        let manifestURL = "https://" + apps[manifestFile].xkey + ".simulator";
         let self = this;
-        this.fetchReceipt(manifest_url, receiptType, function(err, receipt) {
+        this.fetchReceipt(manifestURL, receiptType, function(err, receipt) {
           if (err || !receipt) {
             console.error(err || "No receipt");
           } else {
@@ -323,7 +323,7 @@ let simulator = module.exports = {
 
     let onInstall = function onInstall(res) {
       console.debug("webappsActor install app reply: ",
-      JSON.stringify(res));
+                    JSON.stringify(res));
       if (typeof next === "function") {
         // detect success/error and report to the "next" callback
         if (res.error) {
@@ -447,14 +447,14 @@ let simulator = module.exports = {
     }
   },
 
-  fetchReceipt: function fetchReceipt(manifest_url, receipt_type, cb) {
-    console.log("Fetching test receipt from marketplace",
-      JSON.stringify({ manifest_url: manifest_url, receipt_type: receipt_type }));
+  fetchReceipt: function fetchReceipt(manifestURL, receiptType, cb) {
+    console.log("Fetching " + receiptType + " test receipt for " + manifestURL);
     Request({
       url: TEST_RECEIPT_URL,
       content: {
-        manifest_url: manifest_url,
-        receipt_type: receipt_type,
+        // request params use undescore case
+        manifest_url: manifestURL,
+        receipt_type: receiptType,
       },
       onComplete: function(response) {
         const INVALID_MESSAGE = "INVALID_RECEIPT";
@@ -581,7 +581,6 @@ let simulator = module.exports = {
     let found = false;
     let tab = null;
     let title = null;
-    let self = this;
     for each (tab in Tabs) {
       if (tab.url == tabUrl) {
         found = true;
@@ -597,7 +596,7 @@ let simulator = module.exports = {
             simulator.addAppByTabUrl(tabUrl, true, receiptType);
           } else {
             if (receiptType && receiptType !== "none") {
-              self.fetchReceipt(origin, receiptType, function (err, receipt) {
+              simulator.fetchReceipt(origin, receiptType, function (err, receipt) {
                 if (err || !receipt) {
                   console.error(err || "No receipt");
                 } else {
@@ -641,7 +640,7 @@ let simulator = module.exports = {
           console.error(err || "No receipt");
         } else {
           addManifestArgs.receipt = receipt;
-          self.addManifest(addManifestArgs);
+          simulator.addManifest(addManifestArgs);
         }
       });
     } else {
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -149,7 +149,7 @@ let simulator = module.exports = {
       apps[manifestFile] = {
         type: "local",
         xkey: UUID.uuid().toString().slice(1, -1),
-        receipts: null,
+        receipt: null,
       };
       console.log("Registered App " + JSON.stringify(apps[manifestFile]));
 
@@ -169,7 +169,7 @@ let simulator = module.exports = {
           if (err || !receipt) {
             console.error(err || "No receipt");
           } else {
-            apps[manifestFile].receipts = receipt;
+            apps[manifestFile].receipt = receipt;
             self.updateApp(manifestFile, next);
           }
         });
@@ -337,7 +337,7 @@ let simulator = module.exports = {
     let appInfo = {
       appId: config.xkey,
       appType: null,
-      appReceipt: config.receipts,
+      appReceipt: config.receipt,
     };
 
     if (config.type == "local") {
@@ -633,14 +633,14 @@ let simulator = module.exports = {
       webapp: webapp,
       installOrigin: origin,
       generated: true,
-      receipts: null,
+      receipt: null,
     };
     if (receiptType && receiptType !== "none") {
       this.fetchReceipt(origin, receiptType, function (err, receipt) {
         if (err || !receipt) {
           console.error(err || "No receipt");
         } else {
-          addManifestArgs.receipts = receipt;
+          addManifestArgs.receipt = receipt;
           self.addManifest(addManifestArgs);
         }
       });
@@ -649,7 +649,7 @@ let simulator = module.exports = {
     }
   },
 
-  addManifestUrl: function(manifestUrl, receipts) {
+  addManifestUrl: function(manifestUrl, receipt) {
     console.log("Simulator.addManifestUrl " + manifestUrl);
 
     Request({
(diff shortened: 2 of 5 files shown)


nickdesaulniers:
Also, this patch introduces a bug where if you cmd+r an app, it is reinstalled with a receipt.  Should that be fixed before or after this can land?

addon/lib/simulator.js
```diff
@@ -1107,6 +1174,7 @@ let simulator = module.exports = {
   onMessage: function onMessage(message) {
     console.log("Simulator.onMessage " + message.name);
     let app = null;
+    let params = null;
```
nickdesaulniers: oh, this isn't used anywhere

addon/data/content/js/main.js
```diff
@@ -55,9 +55,11 @@ var Simulator = {
 
       var input = $('#add-app-url');
       var valid = input[0].checkValidity();
+      // hosted and generated apps
       window.postMessage({
         name: "addAppByTab",
-        url: input.val().trim()
+        url: input.val().trim(),
+        receiptType: Simulator.getReceiptType()
```
mykmelez: Nit: lately I've been putting a trailing comma at the end of the list of name-value pairs for an object literal. That way I don't have to make a spurious modification to the last pair the next time I add one to the list. We should do that here and elsewhere too.

addon/lib/simulator.js
```diff
@@ -434,6 +469,38 @@ let simulator = module.exports = {
...
 
+  fetchReceipt: function fetchReceipt(params, cb) {
+    console.log("Fetching test receipt from marketplace",
+      JSON.stringify(params));
+    Request({
+      url: TEST_RECEIPT_URL,
+      content: params,
+      onComplete: function(response) {
+        const INVALID_MESSAGE = "INVALID_RECEIPT";
+        console.log("request complete");
```
mykmelez: Nit: a message like this will be hard to connect to the request in question if other messages are getting logged to the console at the same time. Since the original message said "fetching test receipt from marketplace", this one should say something like "test receipt fetched".

addon/lib/simulator.js
```diff
@@ -52,6 +53,7 @@ const MANIFEST_CONTENT_TYPE = "application/x-web-app-manifest+json";
 
 let worker, remoteSimulator;
 let deviceConnected, adbReady, debuggerReady;
+let receiptType = null;
```
mykmelez: Storing this value in a global variable is brittle and error-prone, so we should avoid doing it. We previously discussed closing over a scope containing a value vs. propagating it through a long call chain, and we agreed that the former is better; but I didn't realize at the time that the scope you had in mind was the global scope, which all functions close over, and in which it is thus dangerous to store call-specific state.

So we should propagate the value through the call chain instead. But that doesn't mean we have to propagate it through the whole chain. We could instead retrieve receipts earlier, as soon as we identify the origin of an app. In the _addAppByTab_ case, that happens early, as soon as, addAppByTabUrl(), so we would only need to pass the value to that one function. It happens later in the _addAppByDirectory_ case, not until _updateApp(), which generates the xkey. But I don't see why we have to generate the xkey so late; we could move xkey generation up to addAppByDirectory(), in which case we would again only need to pass the value into one function.

mykmelez:
I'm confused about the way receiptType is set only when an app is added to the Dashboard. Do we expect the user to choose a receipt type only when adding the app? It seems to me that a user testing receipts will want to change the type of receipt associated with their app multiple times, and they won't want to remove and readd the app to do it. So the receipt type selector is something we would want to build into the Dashboard entry rather than something we expose to the user only when they initially add the app.

mykmelez:
> Also, this patch introduces a bug where if you cmd+r an app, it is reinstalled with a receipt. Should that be fixed before or after this can land?

We should fix this issue before the change lands.

I'm not sure if this is the problem, but it seems like the current implementation requires a receipt type when adding an app, and there's no option for "no receipt". I recommend we add "no receipt" as an option, make that the default, and don't retrieve a receipt when that type is selected. Then, provided my previous comment about the receipt type selector makes sense, we persist the receipt type for a given app in its app record.

nickdesaulniers:
`> It seems to me that a user testing receipts will want to change the type of receipt associated with their app multiple times, and they won't want to remove and readd the app to do it.`

Until [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=757226) lands, the only way to set the receipt associated with an app is at install time.

addon/lib/simulator.js
```diff
@@ -336,18 +357,26 @@ let simulator = module.exports = {
...
+            if (receiptType) {
+              let receiptParams = {
+                // The API rejects any URL that is not HTTP:// or HTTPS://
+                // The API rejects any URL without a suffix
+                manifest_url: 'http://' + config.xkey + '.simulator',
+                receipt_type: receiptType
+              };
+              receiptType = null;
+              simulator.fetchReceipt(receiptParams, function(err, receipt) {
+                if (err) return next(err, config);
```
mykmelez: Nit: enclose the conditional block in braces and put the statement on the next line per Mozilla code conventions:

```
if (err) {
  return next(err, config);
}
```

addon/lib/simulator.js
```diff
@@ -412,18 +441,24 @@ let simulator = module.exports = {
...
+                    manifest_url: config.origin,
+                    receipt_type: receiptType
+                  };
+                  receiptType = null;
+                  simulator.fetchReceipt(receiptParams, function(err, receipt) {
+                    if (err) return next(err, config);
+                    appInfo.appReceipt = receipt;
+                    simulator.remoteSimulator.install(appInfo, onInstall);
+                  });
+                } else {
```
mykmelez: Nit: this conditional block is nearly identical to the one for packaged apps; might we be able to factor out more of it?

mykmelez:
> > It seems to me that a user testing receipts will want to change the type of receipt associated with their app multiple times, and they won't want to remove and readd the app to do it.
> 
> Until this bug lands, the only way to set the receipt associated with an app is at install time.

As long as we can set the receipt on reinstall, which that bug suggests is the case, then this isn't a problem, since the Dashboard can (and does) reinstall the app as often as needed, f.e. every time the user presses the Update button, selects App > Refresh, or updates the addon.

So it isn't necessary to make the user remove and readd the app from/to the Dashboard to change the receipt type. We can simply reinstall the app (i.e. updateApp) whenever the user selects a different receipt type.

nickdesaulniers:
review? @mykmelez

addon/lib/simulator.js
```diff
@@ -147,11 +148,12 @@ let simulator = module.exports = {
       let apps = simulator.apps;
       apps[manifestFile] = {
         type: "local",
-        xkey: null,
-      }
+        xkey: UUID.uuid().toString().slice(1, -1),
+        receipts: null,
```
nickdesaulniers: I really should rename all references to "receipts", plural, in simulator.js to "receipt", singular.  This is because there will only ever be one receipt fetched from the simulator, though for an app record it must be an array of strings.

nickdesaulniers:
This patch works with command+r for hosted, generated, and packaged apps as well. :)  The only thing is that it still installs a valid receipt by default.  Thoughts?

mykmelez:
> This patch works with command+r for hosted, generated, and packaged apps as well. :) The only thing is that it still installs a valid receipt by default. Thoughts?

I think I addressed this earlier, when you said:

> Also, this patch introduces a bug where if you cmd+r an app, it is reinstalled with a receipt. Should that be fixed before or after this can land?

And I responded:

> We should fix this issue before the change lands.
> 
> I'm not sure if this is the problem, but it seems like the current implementation requires a receipt type when adding an app, and there's no option for "no receipt". I recommend we add "no receipt" as an option, make that the default, and don't retrieve a receipt when that type is selected. Then, provided my previous comment about the receipt type selector makes sense, we persist the receipt type for a given app in its app record.

I.e. we should not install a valid receipt by default. Instead, we should install no receipt by default. Then we'll introduce a UX for specifying a kind of receipt to install (and, if appropriate, introduce a way to specify the default kind; which is something to consider when we design the UX).

nickdesaulniers:
I will switch it to default to none in an incoming commit, but there won't be any way to test my patch works correctly or at all without the UX to specify the functionality.

wfwalker:
I see the wisdom of defaulting to installing No Receipt by default.

At the same time, I don't want the Best to the the enemy of the Good -- I really hope we can find some short-term way to enable installing of valid receipts rather than waiting for an all-singing all-dancing UX solution.

nickdesaulniers:
So this patch uses a simple select input to chose the type at install time.  Now you can use my test app and see the receipt get validated.  The app record will correctly report an empty array for the receipts by default (none), appear green for valid receipts, and red for invalid.  The expired receipt passes validation, but that's an issue on the validation service's end.

addon/lib/simulator.js
```diff
@@ -159,7 +161,21 @@ let simulator = module.exports = {
         if (!error) {
           simulator.runApp(app);
         }
-      });
+      };
+      if (receiptType && receiptType !== "none") {
+        let manifest_url = "https://" + apps[manifestFile].xkey + ".simulator";
```
mykmelez: Nit: use studlyCaps for local variable names, i.e. manifest_url -> manifestURL.

addon/lib/simulator.js
```diff
@@ -434,6 +447,40 @@ let simulator = module.exports = {
     }
   },
 
+  fetchReceipt: function fetchReceipt(manifest_url, receipt_type, cb) {
```
mykmelez: Nit: use studlyCaps for the parameter names, even though the request parameters use snake_case.

addon/lib/simulator.js
```diff
@@ -434,6 +447,40 @@ let simulator = module.exports = {
...
+    console.log("Fetching test receipt from marketplace",
+      JSON.stringify({ manifest_url: manifest_url, receipt_type: receipt_type }));
```
mykmelez: Nit: this would be simpler and avoid a _JSON.stringify_ call as:

```
console.log("Fetching " + receiptType + " test receipt for " + manifestURL);
```

addon/lib/simulator.js
```diff
@@ -434,6 +447,40 @@ let simulator = module.exports = {
...
+      url: TEST_RECEIPT_URL,
+      content: {
+        manifest_url: manifest_url,
+        receipt_type: receipt_type,
+      },
+      onComplete: function(response) {
+        const INVALID_MESSAGE = "INVALID_RECEIPT";
+        if (response.status === 400 && "error_message" in response.json) {
+          console.log("Bad request made to test receipt server: " +
+            JSON.stringify(response.json.error_message));
```
mykmelez, line 466: Does this value actually need _JSON.stringify_-ing? I would expect it to be a simple text string that autostringifies to a reasonable value.

addon/lib/simulator.js
```diff
@@ -521,18 +568,20 @@ let simulator = module.exports = {
...
   },
 
   /**
    * Installs the web page in the active tab as if it was an app.
    */
-  addAppByTabUrl: function(tabUrl, force) {
+  addAppByTabUrl: function(tabUrl, force, receiptType) {
     console.log("Simulator.addAppByTabUrl " + tabUrl);
     let url = URL.URL(tabUrl);
+    let origin = url.toString().substring(0, url.lastIndexOf(url.path));
```
mykmelez, line 583: Note to self: we do this twice and should factor it out and then upstream it to the _URL_ module implementation.

addon/lib/simulator.js
```diff
@@ -305,6 +321,25 @@ let simulator = module.exports = {
     File.mkpath(tempWebappDir);
     console.log("Created " + tempWebappDir);
 
+    let onInstall = function onInstall(res) {
+      console.debug("webappsActor install app reply: ",
+      JSON.stringify(res));
```
mykmelez: Nit: indent the second parameter so it lines up with the first, i.e.:

```
  console.debug("webappsActor install app reply: ",
                JSON.stringify(res));
```

addon/lib/simulator.js
```diff
@@ -545,9 +594,19 @@ let simulator = module.exports = {
...
         this.validateUrl(tabUrl, function(err) {
           if (err) {
-            simulator.addAppByTabUrl(tabUrl, true);
+            simulator.addAppByTabUrl(tabUrl, true, receiptType);
           } else {
-            simulator.addManifestUrl(tabUrl);
+            if (receiptType && receiptType !== "none") {
+              self.fetchReceipt(origin, receiptType, function (err, receipt) {
+                if (err || !receipt) {
+                  console.error(err || "No receipt");
```
mykmelez, line 605: This fails silently (modulo the well-concealed console error) and bails too early, before it finishes populating the packaged app record. So the app never appears in the Dashboard, and the user isn't told why. Also, the Dashboard stops displaying the full list of apps because of errors:

```
[12:04:54.915] Addon-message: listApps
[12:04:54.915] rendering /Users/myk/Mozilla/app-error/manifest.webapp
[12:04:54.931] rendering /Users/myk/Mozilla/app-warning/manifest.webapp
[12:04:54.932] rendering /Users/myk/Mozilla/r2d2b2g/gaia/test_apps/geoloc/manifest.webapp
[12:04:54.933] rendering /Users/myk/Mozilla/r2d2b2g/gaia/test_apps/image-uploader/manifest.webapp
[12:04:54.934] rendering /Users/myk/Mozilla/r2d2b2g/gaia/test_apps/membuster/manifest.webapp
[12:04:54.935] rendering /Users/myk/Mozilla/r2d2b2g/gaia/test_apps/template/manifest.webapp
[12:04:54.936] rendering /Users/myk/Mozilla/simulatorTestReceipt/manifest.webapp
[12:04:54.937] Addon-message: listTabs
[12:04:54.937] Addon-message: setPreference
[12:04:54.926] ReferenceError: reference to undefined property this.str[(this.index - 1)] @ resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/js/vendor/nunjucks-dev.js:1117
[12:04:54.929] TypeError: function root does not always return a value @ resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/js/vendor/nunjucks-dev.js:4008
--
[12:05:18.375] ReferenceError: reference to undefined property cur.disabled @ resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/js/vendor/jquery-1.9.1.js:3109
```

The "add app" code has accumulated some technical debt that makes fixing this complicated. Ideally we would still add the app, but we'd tell the user about the problem and provide UI for re-requesting the receipt. In the short term, however, it's ok to fail to add the app as long as we do so noticeably and don't break the dashboard in the process.

So let's do this if we fail to get the receipt: first, remove the app record from the _apps_ structure; second, use simulator.error() to report the problem and prompt the user to try again, something like "Error retrieving receipt. Please try adding the app again."

addon/lib/simulator.js
```diff
@@ -521,18 +568,20 @@ let simulator = module.exports = {
...
    */
-  addAppByTabUrl: function(tabUrl, force) {
+  addAppByTabUrl: function(tabUrl, force, receiptType) {
     console.log("Simulator.addAppByTabUrl " + tabUrl);
     let url = URL.URL(tabUrl);
+    let origin = url.toString().substring(0, url.lastIndexOf(url.path));
     let found = false;
     let tab = null;
     let title = null;
+    let self = this;
```
mykmelez: Nit: it isn't necessary to define and use _self_, because the _simulator_ singleton refers to the same object. So just use that instead. (If it wasn't available, the nit would instead be that _Function.bind_ is the new hotness.)

mykmelez:
> This fails silently (modulo the well-concealed console error) and bails too early, before it finishes populating the packaged app record. So the app never appears in the Dashboard, and the user isn't told why. Also, the Dashboard stops displaying the full list of apps because of errors:

Erm, sorry, that second issue (the Dashboard no longer displaying the full list of apps) is actually specific to packaged apps, and it doesn't happen for hosted apps. The first issue (silent failure), however, happens for both packaged and hosted apps.

addon/lib/simulator.js
```diff
@@ -147,11 +148,12 @@ let simulator = module.exports = {
       let apps = simulator.apps;
       apps[manifestFile] = {
         type: "local",
-        xkey: null,
-      }
+        xkey: UUID.uuid().toString().slice(1, -1),
+        receipt: null,
```
mykmelez, line 152: Also add a _receiptType_ property to the record, both here and for hosted apps, and set it to the provided type, so we know what type of receipt the app is supposed to have, so we can add receipt-re-retrieval (and verify that the receipt associated with the app is the kind the user requested).

mykmelez:
> At the same time, I don't want the Best to the the enemy of the Good -- I really hope we can find some short-term way to enable installing of valid receipts rather than waiting for an all-singing all-dancing UX solution.

Agreed, and I'm not vying for the perfect solution. But I also want to make sure that the minimal viable solution we implement initially will behave in a way that users understand and expect.

> So this patch uses a simple select input to chose the type at install time. Now you can use my test app and see the receipt get validated. The app record will correctly report an empty array for the receipts by default (none), appear green for valid receipts, and red for invalid. The expired receipt passes validation, but that's an issue on the validation service's end.

Ok, this seems reasonable. And it looks and works pretty well, modulo the few issues I recently noted.

addon/lib/simulator.js
nickdesaulniers, line 466: Yes, because `curl -v -d "eceipt_type=ok&manifest_url=http://6ea60555-e8c8-5648-b04a-0da73138abf9.simulator" https://marketplace.firefox.com/api/v1/receipts/test/` yields:

`{"error_message": {"receipt_type": ["This field is required."]}}`

and `curl -v -d "receipt_type=ok&manifest_url=httpk://6ea60555-e8c8-5648-b04a-0da73138abf9.simulator" https://marketplace.firefox.com/api/v1/receipts/test/` yields:

`{"error_message": {"manifest_url": ["Enter a valid URL."]}}`

and `curl -v -d '' https://marketplace.firefox.com/api/v1/receipts/test/` yields:

`{"error_message": {"manifest_url": ["This field is required."], "receipt_type": ["This field is required."]}}`

Thoughts? @mykmelez

mykmelez, line 466: If it's a JSON blob, then stringifying is fine!

addon/lib/simulator.js
nickdesaulniers, line 605: `> remove the app record from the apps structure`,
as a note, the app has only been registered in the apps structure by the time this error would occur only for packaged apps.

addon/lib/simulator.js
```diff
@@ -434,6 +450,40 @@ let simulator = module.exports = {
     }
   },
 
+  fetchReceipt: function fetchReceipt(manifestURL, receiptType, cb) {
+    console.log("Fetching " + receiptType + " test receipt for " + manifestURL);
+    Request({
+      url: TEST_RECEIPT_URL,
+      content: {
+        // request params use undescore case
```
mykmelez, line 458: Nit: undescore -> underscore!

mykmelez:
Ok, this looks good, and I'm happy to land it now in preparation for the v4 release. However, there are a few enhancements I think we'll need to make in order for this feature to be complete enough to ship in v4:
- add UX for users to select a receipt type after adding the app (#494);
- push receipts to device (#496);
- automatically update receipts when they expire.

That last one might be mitigated by #497, which will increase receipt duration; but it seems like we'll still need to auto-update receipts, even if they last longer, so they don't eventually stop working. (Or figure out some other way to keep valid receipts valid that doesn't require user intervention, like making them last forever.)

PR #512, opened 2013-05-10
clean up Makefile

mykmelez:
The change that prompted this branch was the merger of the profile and prosthesis targets, since they depended on each other, and thus it was misleading to have both (and users who did `make profile` would get a broken profile that couldn't be used with the addon because the prosthesis addon wouldn't be installed in B2G).

But along the way I identified a bunch of other things to improve and have implemented them. The commit log should explain each change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,9 @@ adb:
 	mkdir -p addon/data/$(B2G_PLATFORM)
 	cd addon/data/$(B2G_PLATFORM) && rm -rf adb $(ADB_BINARIES)
 	mkdir addon/data/$(B2G_PLATFORM)/adb
-	$(DOWNLOAD_CMD) $(ADB_URL)
+	if [ ! -f $(ADB_PACKAGE) ]; then \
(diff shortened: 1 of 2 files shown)


mykmelez:

ochameau:
+1 I got tricked multiple time when only doing make profile and getting broken result!

Also, I'm wondering if we want to call `really-clean` gaia rule instead of `clean`. 
It also removes xulrunner instead of just trashing the profile folder.

https://github.com/mozilla-b2g/gaia/blob/master/Makefile#L789

mykmelez:
> Also, I'm wondering if we want to call really-clean gaia rule instead of clean.
> It also removes xulrunner instead of just trashing the profile folder.

Hmm, perhaps! It's consistent with what we do in the r2d2b2g superdirectory on `make clean`, since we remove the ADB and B2G packages in that case; although XULRunner is also a lot larger than ADB and B2G, so redownloading it takes a lot more time.

ochameau:
It will soon be waaay smaller! https://bugzilla.mozilla.org/show_bug.cgi?id=871542

Issue #390, opened 2013-03-24
Put icon to the left of app details


davidgomes:
On the list of installed apps on the index.html, on the left of the app details (name, type of app, open location, etc) the icon of the app could be added (if it exists) and the default icon for no-icon if there's no icon.

mykmelez:
I like this idea. It would certainly help to distinguish between the various apps in the list.

mykmelez:
This was just fixed by @louisstow in #513, and the fix will make it into the next preview build!

PR #513, opened 2013-05-10
* Display an icon next to the app

louisstow:
Implement issue #390 (https://github.com/mozilla/r2d2b2g/issues/390)


louisstow:
Just noticed it needs to handle hosted app icons. One tic

digitarald:
Icons can also be data uris

mykmelez:
> Icons can also be data uris

Do you have an example app that uses a data: URL icon? That'd be very helpful for development and testing of this change!

mykmelez:
When I add my Mykzilla testing app to the Dashboard, the icon doesn't appear, and it looks like its URL has the wrong path. The manifest URL is http://mykzilla.org/app/manifest.webapp, and the icon path is /app/img/icon-128.png, but the URL of the icon in the Dashboard is http://mykzilla.org/app/app/img/icon-128.png (with an extra "app/" path part).

addon/data/content/js/applist.js
```diff
@@ -56,7 +56,26 @@ var AppList = (function() {
             app.prettyLastUpdate = timedelta(app.lastUpdate);
         }
         app.prettyType = Simulator.APP_TYPES[app.type];
+        
+        //use a default icon
```
mykmelez, line 60: Nit: Mozilla convention (from its [Coding Style Guide](https://developer.mozilla.org/En/Mozilla_Coding_Style_Guide) and usage in this script and elsewhere) is to put a space after comment slashes.

addon/data/content/js/applist.js
```diff
@@ -56,7 +56,26 @@ var AppList = (function() {
...
+        var iconPath = "default.png";
+
+        if (app.icon) {
+            if (app.icon.indexOf("data:image/") === 0) {
+                //set the src to the data uri
+                iconPath = app.icon;
+            }
+            else if (app.type === "hosted") {
```
mykmelez, line 68: Nit: Mozilla convention is to cuddle elses, i.e. put them on the same line as the closing brace from the previous _if_ block.

mykmelez:
This looks great, and I only see a couple of very minor nits. I'll merge this with those nits fixed shortly. @louisstow Thanks for the fix!

PR #516, opened 2013-05-11
Fix app installation on Linux systems

ghost:
Simulator uses the TmpD  loaction set by Linux systems rather than using a predetemined Value(/tmp in this case).

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -7,10 +7,12 @@
 
 const { Cc, Ci, Cu, ChromeWorker } = require("chrome");
 
+Cu.import("resource://gre/modules/Services.jsm");
+
 const { EventTarget } = require("sdk/event/target");
 const { emit, off } = require("sdk/event/core");
 const { Class } = require("sdk/core/heritage");
-
+const Environment = require("sdk/system/environment").env;
 const Runtime = require("runtime");
 const Self = require("self");
 const URL = require("url");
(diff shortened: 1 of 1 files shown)


addon/lib/remote-simulator-client.js
```diff
@@ -159,13 +161,23 @@ const RemoteSimulatorClient = Class({
           arguments: ["-e", 'tell application "' + path + '" to activate'],
         });
       }
-    });    
-
+    });  
+    
+    
+
```
mykmelez, line 167: Nit: one line of whitespace would be more conventional here.

addon/lib/remote-simulator-client.js
```diff
@@ -159,13 +161,23 @@ const RemoteSimulatorClient = Class({
...
-
+    });  
+    
+    
+    
+    let environment;
+    if(Runtime.OS=="Linux") {
+      environment=["TMPDIR="+Services.dirsvc.get("TmpD",Ci.nsIFile).path,];
+      if("DISPLAY" in Environment){
+	environment.push("DISPLAY="+Environment.DISPLAY);
```
mykmelez, line 172: Nit: this line should be indented to eight columns instead of two.

Also, the Mozilla convention (from its [Coding Style Guide](https://developer.mozilla.org/En/Mozilla_Coding_Style_Guide) and usage in this script and elsewhere) is to put a character of space around operators like the plus sign and the equivalence operator as well as before/after opening/closing  parentheses.

Finally, since the _environment_ array is defined on a single line, it shouldn't include a trailing comma, which is only helpful when an array definition stretches across multiple lines, with one item per line (since that makes it possible to add an item without modifying the line of the last item).

addon/lib/remote-simulator-client.js
```diff
@@ -159,13 +161,23 @@ const RemoteSimulatorClient = Class({
...
+      }
+    }
+    
     // spawn a b2g instance
     this.process = Subprocess.call({
       command: b2gExecutable,
       arguments: this.b2gArguments,
-
+      environment:environment,
```
mykmelez, line 180: Nit: Mozilla convention is to put a space after the colon in a property: value pair.

Also, I would leave a blank line after the list of non-method properties, as it makes it easier to distinguish them from the methods below.

mykmelez:
Just some minor nits, nothing significant. Thanks for the fix, @SoumenG! I'll merge this with fixes for the nits.

mykmelez:
> Nit: this line should be indented to eight columns instead of two.

Erm, I just noticed that the issue is actually that the indentation uses a tab character, so this nit should actually be to use space characters instead of tabs for all indentation.

Issue #514, opened 2013-05-10
app manifest validation fails to complain about bogus value for appcache_path


wfwalker:
It appears the otherwise excellent manifest validation in the simulator is not checking the validity of appcache_path.

Steps to Reproduce:

git clone git://github.com/wfwalker/opensun.git
add opensun/www/online.manifest as app to Simulator
should validate successfully
change appcache_path to something bogus
expected behavior: validation error
actual behavior: validation success

mykmelez:
Note related issue #306: when the appcache path is invalid, not only does the validator not warn you, but installation fails silently.

ochameau:
Note that appcache isn't supported for packaged app. In PR #523, I'm displaying a warning when we try to use appcache in packaged app.

Wouldn't it be interesting to introduce two modes for "Add directory" where we could select either hosted or packaged app type? The packaged app type would be the existing behavior. The hosted would be something similar to gaia DEBUG=1 mode, where we would automagically fire an httpd instance binded to the folder you gave. It sounds like some subtential amount of work, but would ease various painfull environnment step you have to jump before having a working hosted app. Like having correct content-types...

mykmelez:
> Wouldn't it be interesting to introduce two modes for "Add directory" where we could select either hosted or packaged app type? The packaged app type would be the existing behavior. The hosted would be something similar to gaia DEBUG=1 mode, where we would automagically fire an httpd instance binded to the folder you gave. It sounds like some subtential amount of work, but would ease various painfull environnment step you have to jump before having a working hosted app. Like having correct content-types...

I like that idea a lot! I would just be careful to distinguish between the "Gaia DEBUG=1" use case and the "third-party developer of hosted app" use case.

"Gaia DEBUG=1" is a mode in which a packaged app is served from a web server as if it was a hosted app, which seems pretty unique to Gaia development and isn't particularly useful in the Simulator, given that the Simulator knows how to package, install, and update a packaged app and can do so just as quickly as Gaia DEBUG=1 can reload a hosted app.

But it would be potentially very useful for the third-party developer of an app that is actually intended to be hosted to be able to point to a local directory containing its files and have the Simulator spin up a web server that hosts it (taking care of registering content types, rerouting domain names, etc. as necessary), so the developer doesn't have to go to the trouble of spinning up that server!

[...]

Issue #306, opened 2013-02-11
When appcache fails to resolve, the installation fails silently (simulator only)


mattbasta:
If the "appcache_path" of the manifest points to an invalid appcache or an appcache which does not resolve (i.e.: a non-200 response), Firefox OS would otherwise throw an error. In the simulator, this error does not appear and the app installation silently fails.

nickdesaulniers:
This may be because the simulator doesn't always have the latest builds of B2G and Gaia, so it wouldn't surprise me if we just happened to take a snapshot during a regression.  In the meanwhile, can you push a dummy branch of the app you were testing with the bad appcache and link it here, for testing?

mattbasta:
You can just use something like this:

http://badappcache.testmanifest.com/manifest.webapp

mykmelez:
Fixed by #523.

PR #523, opened 2013-05-14
Add limited set of sanity checks for usage of appcache in webapps.

ochameau:
I have the feeling that it would be better to improve installation and runtime code to expose error messages instead of doing nothing -or- dumping almost meaningless error code in b2g stdout.
Another middle step could be to start sharing sanity check methods between apps platform code and the simulator.
If we continue on that road one another farer step could be to start implementing checks in simulator in order to eventually upstream them in this apps platform code, that's what I'm trying to do in this pull request with a first module that focuses on checking appcache related failures.

If you like this approach, I'd like to continue and factor out of simulator.js all validation code, especially `_validateNameIcons`, `_validateWebAPIs`. And also try to move prosthesis validation code hosted in simulator actor (`onValidateManifest` and `_validateManifestPermissions`) to the firefox addon [if that's reasonnably possible].

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -23,6 +23,7 @@ const JsonLint = require("jsonlint/jsonlint");
 const ADB = require("adb");
 const Promise = require("sdk/core/promise");
 const Runtime = require("runtime");
+const Validator = require("./validator");
 
 // The b2gremote debugger module that installs apps to devices.
 const Debugger = require("debugger");
@@ -755,6 +756,15 @@ let simulator = module.exports = {
     // NOTE: add errors/warnings for WebAPIs not supported by the simulator
     simulator._validateWebAPIs(app.validation.errors, app.validation.warnings,
                                app.manifest);
+
+    // Appcache checks
+    if (["generated", "hosted"].indexOf(app.type) !== -1) {
+      // Only verify appcache for hosted apps
+      Validator.validateAppCache(app.validation.errors, app.validation.warnings,
+                                 app.manifest, app.origin);
+    } else if ("appcache_path" in app.manifest) {
+      app.validation.warnings.push("Packaged apps don't support appcache");
+    }
 
     if (["generated", "hosted"].indexOf(app.type) !== -1 &&
         ["certified", "privileged"].indexOf(app.manifest.type) !== -1) {
--- a/addon/lib/validator.js
+++ b/addon/lib/validator.js
@@ -1,0 +1,103 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const { Ci, Cu, Cr } = require("chrome");
+
+const AppsUtils = Cu.import("resource://gre/modules/AppsUtils.jsm");
+const { Services } = Cu.import("resource://gre/modules/Services.jsm");
+const { defer } = require('sdk/core/promise');
+
+exports.validateAppCache = function(errors, warnings, rawManifest, origin) {
+  let deferred = defer();
+
+  // Use ManifestHelper to normalize and retrieve the absolute appcache URI
+  let manifest = new AppsUtils.ManifestHelper(rawManifest, origin);
+  if (!manifest.appcache_path) {
+    deferred.resolve();
+    return deferred.promise;
+  }
+  let path = manifest.fullAppcachePath();
+  let uri = Services.io.newURI(path, null, null);
+
+  // Appcache only accepts http/https uris
+  // http://hg.mozilla.org/mozilla-central/file/c80dc6ffe865/uriloader/prefetch/nsOfflineCacheUpdate.cpp#l1249
+  if (uri.scheme != "http" && uri.scheme != "https") {
+    errors.push("Appcache only accepts http/https URIs");
+    deferred.resolve();
+  } else {
+    // Appcache reject any redirect
+    // http://hg.mozilla.org/mozilla-central/file/044d554846ff/uriloader/prefetch/nsOfflineCacheUpdate.cpp#l292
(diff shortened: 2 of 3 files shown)


ochameau:
This patch should fix issues #514 and #306.

rpl:
@ochameau Initially I tried to keep all the validation code on the firefox addon side, but I was afraid that differences in the "permissions table" and "manifest validation code" between Firefox and B2G builds would lead to misleading errors, then I preferred to move permissions checks on the "simulator actor"-side and be sure that our validation checks and mozApps API internal checks are directly comparable.

There are probably better solutions to achieve the same result, but I wanted to let you know the initial reason in case it may be helpful (because it may be not so obvious from the code).

ochameau:
Thanks for the update, I'll try to figure out if we can eventually shift this validation code from b2g to firefox.

Otherwise, you can test this patch with apps served over here with various broken/working cases: http://test.techno-barje.fr/appcache/.

addon/lib/simulator.js
```diff
@@ -23,6 +23,7 @@ const JsonLint = require("jsonlint/jsonlint");
 const ADB = require("adb");
 const Promise = require("sdk/core/promise");
 const Runtime = require("runtime");
+const validator = require("./validator");
```
mykmelez: Nit: I've been capitalizing these global singletons, whether or not they are constructors, to differentiate them from local variables.

addon/lib/validator.js
```diff
@@ -0,0 +1,96 @@
+const { Ci, Cu, Cr } = require("chrome");
```
mykmelez, line 6: Nit: this file needs a license header.

addon/lib/validator.js
```diff
@@ -0,0 +1,96 @@
...
+    deferred.resolve();
+    return deferred.promise;
+  }
+  let path = manifest.fullAppcachePath();
+  let uri = Services.io.newURI(path, null, null);
+
+  // Appcache only accept http/https uris
+  // http://hg.mozilla.org/mozilla-central/file/c80dc6ffe865/uriloader/prefetch/nsOfflineCacheUpdate.cpp#l1249
+  if (uri.scheme != "http" && uri.scheme != "https") {
+    errors.push("Appcache only accept http/https URIs");
```
mykmelez: Nit: accept -> accepts

addon/lib/validator.js
```diff
@@ -0,0 +1,96 @@
...
+    deferred.resolve();
+  } else {
+    // Appcache reject any redirect
+    // http://hg.mozilla.org/mozilla-central/file/044d554846ff/uriloader/prefetch/nsOfflineCacheUpdate.cpp#l292
+    let channel = Services.io.newChannel(uri.spec, null, null);
+    let listener = new ChannelListener(function (redirected, status) {
+      if (redirected)
+        errors.push("Redirected appcache manifests are ignored. " + uri.spec +
+                    " has been redirected to " + redirected);
```
mykmelez: Nit: we should observe the Mozilla convention of always bracing controlled statements, even one-liners, and especially when they span multiple lines, as described in the [Coding Style Guide](https://developer.mozilla.org/en-US/docs/Developer_Guide/Coding_Style), which [says about control structures](https://developer.mozilla.org/en-US/docs/Developer_Guide/Coding_Style#Control_Structures):

> Always brace controlled statements, even a single-line consequent of an if else else. This is redundant typically, but it avoids dangling else bugs, so it's safer at scale than fine-tuning.

addon/lib/validator.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+    channel.asyncOpen(listener, null);
+    channel.notificationCallbacks = listener;
+  }
+
+  return deferred.promise;
+}
+
+// nsIStreamListener+nsIChannelEventSink instance to check if a given nsIChannel
+// is being explicitely redirected. `onDone` function will be called
```
mykmelez: Nit: explicitely -> explicitly

addon/lib/validator.js
```diff
@@ -0,0 +1,96 @@
...
+// and also the http status code
+function ChannelListener(onDone) {
+  this.redirected = false;
+  this.onDone = onDone;
+}
+ChannelListener.prototype = {
+  onStartRequest: function(request, context) {},
+  onDataAvailable: function(request, context, stream, offset, count) {},
+  onStopRequest: function(request, context, status) {
+    console.log("stopRequest...");
```
mykmelez: Nit: this message will confuse folks reading the log who aren't already familiar with this module. It'd be better to provide some context, even if it's just a simple hierarchy that identifies the location of the statement, i.e.:

```
console.log("Validator: ChannelListener: stopRequest");
```

addon/lib/validator.js
```diff
@@ -0,0 +1,96 @@
...
+    channel.asyncOpen(listener, null);
+    channel.notificationCallbacks = listener;
+  }
+
+  return deferred.promise;
+}
+
+// nsIStreamListener+nsIChannelEventSink instance to check if a given nsIChannel
+// is being explicitely redirected. `onDone` function will be called
+// when the channel is loaded and receive a boolean to tell if it was redirected
```
mykmelez: Nit: this comment describes the first argument to the callback as a boolean, and it is a boolean if the request wasn't redirected; but otherwise it's a string, which confused me at first. It'd be clearer to describe the argument as a "string representing the URL to which the request was redirected, or null if the request wasn't redirected"; and then pass `null` instead of `false` in the latter case.

addon/test/test-validator.js
```diff
@@ -0,0 +1,105 @@
+const validator = require("validator");
```
mykmelez, line 6: Nit: this file could also use a license header.

mykmelez:
Looks good, just a few nits, fix at your discretion, r=myk!

ochameau:
All nits fixed. (The console.log(stopRequest) was just a debugging leftover, I removed it.)

PR #525, opened 2013-05-14
Allow to install apps without any content type for webapp's manifest.

ochameau:
Currently, when you try to add an hosted app whose manifest is served with no content type header, the installation will fail due to various exception in simulator.js.
We should allow installation and also warn the user about that.

You can test this issue with apps from:
  http://test.techno-barje.fr/manifests/


mykmelez:
Looks great!

PR #528, opened 2013-05-15
parseInt injected port to reassure AMO reviewers it isn't security hole

### LINKED ISSUE #524: reassure AMO reviewers that setting port value is safe
Code like the following made @kmaglione nervous in his AMO review of v3:

```
      contentScript: "window.addEventListener('load', function() { document.getElementById('port').value = '"+port+"'; }, true);"
```

@kmaglione suggests that we "at least escape the port with something like JSON.stringify(port)". There's no actual vulnerability, since we generate the port value and know that it's an integer. But we should do something to reassure future AMO reviewers of that fact.

mykmelez:
This fixes #524.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -910,9 +910,18 @@ let simulator = module.exports = {
     Tabs.open({
       url: "chrome://browser/content/devtools/connect.xhtml",
       onReady: function(tab) {
-        // NOTE: inject the allocated remoteDebuggerPort on the opened tab
+        // Inject the allocated remote debugger port into the opened tab.
+        // We know it's a Number, so we don't actually need to parseInt it,
+        // but doing so reassures AMO reviewers that we aren't exposing
+        // a security issue.
         tab.attach({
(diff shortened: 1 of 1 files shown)


mykmelez:
@nickdesaulniers How does this look to you?

Note that I reformatted the code block in the process to make it easier to read and so it doesn't extend past 80 columns. I didn't check the value with _isNaN_, though, as that would require us to either _parseInt_ in a separate statement (which is not [what @kmaglione wants](https://github.com/mozilla/r2d2b2g/issues/524#issuecomment-17912015)):

```
port = parseInt(port);
…
+ (isNaN(port) ? "something" : port) +
```

Or _parseInt_ twice:

```
+ (isNaN(parseInt(port)) ? "something" : parseInt(port)) +
```

It also seems unnecessary, since _remoteSimulator.remoteDebuggerPort_ should always return a valid integer Number; and when it doesn't, inserting the literal string `NaN` into the _Port_ field is a reasonable way to alert the user that something is wrong and they should report the issue.

nickdesaulniers:
Ship it!

PR #529, opened 2013-05-15
escape warning and error messages

### LINKED ISSUE #522: remote data unescaped when updating cached manifest
In his AMO review of v3, @kmaglione notes that the Simulator "generate(s) HTML fragments from remote data without escaping it in simulator.js:_updateCachedManifest(), which I suspect opens up a privilege escalation vulnerability when a manifest is served with a properly crafted content-type header or malformed JSON."

I'm not sure there's actually a vulnerability, but we should confirm that and fix any vulnerability that exists there.

mykmelez:
Fixes #522.

After looking into this, I think @kmaglione is right that there's a potential vector here, although I haven't identified a specific attack. But we want defense in depth, so we should mitigate against the potential risk, even without a known exploit.

Here's a fix that unconditionally escapes the warning and error messages in the Dashboard template. This has the side-effect of &lt;pre>formatting all messages, since the template can't distinguish between JSON linter messages (which are preformatted) and other messages (which aren't). I think that's ok, since the other messages are short one-liners that still look reasonable. But in the future we might consider annotating each message with a hint about how to format it.

@potch introduced this template post-3.0, so I'll need to craft a different fix for a 3.0 hotfix update if we decide to include this in one.

### DIFF
--- a/addon/data/content/index.html
+++ b/addon/data/content/index.html
@@ -114,7 +114,7 @@
                                 Errors:
                                 <ul class='app-validation-errors'>
                                     {% for msg in errors %}
-                                        <li>{{ msg }}</li>
+                                        <li><pre>{{ msg | escape }}</pre></li>
                                     {% endfor %}
                                 </ul>
                             {% endif %}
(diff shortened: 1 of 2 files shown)


mykmelez:
@ochameau How does this look to you?

mykmelez:
> @potch introduced this template post-3.0, so I'll need to craft a different fix for a 3.0 hotfix update if we decide to include this in one.

I created pull #531 for the 3.0 branch fix.

ochameau:
Looks good. We may want to also espace id/name: https://github.com/mykmelez/r2d2b2g/blob/38919575fe4a04472d16489f53e2698c6e01f977/addon/data/content/index.html#L84-L87

PR #527, opened 2013-05-15
update gcli submodule to maintain compatibility w/Firefox nightly builds

mykmelez:

PR #531, opened 2013-05-15
escape warning and error messages

mykmelez:
Here's the 3.0 branch equivalent of the fix for #522. If we decide to do a 3.0.1 hotfix update, which I'm leaning toward (primarily to pick up #527), we should take this too.

It is cherry picked from commit 38919575fe4a04472d16489f53e2698c6e01f977, which had conflicts in addon/data/content/index.html (namely: the master version modified the Nunjucks template in that HTML file, while the branch version has to modify the jQuery code in main.js).


mykmelez:
@ochameau Does this seem reasonable?

ochameau:
Looks good also.

PR #532, opened 2013-05-16
escape id, name, and iconPath values

mykmelez:


mykmelez:
@potch While reviewing #529, @ochameau identified the _id_, _name_, and _iconPath_ variables as potential additional vectors for privilege escalation attacks. I think he's right, and we should escape them, so this change does that. How does it look to you?

@ochameau also suggested escaping all values by default, which is a good idea. Unfortunately, Nunjucks doesn't support autoescaping according to [How Nunjucks is Different from Jinja2 ](http://nunjucks.jlongster.com/differences). But the [Nunjucks readme](https://github.com/jlongster/nunjucks#readme) says, "The next version is coming very soon (around 5/15/2013) and will include some major new features, such as autoescaping and custom block tags." Perhaps @jlongster knows the status of that!

Note that this doesn't appear to be an issue in 3.0, where the values are inserted into the DOM using jQuery's text() method.

mykmelez:
Thanks @potch! I'll go ahead and merge this then. And note that you can always merge a change you're reviewing if it looks good to you!

Issue #246, opened 2012-12-31
Menuitem doesn't appear in Nightly (2012-12-31)


leibovic:
I don't see "Firefox OS Simulator" in my "Web Developer" menu on the latest Nightly.

I suspect new menuitems might have landed and broken whatever logic is in place to add this menuitem.

mykmelez:
This is probably caused by [Add-on SDK bug 825471](https://bugzilla.mozilla.org/show_bug.cgi?id=825471).

jincreator:
I'm having same problem with Nightly.
Typing resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/index.html at address bar can make me access to Firefox OS Simulator page, but seems simulator is still not working.

rpl:
@mykmelez confirmed, this issue is related to private browsing api changes and prevents the r2d2b2g to be loaded on the last nightly builds.

@jincreator this is a temporary workaround (to be applied in the addon-sdk):

```
diff --git a/lib/sdk/private-browsing/utils.js b/lib/sdk/private-browsing/utils.js
index 1f57ca0..9996daa 100644
--- a/lib/sdk/private-browsing/utils.js
+++ b/lib/sdk/private-browsing/utils.js
@@ -21,14 +21,18 @@ let pbService;

 // Currently, only Firefox implements the private browsing service.
 if (require("../system/xul-app").is("Firefox")) {
-  pbService = Cc["@mozilla.org/privatebrowsing;1"].
-              getService(Ci.nsIPrivateBrowsingService);
+  try {
+    pbService = Cc["@mozilla.org/privatebrowsing;1"].
+                getService(Ci.nsIPrivateBrowsingService);

-  // set up an observer for private browsing switches.
-  observers.add('private-browsing-transition-complete', function onChange() {
-    // Emit event with in next turn of event loop.
-    deferredEmit(exports, pbService.privateBrowsingEnabled ? 'start' : 'stop');
-  });
+    // set up an observer for private browsing switches.
+    observers.add('private-browsing-transition-complete', function onChange() {
+      // Emit event with in next turn of event loop.
+      deferredEmit(exports, pbService.privateBrowsingEnabled ? 'start' : 'stop');
+    });
+  } catch(e) {
+    console.debug("EXCEPTION: ",e,e.fileName,e.lineNumber);
+  }
 }
```

juliangruber:
disabling and enabling the addon makes the menubar item show up again

[...]

PR #541, opened 2013-05-21
Set private browsing permission to enable FF os menuitem in private browsing only sessions fixes #431

### LINKED ISSUE #431: can't find Addon when 'History' is set to ' Never remember history' in preferences
seperate from #385 because it's absolutely different bug
### details

I have problem find Addon at Nightly 22.0a1(not nightly anymore) + FxOS Simulator 3.0pre4.
I'm now using 23.0a1 + FxOs Simulator 3.0pre6 and still confirm this issue, but seems only me.
I use Firefox with 'Never remember history' and when I change preference and restart, I can see Addon back.

It's hard to find people who set preference like me, so it makes hard to find what makes this bug.
Also I guess this bug will happen in Aurora 22, but didn't test it.

ochameau:


mykmelez:
@ochameau does this also fix #246? I.e. does it work around [SDK bug 855484](https://bugzilla.mozilla.org/show_bug.cgi?id=855484)?

ochameau:
Yes and no. It ends up doing something weird.
The menu now does appear, but in the last non-private window instead of the newly opened private window.

ochameau:
Let me rephrase it, the meny appears but when you click on it, the simulator is opened in a tab in the non-private window in the background.

PR #543, opened 2013-05-21
Use HEAD requests to determine if URL is a manifest

### LINKED ISSUE #179: changing the URL (to a large XPI?) freezes Firefox
Kris Maglione notes in his preliminary review of Simulator 1.0:

> It might be worth noting that I also managed to freeze Firefox pretty hard while changing the URL trying to get this to work. After a while changes to the URL field stopped producing error console output, a while later I got an unresponsive script dialog in request.js, and finally Firefox just froze and wouldn't come back. I'm not entirely sure how to reproduce this, but it probably has something to do with me accidentally pasting the URL of a a 68MB r2d2b2g XPI in the field at one point... In any case, I thought it worth mentioning.

bkase:
Fixes #179
URLs are no longer checked with GET requests in the simulator-- HEAD requests instead.

Included is a copy of the patched version of the Request module for the Add-on SDK that supports HEAD requests (awaiting review here: https://github.com/mozilla/addon-sdk/pull/1004 )

### DIFF
--- a/addon/lib/request.js
+++ b/addon/lib/request.js
@@ -56,7 +56,7 @@ function runRequest(mode, target) {
   let source = request(target)
   let { xhr, url, content, contentType, headers, overrideMimeType } = source;
 
-  let isGetOrHead = function(mode) mode == "GET" || mode == "HEAD";
+  let isGetOrHead = (mode == "GET" || mode == "HEAD");
 
   // If this request has already been used, then we can't reuse it.
   // Throw an error.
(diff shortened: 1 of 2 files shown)


mykmelez:
This looks great! And it's ready to merge, modulo any changes you also make to the version we land upstream (i.e. the nit that @ochameau identified). We should keep the two implementations in sync, to ease the AMO review process, and so there's no chance of their behaviors diverging.

Issue #519, opened 2013-05-13
reported geolocation doesn't change immediately after being changed by user


mykmelez:
The geolocation reported to apps by the geolocation simulation doesn't always change immediately after being changed by the user. Sometimes apps that check the location right after the user changes it still see the old location for some period of time.

I've seen this with my [Mykzilla test app](http://www.mykzilla.org/app/), and I've heard of it happening to other users. In fact it seems to be quite common. Which makes it hard to debug geolocation use in apps, since developers can't reliably test the effect of a location change on their apps.

@nickdesaulniers Do you have any ideas about what might be going on here?

nickdesaulniers:
Because this isn't hooked up to watch position, there is a [timer that takes a second to update](https://github.com/mozilla/r2d2b2g/blob/96bbe42e665f91401784e1d484c6e6139db4c473/prosthesis/components/FakeGeolocationProvider.js#L85) the coordinates.  I think this is a duplicate of #470 .

nickdesaulniers:
fixed in #544

PR #544, opened 2013-05-22
watch position for changes when user says to use 'current' position (fixes #470)

### LINKED ISSUE #470: Use watch position for geolocation
The user/developer's current coordinates may change as they develop.  Use watchPosition instead of [getCurrentPoition](https://github.com/mozilla/r2d2b2g/blob/6d2b02252b55acd17485457e45cdfbd2438ba1ff/addon/lib/remote-simulator-client.js#L253).  You would then need to add an r2d2b2g:geolocation-stop request from shell on custom selection to the dbg-simulator-actors, then to remote-simulator-client.

The issues that arise involves the [timer in the FakeGeolocationProvider](https://github.com/mozilla/r2d2b2g/blob/6d2b02252b55acd17485457e45cdfbd2438ba1ff/prosthesis/components/FakeGeolocationProvider.js#L85).

When using custom coordinates, how often should the timer fire?  Currently, it's every second.  When we use current coordinates though, we don't want to use a timer, and instead want to update whenever the Firefox process (remote-simulator-client) receives a watch update.  That way, the app behaves as if it's using a real life watchPosition call (because it is).  Maybe a timer could be defined during start up, but reinitialized during each update from the shell (meaning only for custom coordinates).

Also, is it important that the developer be able to specify the period of the timer?

mykmelez:


mykmelez:
@nickdesaulniers here's an initial implementation of the fix for #470 (via #519, which is what I originally set out to fix). Before you review this, though, I want to pay down some technical debt that has accumulated in the code I'm touching. So hold off a bit; I'll push some more changes tomorrow and then let you know when this is ready to review!

mykmelez:
@nickdesaulniers Ok, this is ready for review!

Note that there appears to be a bug in the implementation of _clearWatch_ that causes it to fail to clear the first position watcher registered via _watchPosition_. So if you set the simulator to current, set it to custom, and then set it back to current again, there will be two watchers watching for geolocation changes, and the simulator will get duplicate notifications.

Clearing subsequent watchers works fine, so if you toggle the setting again, you still just have two watchers. I'll file a platform bug on this behavior once I can reduce a testcase.

Note also that the second and subsequent watchers may get called twice when they are first registered. That's because the implementation caches the position obtained when a previous watcher was registered, and it tells the new watcher about that position immediately; but it also checks for an update and provides it to the new watcher. That isn't a bug, it's intended behavior! To give new watchers a position as quickly as possible while also giving them an accurate position as soon as possible thereafter.

nickdesaulniers:
Code looks clean, just had some discrepancies when testing with [GeoLocTest](https://github.com/nickdesaulniers/GeoLocTest) and not opening the geolocation dialog.

mykmelez:
@nickdesaulniers noticed an issue with this change: his GeoLocTest app would time out waiting for a position.

It turns out that nsGeolocationService assumes the watcher won't be updated until after _watch_ returns:

https://github.com/mozilla/mozilla-central/blob/02b984bc301ecddc3154938239041918e4508ab3/dom/src/geolocation/nsGeolocation.cpp#L853-L857

So we have to wait until a turn of the event loop to call the watcher's _update_ method for the first time.

nickdesaulniers:
Tested with [mykzilla](http://www.mykzilla.org/app/manifest.webapp) and [GeoLocTest](https://github.com/nickdesaulniers/GeoLocTest).

PR #549, opened 2013-05-22
Move all manifest validation logic to /addon/lib/validator.js and unit test it.

ochameau:
So I looked at differences between gecko18 and m-c on PermissionTables and AppsUtils and the result is that:
- there is no differences for AppsUtils
- only one commit, that should ideally be uplifted on PermissionTables - http://hg.mozilla.org/mozilla-central/rev/01f0d7b85857

Given that, it would be really easier and less error prone to move all validation logic to the addon. It may save us from races involved in this complex asynchronous code.

Also, it highlights that we are doing the same check twice.
The main fact that prevent sharing validation code is that the platform code is extremelly bad at reporting what is going wrong. It just returns false and sometimes tries to print something to the console.
Speaking about developer tools, we should aim to fix that at the source and improve error reporting when installing an App. By doing so, I think we should be able to reuse the validation logic being involved in the platform in order to execute some check before installing the app.

### DIFF
--- a/addon/lib/validator.js
+++ b/addon/lib/validator.js
@@ -5,15 +5,16 @@
 
 const { Ci, Cu, Cr } = require("chrome");
 
-const AppsUtils = Cu.import("resource://gre/modules/AppsUtils.jsm");
+const { AppsUtils, ManifestHelper } = Cu.import("resource://gre/modules/AppsUtils.jsm");
 const { Services } = Cu.import("resource://gre/modules/Services.jsm");
+const { PermissionsTable, expandPermissions } = Cu.import("resource://gre/modules/PermissionsTable.jsm");
 const { defer } = require('sdk/core/promise');
 
 exports.validateAppCache = function(errors, warnings, rawManifest, origin) {
   let deferred = defer();
 
   // Use ManifestHelper to normalize and retrieve the absolute appcache URI
-  let manifest = new AppsUtils.ManifestHelper(rawManifest, origin);
+  let manifest = new ManifestHelper(rawManifest, origin);
   if (!manifest.appcache_path) {
     deferred.resolve();
     return deferred.promise;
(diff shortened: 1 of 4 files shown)


ochameau:
In addition to unit tests, you can manually test this patch by installing this webapps manifest locally:
http://test.techno-barje.fr/apps/manifests.zip

addon/lib/validator.js
```diff
@@ -5,15 +5,16 @@
 
 const { Ci, Cu, Cr } = require("chrome");
 
-const AppsUtils = Cu.import("resource://gre/modules/AppsUtils.jsm");
+const AppsUtilsGlobal = Cu.import("resource://gre/modules/AppsUtils.jsm");
```
mykmelez: Nit: we might as well just import AppsUtils and ManifestHelper directly:

const { AppsUtils, ManifestHelper } = Cu.import("resource://gre/modules/AppsUtils.jsm");

addon/lib/validator.js
```diff
@@ -101,3 +102,108 @@ ChannelListener.prototype = {
...
+  if (!manifest.icons || Object.keys(manifest.icons).length == 0) {
+    warnings.push("Missing 'icons' in Manifest.");
+  } else {
+    // update registered app icon
+    let size = Object.keys(manifest.icons).sort(function(a, b) b - a)[0] || null;
+    if (size) {
+      app.icon = manifest.icons[size];
+    }
+
+    // NOTE: add warnins if 128x128 icon is missing
```
mykmelez: Nit: warnins -> warnings

addon/lib/validator.js
```diff
@@ -101,3 +102,108 @@ ChannelListener.prototype = {
...
+    if (! manifest.icons["128"]) {
```
mykmelez: Nit: there's an unusual space between the ! and manifest.icons here.

addon/lib/validator.js
```diff
@@ -101,3 +102,108 @@ ChannelListener.prototype = {
...
+      } else if (permissionAction === Ci.nsIPermissionManager.DENY_ACTION) {
+        errors.push("Denied permission '" + name + "' for app type '" + appType + "'.");
+      } else {
+        let access = manifest.permissions[name].access;
+        try {
+          if (access && expandPermissions(name, access).length === 0) {
+            errors.push("Invalid access '" + access + "' in permission '" + name + "'.");
+          }
+        } catch(e) {
+          console.exception(e);
```
mykmelez: This writes some pretty gnarly messages into the console:

```
info: r2d2b2g: executing 'r2d2b2g/tests/test-validator.test permission access on permission without access'
Error: PermissionsTable.jsm: expandPermissions: Invalid Manifest : alarms readonly

resource://gre/modules/PermissionsTable.jsm 323
    callback.apply(null, args);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 288, in 
    timer.setTimeout(function() { onDone(self); }, 0);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 458, in 
    options.onDone(self);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 440, in runNextTest
    self.start({test: test, onDone: runNextTest});
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 471, in start
    this.test.testFunction(this);
  File "resource://gre/modules/commonjs/sdk/test.js", line 72, in 
    test(assert);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/tests/test-validator.js", line 244, in exports["test permission access on permission without access"]
    validator.validatePermissions(errors, warnings, manifest, app);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/validator.js", line 185, in 
    permissionsNames.forEach(function(name) {
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/validator.js", line 197, in 
    if (access && expandPermissions(name, access).length === 0) {
  File "resource://gre/modules/PermissionsTable.jsm", line 323, in expandPermissions
    throw new Error(errorMsg);
```

I think we'd be better off calling _console.error_ or even just _console.warn_ to write a simple one-line message like `Error: PermissionsTable.jsm: expandPermissions: Invalid Manifest : alarms readonly`, since the stack isn't significant to users, nor do we particularly benefit from it.

mykmelez: Alternately, we could skip logging to the console entirely, since we add this error to the list of validation errors, and we display it to the user in a more visible location.

addon/test/test-validator.js
```diff
@@ -102,6 +102,176 @@ exports["test nonexistent"] = function(assert, done) {
...
+  let errors = [], warnings = [];
+  let manifest = {}, app = {};
+  validator.validateNameIcons(errors, warnings, manifest, app);
+  assert.equal(errors.length, 1, "missing name in manifest is an error");
+  assert.equal(errors[0], "Missing mandatory 'name' in Manifest.");
+  assert.equal(warnings.length, 1, "missing icons in manifest causes a warning");
+  assert.equal(warnings[0], "Missing 'icons' in Manifest.");
+}
+
+exports["test bigest icon selection"] = function(assert) {
```
mykmelez: Nit: bigest -> biggest

addon/test/test-validator.js
```diff
@@ -102,6 +102,176 @@ exports["test nonexistent"] = function(assert, done) {
...
+  let manifest = {
+    name: "foo",
+    icons: {"16": "16.png", "128": "128.png", "32": "32.png"}
+  };
+  let app = {};
+  validator.validateNameIcons(errors, warnings, manifest, app);
+  assert.equal(errors.length, 0);
+  assert.equal(warnings.length, 0);
+  assert.equal(app.name, "foo", "app.name has been set");
+  assert.equal(app.icon, "128.png", "selected the bigest icon");
```
mykmelez: Nit: bigest -> biggest

mykmelez:
Note that I get a test timeout when running tests on Nightly/Aurora, although they work on Beta/Release, and they behave the same on the master branch (so it isn't an issue specific to this branch):

```
  File "/Users/myk/Mozilla/r2d2b2g/addon-sdk/bin/cfx", line 33, in <module>
    cuddlefish.run()
  File "/Users/myk/Mozilla/r2d2b2g/addon-sdk/python-lib/cuddlefish/__init__.py", line 928, in run
    bundle_sdk=options.bundle_sdk)
  File "/Users/myk/Mozilla/r2d2b2g/addon-sdk/python-lib/cuddlefish/runner.py", line 743, in run_app
    OUTPUT_TIMEOUT, test_name, parseable)
Exception: Test output exceeded timeout (60s).
make: *** [test] Error 1
```

addon/lib/validator.js
ochameau: I just removed the console.exception.

addon/lib/validator.js
ochameau: Ahah, I'm already loosing all my smart jetpack code practices :)

ochameau:
This time I was able to reproduce this timeout issue, before only before rebasing this branch to last master.
The failing test was the remote-simulator-client. I haven't try to look more deeply why it was failing as I'm not sure it was related to this patch.

mykmelez:
>  The failing test was the remote-simulator-client. I haven't try to look more deeply why it was failing as I'm not sure it was related to this patch.

I see it both with and without your patch, so I don't think it's related to your patch.

PR #550, opened 2013-05-22
Stop injecting custom b2g content stylesheet thanks to bug 874984. Fixes #495

### LINKED ISSUE #495: Scrollbars repaint incorrectly on Firefox OS Simulator, on Ubuntu 13.04
STR:
1. Start up Firefox OS Simulator. (I'm using 3.0 on Ubuntu 13.04 64-bit)
2. Launch settings app.
3. Scroll. (either with mousewheel, or click-and-fling)

ACTUAL RESULTS: The scrollbar repaints incorrectly, as you scroll. (until it starts to fade away, after about a second, at which point it's fully repainted correctly)

Basically, it looks like we're not invalidating enough of the scrollbar. We _are_ invalidating the newly-covered-by-scrollbar and the no-longer-covered-by-scrollbar regions -- but we're apparently not invalidating the still-covered-by-scollbar-but-now-covered-by-a-different-looking-part-of-the-scrollbar. 

This might be an underlying gecko bug, but I've only seen it in Firefox OS simulator so far, so I'm filing it here to start.
Screenshot:
broken-scrollbar-screenshot

ochameau:
Following platform b2g desktop bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=874984
is going to fix multiple things for us:
- it will no longer apply to xul content, so that we no longer need to avoid applying the existing content.css file shipped in b2g desktop,
- it fixes the scrollbar look'n feel in order to have exactly the same appearance that we have on device,
- it also fixes scrollbar repaint issues reported in #495.

We should wait before landing this patch to either uplift this bug or cherry pick it on our build.


ochameau:
As bug 874984 is controversial and I would really like to see it fixed for v4, I've updated the patch to no longer completely remove b2g.css and use m-c one. But instead still ship a custom one that won't apply to xul documents.
It allows to drop a bunch of code that was trying to apply this CSS manually. For some reason, it looks like we were trying to inject it twice?

I've also updated b2g.css to match very last m-c version that fixes scrollbars (we now have mobile look'n feel and no longer suffer from the repaint bug #495) and also fixes <input type=range> issues (https://bugzilla.mozilla.org/show_bug.cgi?id=877128).

And when the platform issue will be finally fixed, we will be able to drop the chrome.manifest line and b2g.css.

mykmelez:
This seems ok, but wouldn't it be more performant to add an explicit "html" namespace (in addition to the default namespace) and use it wherever possible?

```
@namespace html url("http://www.w3.org/1999/xhtml");
```

Then we could do:

```
html|select,
html|textarea,
html|button,
html xul|button,
html|* > input:not([type="image"]):not([type="range"]) {
  -moz-appearance: none !important; /* See bug 598421 for fixing the platform */
  border-radius: 3px;
}
```

Where `html|select`, `html|textarea`, `html|button`, and `html|*` now match specific HTML elements instead of elements that are descendants of an `html` tag.

`html xul|button` still has to be an expensive descendant selector, but it's one of only three such selectors, which is still a lot better.

Also, the other two are `xul|window xul|scrollbar` and `html xul|scrollbar`, but it seems like `html xul|scrollbar` could be simply `xul|scrollbar`, since `xul|window xul|scrollbar` sets `display: none` to hide the scrollbar in XUL windows, so it shouldn't matter if `xul|scrollbar` rules apply to them.

ochameau:
I think you are right and we should be able to simplify this css.
We prefixed all rules with `html` to apply only on html document in order to avoid modifying xul documents.
So `html select` isn't a pure equivalent of `html|select`. Actually, `html|select` will be equivalent to `select`.
But at the end, I think we can simplify to what you suggested and only address `xul|...` selectors like `xul|button`.
I'll try to confirm this with Vivien who was involved with these CSS.

ochameau:
Ok, so the story of this content.css is that it has been copy pasted from Fennec,
and some rules aren't really usefull for b2g. So far the `xul|window xul|scrollbar` scenario doesn't really exists on b2g. Then the most problematic rule for us is this `xul|button` one which seems to be related to some xbl component.

I've updated the patch to only take fixes from bug 877128 and avoid breaking xul button style in geolocation popup, by doing what you suggested in previous comment. I removed the `xul|window xul|scrollbar` rule to prevent hiding the scrollbar in Error console.
The is only one difference left between master and this branch now: the scrollbar in Error console isn't the native one but the mobile one. If that's an issue, we will have to readd `html` prefix in front of all `xul|scrollbar*` rules.

mykmelez:
>  The is only one difference left between master and this branch now: the scrollbar in Error console isn't the native one but the mobile one. If that's an issue, we will have to readd html prefix in front of all xul|scrollbar\* rules.

Hmm, I think we do want that, since otherwise the scrollbar in the Error Console will look quite strange and be difficult to use.

This means the Simulator stylesheet will be different from the default one for B2G, which is suboptimal. But the Simulator has different requirements, since it displays XUL windows that are intended to look like native desktop OS windows; and it targets desktop OSes where the performance of these inefficient rules will be less of a factor.

Thus it is something I can live with until the underlying platform limitation is removed.

ochameau:
I've updated the patch to prevent modifying xul scrollbars. 
Note that it won't fix the repaint bug for xul scrollbars in the error console.

mykmelez:
>  Note that it won't fix the repaint bug for xul scrollbars in the error console.

Yup, that's fine, as it's a separate issue.

PR #510, opened 2013-05-07
Start using gaia's desktop helper to fix touch events and gain API mocks

ochameau:
By using desktop helper, we will be able to fix touch events and also share API mocks available here:
https://github.com/mozilla-b2g/gaia/tree/master/tools/extensions/desktop-helper/content/data/lib

This patch depends on gaia modification from https://bugzilla.mozilla.org/show_bug.cgi?id=869540.

PR #555, opened 2013-05-23
FIX: workaround lock rotation regression on "b2g-18.0.2013-05-22"

rpl:
gaia system app nodePrincipal.appStatus reported as "app not installed"
(nodePrincipal.appStatus == APP_STATUS_NOT_INSTALLED)

if origin is "app://system.gaiamobile.org" in mozLockOrientation method, it doesn't deny rotation lock even if its appStatus is reported as not installed.

Console Messages:

```
Error: deny mozLockOrientation:
Source File: jar:file:///.../b2g-prosthesis@mozilla.org.xpi!/components/SimulatorScreen.js
Line: 210
```

### DIFF
--- a/prosthesis/components/SimulatorScreen.js
+++ b/prosthesis/components/SimulatorScreen.js
@@ -130,7 +130,6 @@ SimulatorScreen.prototype = {
     let globalScreen = GlobalSimulatorScreen;
 
     let nodePrincipal = this.nodePrincipal = aWindow.document.nodePrincipal;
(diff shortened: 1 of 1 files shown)


ochameau:
Sorry, I introduced this regression via #510...
We end up with a appStatus=0 for system app because of the usage of the browser-specific workaround.

rpl:
closing because #557 fix the regression (which is much better than a workaround)

PR #556, opened 2013-05-23
FIX: clear appCache entries on run/update app requests using Services.evictEntries

### LINKED ISSUE #66: Clearing App Cache
Seems like apps don't have a separated cache (correct me if I am wrong) in B2G, which makes cleaning cache for just one app _hard_.

Some ideas:
- Clearing the cache when the user _updates_ an app. Might have unexpected results, as other apps that weren't updated will also be reset.
- An option to `Clear Cache on Launch`, similar to `JS Console`.
- Alternatives, like changing install flow (to give apps a new ID on install), experimenting.

What are your thoughts on that, @dangoor, @mykmelez ?

rpl:
clear appCache entries on run app requests on run / update user requests (should fix #66)

### DIFF
--- a/prosthesis/content/dbg-simulator-actors.js
+++ b/prosthesis/content/dbg-simulator-actors.js
@@ -108,16 +108,13 @@ SimulatorActor.prototype = {
     let appOrigin = DOMApplicationRegistry.webapps[appId].origin;
     let localId = DOMApplicationRegistry.webapps[appId].localId;
 
-    if (localId) {
-      debug("RUNAPP: clear all associated appCache entries");
-      DOMApplicationRegistry._clearPrivateData(localId, false);
-    }
-
     let debug = this.debug.bind(this);
 
     let runnable = {
       run: function() {
         try {
+          runnable.cleanAppCache();
+
           runnable.waitHomescreenReady(function () {
             runnable.tryAppReloadByOrigin(appOrigin, function () {
               runnable.findAppByOrigin(appOrigin, function (e, app) {
@@ -138,6 +135,26 @@ SimulatorActor.prototype = {
           });
         } catch(e) {
           debug(["EXCEPTION:", e, e.fileName, e.lineNumber].join(' '));
+        }
+      },
+      cleanAppCache: function() {
+        if (!localId) {
+          return;
+        }
+
+        debug("RUNAPP: clear all associated appCache entries");
+        try {
+          let cacheService = Cc["@mozilla.org/network/application-cache-service;1"].
+          getService(Ci.nsIApplicationCacheService);
+          cacheService.discardByAppId(localId, false);
+        } catch(e) {
+          // NOTE: currently cacheService.discardByAppId always raise an expection
+          // even if it's working correctly (and cache entries are refreshed on app
+          // window reload)
+          // EXCEPTION on localId NNNN: [Exception... "Component returned failure code: 0x8000ffff \
+          // (NS_ERROR_UNEXPECTED) [nsIApplicationCacheService.discardByAppId]" \
+          // nsresult: "0x8000ffff (NS_ERROR_UNEXPECTED)" ...
+          // debug("EXCEPTION on localId " + localId + ": " + e);
         }
       },
       waitHomescreenReady: function(cb) {
--- a/prosthesis/content/dbg-simulator-actors.js
+++ b/prosthesis/content/dbg-simulator-actors.js
@@ -102,10 +102,16 @@ SimulatorActor.prototype = {
     let appId = aRequest.appId;
 
     if (!DOMApplicationRegistry.webapps[appId]) {
-      return { success: false, error: 'app-not-installed', message: 'App not installed.'}
+      return { success: false, error: 'app-not-installed', message: 'App not installed.'};
     }
 
     let appOrigin = DOMApplicationRegistry.webapps[appId].origin;
+    let localId = DOMApplicationRegistry.webapps[appId].localId;
+
+    if (localId) {
+      debug("RUNAPP: clear all associated appCache entries");
+      DOMApplicationRegistry._clearPrivateData(localId, false);
+    }
 
     let debug = this.debug.bind(this);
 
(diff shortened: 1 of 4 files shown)


prosthesis/content/dbg-simulator-actors.js
```diff
@@ -159,6 +159,8 @@ SimulatorActor.prototype = {
         }
       },
       tryAppReloadByOrigin: function(origin, cb) {
+        debug("RUNAPP: clear all appCache entries");
+        Services.cache.evictEntries(Ci.nsICache.STORE_OFFLINE);
```
ochameau: By doing so, you will remove cache for all apps.
You can focus on one particular app by doing something like:

```
var cacheService = Cc["@mozilla.org/network/application-cache-service;1"].
                            getService(Ci.nsIApplicationCacheService);
cacheService.discardByAppId(appId, false);
```

See this IDL for more info:
http://mxr.mozilla.org/mozilla-central/source/netwerk/base/public/nsIApplicationCacheService.idl#72

rpl: awesome! it's much much better :-)

in the last change I've updated this pull request to use DOMApplicationRegistry._clearPrivateData (which sends an async request to an application cache service observer)

rpl:
@ochameau awesome! it's much much better :-)

in the last change I've updated this pull request to use DOMApplicationRegistry._clearPrivateData (which sends an async request to an application cache service observer)

ochameau:
You should not call clearPrivateData either ;-)
As it also remove app permission:
http://mxr.mozilla.org/mozilla-central/source/extensions/cookie/nsPermissionManager.cpp#172
`cacheService.discardByAppId` does **exactly** what you want to do, no more, no less.

rpl:
@ochameau I've pushed another change:
now I use cacheService.dicardByAppId directly.

I was a bit confused because it always raises an exception
and I thought I'm not using it correctly (and I've not found any javascript code on mxr which it actually call it directly as we do now),
even if it actually works (which I know because all appCache entries are refreshed on an app window reload)

I put a comment so we can remember why it's surrounded by its own try/catch but we don't print its exception into the debugging logs.

mykmelez:
I don't think we should do this by default, since that would make it impossible to test ordinary reinstallation of an app, which doesn't clear these entries. But I do think we should make it possible to do this. Now that we have a Ctrl/Cmd-R shortcut for refreshing an app, how about we add a Shift-Ctrl/Cmd-R variant that also clears this cache?

rpl:
@mykmelez I like it!
 just pushed: added "Refresh App and Clear AppCache" menu item and "shift
accel R" keyboard shortcut

mykmelez:
Hmm, this feature is hidden in the browser, and I think we should do the same in the Simulator, at least initially. I.e. add the keyboard shortcut but don't add a menu item for it. That makes it less discoverable, which I don't like. But it makes it possible to evolve the feature a bit before we expose it more (f.e. to clear other caches), which I do like. Since it feels like there's more work to do here to figure out exactly what developers want from a "refresh harder" command!

rpl:
@mykmelez sounds good, I've removed menu item and its locale entities.

prosthesis/content/shell.js
```diff
@@ -104,3 +110,25 @@ function simulatorAppUpdate() {
     }
   }, "r2d2b2g:app-update", null);
 }
+
+function simulatorClearAppCache(appId) {
+    let localId = DOMApplicationRegistry.webapps[appId].localId;
```
mykmelez: This function should use two-space indent.

prosthesis/content/shell.js
```diff
@@ -104,3 +110,25 @@ function simulatorAppUpdate() {
...
+
+    if (!localId) {
+        return;
+    }
+
+    try {
+        let cacheService = Cc["@mozilla.org/network/application-cache-service;1"].
+            getService(Ci.nsIApplicationCacheService);
+        cacheService.discardByAppId(localId, false);
```
mykmelez: Nit: in Mozilla code such multi-line statements are conventionally formatted:

```
    let cacheService = Cc["@mozilla.org/network/application-cache-service;1"].
                       getService(Ci.nsIApplicationCacheService);
    cacheService.discardByAppId(localId, false);
```

But in this case it isn't useful to assign the result to a local variable since it's only ever used once, and this can be simply:

```
    Cc["@mozilla.org/network/application-cache-service;1"].
      getService(Ci.nsIApplicationCacheService).discardByAppId(localId, false);
```

prosthesis/content/shell.js
```diff
@@ -104,3 +110,25 @@ function simulatorAppUpdate() {
...
+    } catch(e) {
+        // NOTE: currently cacheService.discardByAppId always raise an expection
+        // even if it's working correctly (and cache entries are refreshed on app
+        // window reload)
+        // EXCEPTION on localId NNNN: [Exception... "Component returned failure code: 0x8000ffff \
+        // (NS_ERROR_UNEXPECTED) [nsIApplicationCacheService.discardByAppId]" \
+        // nsresult: "0x8000ffff (NS_ERROR_UNEXPECTED)" ...
+        // debug("EXCEPTION on localId " + localId + ": " + e);
```
mykmelez: The last four lines of this comment look like something left over from an earlier comment. Are they still meant to be here?

prosthesis/content/shell.xul
```diff
@@ -44,12 +44,17 @@
   <commandset id="mainCommandSet">
     <command id="quitSimulatorCmd" oncommand="goQuitApplication()"/>
     <command id="refreshAppCmd" oncommand="simulatorAppUpdate()"/>
+    <command id="refreshAppAndClearAppCacheCmd" oncommand="simulatorAppUpdate(true)"/>
```
mykmelez: Nit: it'd be nice if _refreshAppAndClearAppCache_ was a bit shorter. The equivalent in Firefox is _ReloadSkipCache_, so perhaps this could be _refreshClearCacheCmd_?

prosthesis/content/shell.js
rpl: @mykmelez here I'm not sure how we can handle this "issue" (which apparently didn't present any side-effects):
- just leave this comment (to remember why we catch an exception without report any error)
- print an error linked to an open issue number on github to track this problem (and to remember that we're aware of the issue)

mykmelez:
> @mykmelez here I'm not sure how we can handle this "issue" (which apparently didn't present any side-effects):
> - just leave this comment (to remember why we catch an exception without report any error)
> - print an error linked to an open issue number on github to track this problem (and to remember that we're aware of the issue)

I think a comment is fine! I just thought perhaps the first three lines of the comment where what you intended to include, while the last four lines were for your help while debugging the problem. But I understand now that you want to include all that information in the comment, which is not a problem!

In that case, the only line I might remove is the last one, since it is just a commented-out _debug_ call rather than actual information about the problem.

PR #563, opened 2013-05-25
Fixed a typo (aDir => appDir)

bkase:
I think this must have been an artifact of an older version of the code where the `appDir` variable was called `aDir`


ochameau:
Thanks for identifying this!
That would be something to also fix on upstream:
http://mxr.mozilla.org/mozilla-central/source/b2g/chrome/content/dbg-webapps-actors.js#289

ochameau:
I opened a bug to fix that in upstream - https://bugzilla.mozilla.org/show_bug.cgi?id=876122

PR #567, opened 2013-05-27
Bypass devtools connect page by automatically connecting to shell.xul.

ochameau:
Here is a patch to bypass the connect page.
It uses the existing remote connection to open a toolbox connected to toplevel shell.xul document. As far as I can tell that's the only target that is confirmed to work.

This patch isn't perfect as it kills the possibility to select another context other than shell.xul, but given that we can't select anything else, it would still be an incremental improvement.

The only issue is that it depends on a feature introduced in platform bug https://bugzilla.mozilla.org/show_bug.cgi?id=875104
Without this, the remote connection will be automatically closed when the user closes the toolbox. We can workaround this by using a new remote connection just for the toolbox, but I'd prefer keeping only one shared connection to avoid introducing bugs related to connection issues.
To ease that, I can easily factor out the toolbox "keepTargetAlive" feature, in order to land this feature and eventually uplift it.
In anycase, this feature would only exists in FF23/24.

Otherwise, there is various things to do after that:
- Open the toolbox in the same window, as a regular devtool toolbox
- Fix webapps actor to list apps actors and be able to target a given app context
- Add UI in the dashboard to connect to a precise app

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -76,6 +76,7 @@ const RemoteSimulatorClient = Class({
       let client = data.client;
       this.once("kill", function () client.close());
       client.request({to: "root", type: "listTabs"}, (function (reply) {
+        this._rootActor = reply;
         emit(this, "clientReady", {
           client: client,
           globals: reply,
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -924,10 +924,21 @@ let simulator = module.exports = {
   },
 
   openConnectDevtools: function() {
+    // We need to workaround existing devtools TabTarget.destroy code,
+    // that tries to close the client when the related toolbox is closed
+    // whereas we want to keep the client alive for other usages!
+    // http://hg.mozilla.org/mozilla-central/annotate/cfcce7c5eb74/browser/devtools/framework/target.js#l427
+    let clientProxy = new Proxy(this.remoteSimulator._remote.client, {
+      get: function (target, name) {
+        if (name == "close")
+          return function () {};
+        return target[name];
+      }
+    });
     // Open a remote toolbox to shell.xul
     let options = {
       form: this.remoteSimulator._rootActor,
-      client: this.remoteSimulator._remote.client,
+      client: clientProxy,
       chrome: true
     };
     let {gDevTools} = Cu.import("resource:///modules/devtools/gDevTools.jsm", {});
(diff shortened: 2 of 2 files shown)


mykmelez:
@ochameau Am I correct in thinking that bug 875104 should land first before we merge this change?

ochameau:
> @ochameau Am I correct in thinking that bug 875104 should land first before we merge this change?

You are 100% right!

ochameau:
As I'd really like to see this fixed in v4, I pushed a new changeset to remove to dependency to the platform fix.
I also tested this patch on FF 21, 22, 23 and 24 and fixed issues happening on 21.

I'll submit another patch when the related platform patch gets landed in order to eventually remove the workaround.

mykmelez:
This appears to break showing sources in the Debugger pane of the toolbox, which reports that "this page has no sources" when I open a toolbox with this change applied.

mykmelez:
(Whereas when I open a toolbox for shell.xul without the change, the Debugger pane shows a bunch of sources.)

ochameau:
A quick update here. I told you I'd possibly have a toolbox connected to the App scope.
I have something finally working! But I haven't had time to put the patch in a reviewable shape.
I also figured out that the "Main process" target displayed in the connect doesn't get a working Debugger panel (the console work fine). We get a consistent behavior against b2g desktop shipped in the simulator, but also on devices. So there is something wrong here at devtools level.
I'll try to see with Panos the underlying issue behind this. I think that the shell.xul actor being explicitely created and returned in the `tabs` actor list is a workaround for this issue. As Main process and shell.xul should end up being the same thing...

I'll get back to this on Monday. Otherwise, I was considering submitting a bit more substantial patch that will remove the global Connect button in favor of a connect button next to each app. I think that's what apps developers are expecting. But do you think we should keep offering ways to connect to shell.xul and all builtin apps? I'm open minded here. I think we have a big opportunity of improvement here and would like to avoid modifying devtools behavior too many times.

mykmelez:
> Otherwise, I was considering submitting a bit more substantial patch that will remove the global Connect button in favor of a connect button next to each app. I think that's what apps developers are expecting. But do you think we should keep offering ways to connect to shell.xul and all builtin apps? I'm open minded here. I think we have a big opportunity of improvement here and would like to avoid modifying devtools behavior too many times.

If we can connect a toolbox to a specific app, then no, I don't think we should keep offering ways to connect to all apps (nor chrome). At least not visibly.

It's unlikely that developers will want to debug more than a single app at a time; so that's the 80% case, for which an app-specific toolbox is optimal. If they do, however, it will probably only be a couple apps (like two apps that use activities to interact); which is the 95% case and is adequately handled by app-specific toolboxes.

I do think a global toolbox is helpful to Simulator developers. So I wouldn't mind still having that option. But there's no reason to expose app developers to it.

For that matter: once we do this, we should also hide the (Error) Console checkbox, which is similarly useful for Simulator developers but not for app developers who have an app-specific Web Console.

mykmelez:
Now that we have "connect to app", this is obsolete, so I'm going to close the request!

PR #568, opened 2013-05-27
Import style editor actor via prosthesis.

ochameau:
Here is a tentative to load styleeditor actor via prosthesis.
Unfortunately, for some reason it doesn't work. We end up with a style editor panel with infinite loading throbbers.

### DIFF
--- a/prosthesis/content/dbg-styleeditor-actors.js
+++ b/prosthesis/content/dbg-styleeditor-actors.js
@@ -1,0 +1,741 @@
+// https://hg.mozilla.org/releases/mozilla-aurora/annotate/ea2b0db82289/toolkit/devtools/styleeditor/dbg-styleeditor-actors.js
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+let Cc = Components.classes;
+let Ci = Components.interfaces;
+let Cu = Components.utils;
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/NetUtil.jsm");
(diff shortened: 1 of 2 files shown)


prosthesis/content/dbg-styleeditor-actors.js
```diff
@@ -0,0 +1,741 @@
...
+  },
+
+  /**
+   * Clear all the current stylesheet actors in map.
+   */
+  _clearStyleSheetActors: function() {
+    for (let actor in this._sheets) {
+      this.releaseActor(this._sheets[actor]);
+    }
+    this._sheets.clear();
```
mykmelez, line 262: The problem turns out to be simple: according to the [Map documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), Map.clear() was implemented in Mozilla 19, so it isn't available on the b2g18 branch, where we're using this code, which causes this call to throw an exception. The fix is to catch the exception and use a less efficient but backward-compatible loop to clear the sheets (both here and in _disconnect_):

```
    try {
      this._sheets.clear();
    } catch(e) {
      // Mozilla 18 doesn't have Map.clear(), which was introduced in 19.
      for (let actor in this._sheets) {
        this._sheets.delete(actor);
      }
    }
```

mykmelez:
The only other issue with this change is that the line-endings in dbg-styleeditor-actors.js are Windows-style \r\n pairs instead of conventional Unix-style \n characters. I'll merge the change with fixes for both issues.

ochameau:
Oh, awesome! Thanks for having looked at that, I wonder how I missed that, was it silently failing ?

@harthur @campd Look, we have the style editor for gecko18 \o/
https://github.com/mozilla/r2d2b2g/blob/master/prosthesis/content/dbg-styleeditor-actors.js

mykmelez:
> Oh, awesome! Thanks for having looked at that, I wonder how I missed that, was it silently failing ?

Nope, it was throwing an exception to the web consoles I connected to the main process and shell.xul! But perhaps that exception wasn't showing up in the Error Console?

Issue #571, opened 2013-05-29
Simulator actors don't initialize until the dashboard sends an app


bkase:
Simulator actors in `prosthesis/content/dbg-simulator-actors.js` are not initialized until the dashboard runs an app.

This means that the shell in prosthesis is unable to communicate via observers to the simulator actors until an app is run from the dashboard. The simulator actors should be initialized prior to any observer communication.

bkase:
We could just send a wakeup message to the simulator actors when someone toggles the simulator to running. 
Is this too hacky?

mykmelez:
It isn't too hacky, although we should still also figure out the underlying problem.

nickdesaulniers:
fixed in #653

PR #573, opened 2013-05-29
Refresh hosted apps installed outside of dashboard. Fixes #443

### LINKED ISSUE #443: refresh hosted app that wasn't installed by Dashboard
If a hosted app wasn't installed by the Dashboard, the Simulator should still have enough information about the app to reinstall and restart it, so it should do so when the user selects Refresh. That way a user who installs a hosted app directly in the Simulator, for whatever reason (f.e. @nickdesaulniers's [test app with receipt use case](https://github.com/mozilla/r2d2b2g/issues/422#issuecomment-16589426)), can use Refresh to refresh it!

@rpl What do you think?

bkase:
Refreshing now works with hosted apps installed outside of the dashboard.
Try it on any app that calls `mozApps.install` or a marketplace app post-installation.

For now, it won't work until the simulator actors are initialized (see #571), so before refreshing any app, you must first run something from the dashboard.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -1088,7 +1088,6 @@ let simulator = module.exports = {
         });
 
         if (!foundAppKey) {
-          console.log("Sending appNotFound: " + appId);
           simulator.remoteSimulator.appNotFound(appId);
         }
 
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -1088,7 +1088,8 @@ let simulator = module.exports = {
         });
 
         if (!foundAppKey) {
-          simulator.remoteSimulator.showNotification("App not updated (not found)");
+          console.log("Sending appNotFound: " + appId);
+          simulator.remoteSimulator.appNotFound(appId);
         }
 
         simulator.updateApp(foundAppKey, function next(error, app) {
--- a/prosthesis/content/dbg-simulator-actors.js
+++ b/prosthesis/content/dbg-simulator-actors.js
@@ -4,10 +4,6 @@
 
 "use strict";
 
-let Cu = Components.utils;
-let Cc = Components.classes;
-let Ci = Components.interfaces;
-
 Cu.import("resource://gre/modules/Services.jsm");
 let { AppsUtils } = Cu.import("resource://gre/modules/AppsUtils.jsm");
 
@@ -288,7 +284,7 @@ SimulatorActor.prototype = {
       let contentType = req.getResponseHeader("content-type");
 
       if (AppsUtils.checkManifestContentType(installOrigin, origin, contentType) &&
-          manifestURL.indexOf('http://') === 0) {
+          manifestURL.indexOf('app:') !== 0) {
         onGoodContentType();
       } else {
         onBadContentType(contentType);
@@ -340,7 +336,7 @@ SimulatorActor.prototype = {
         uninstallReq.onsuccess = function onsuccess() {
           // Purge app cache
           try {
-            // This seems to legitamtely fail
+            // This seems to legitimately fail
             // (not actually flush the cache) if an app is:
             //    1. Installed from the dashboard
             //    2. Removed from the dashboard
--- a/prosthesis/content/dbg-simulator-actors.js
+++ b/prosthesis/content/dbg-simulator-actors.js
@@ -4,7 +4,12 @@
 
 "use strict";
 
-Components.utils.import("resource://gre/modules/Services.jsm");
+let Cu = Components.utils;
+let Cc = Components.classes;
+let Ci = Components.interfaces;
+
+Cu.import("resource://gre/modules/Services.jsm");
+let { AppsUtils } = Cu.import("resource://gre/modules/AppsUtils.jsm");
 
 /**
   * Creates a SimulatorActor. SimulatorActor provides remote access to the
@@ -269,24 +274,24 @@ SimulatorActor.prototype = {
 
   // ensure that an app can be installed by examining the Content-Type
   // header returned by HEAD request to the manifestURL
-  _ensureInstallable: function(manifestURL, cbs) {
-    const MANIFEST_REGEX = /Content\-Type: application\/x\-web\-app\-manifest\+json/;
+  _ensureInstallable: function(app, onBadContentType, onGoodContentType) {
+    let { manifestURL, origin, installOrigin } = app;
 
-    var httpreq = Components.classes["@mozilla.org/xmlextras/xmlhttprequest;1"]
-      .createInstance();
-    httpreq.open('HEAD', manifestURL, true);
-    httpreq.send('');
-    httpreq.onreadystatechange = function() {
-      let rawHeaders = httpreq.getAllResponseHeaders() || "";
-
-      if (httpreq.readyState !== 4) {
+    let req = Cc["@mozilla.org/xmlextras/xmlhttprequest;1"].createInstance();
+    req.open('HEAD', manifestURL, true);
+    req.send('');
+    req.onreadystatechange = function() {
+      if (req.readyState !== 4) {
         return;
       }
 
-      if (MANIFEST_REGEX.test(rawHeaders)) {
-        cbs.onGoodContentType();
+      let contentType = req.getResponseHeader("content-type");
+
+      if (AppsUtils.checkManifestContentType(installOrigin, origin, contentType) &&
+          manifestURL.indexOf('http://') === 0) {
+        onGoodContentType();
       } else {
-        cbs.onBadContentType();
+        onBadContentType(contentType);
       }
     };
   },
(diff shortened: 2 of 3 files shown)


addon/lib/simulator.js
```diff
@@ -1088,7 +1088,8 @@ let simulator = module.exports = {
         });
 
         if (!foundAppKey) {
-          simulator.remoteSimulator.showNotification("App not updated (not found)");
+          console.log("Sending appNotFound: " + appId);
```
mykmelez: Nit: it'd be better to log this message inside _appNotFound_, so we log it for future consumers of that function as well.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -260,10 +266,117 @@ SimulatorActor.prototype = {
     };
   },
 
+  // ensure that an app can be installed by examining the Content-Type
+  // header returned by HEAD request to the manifestURL
+  _ensureInstallable: function(manifestURL, cbs) {
+    const MANIFEST_REGEX = /Content\-Type: application\/x\-web\-app\-manifest\+json/;
+
+    var httpreq = Components.classes["@mozilla.org/xmlextras/xmlhttprequest;1"]
+      .createInstance();
```
mykmelez: Nits: per convention, use _let_ to declare variables; replace _Components.classes_ with its shortcut _Cc_; place the member operator (i.e. the dot) at the end of the previous line rather than the beginning of the next one when breaking a member reference across lines; and align the member reference in the second line with its referent in the first one, i.e.:

```
    let httpreq = Cc["@mozilla.org/xmlextras/xmlhttprequest;1"].
                  createInstance();
```

Finally, I'm not sure _httpreq_ is the best name for this variable, since it isn't necessarily an HTTP request (at the very least, it could also be HTTPS; and the function parameter just takes a generic URL). I would call it simply _request_ or _req_.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+    httpreq.open('HEAD', manifestURL, true);
+    httpreq.send('');
+    httpreq.onreadystatechange = function() {
+      let rawHeaders = httpreq.getAllResponseHeaders() || "";
+
+      if (httpreq.readyState !== 4) {
+        return;
+      }
+
+      if (MANIFEST_REGEX.test(rawHeaders)) {
```
mykmelez: Instead of using a regex, use _nsINetUtil::parseContentType_:

```
    const MANIFEST_TYPE = "application/x-web-app-manifest+json";
    …
    let netUtil = Cc["@mozilla.org/network/util;1"].getService(Ci.nsINetUtil);
    let contentType = httpreq.getResponseHeader("content-type");
    if (netUtil.parseContentType(contentType, {}, {}) == MANIFEST_TYPE) {
      …
    }
    else {
      …
    }
```

Also, you might create a global lazy getter for _netUtil_, in which case I would capitalize its initial letter to more clearly identify it as a global singleton, i.e.:

```
XPCOMUtils.defineLazyGetter(this, "NetUtil", function() {
  return Cc["@mozilla.org/network/util;1"].getService(Ci.nsINetUtil);
});
```

( _XPCOMUtils_ comes from the XPCOMUtils.jsm JavaScript Module, which gets imported by the debugger server into the same scope in which it loads these actors.)

Finally, note that [AppsUtils::checkManifestContentType](https://github.com/mozilla/mozilla-central/blob/72c0478d6d447598ae6e899d940257cc420e560a/dom/apps/src/AppsUtils.jsm#L341-L351) doesn't require manifests to have the `application/x-web-app-manifest+json` type if their app origin is the same as their install origin. Perhaps we should do the same here (or simply defer to that function).

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+      this._displayNotification("App not updated (not found)");
+      return {};
+    }
+
+    let actor = this;
+    let debug = this.debug.bind(this);
+    this._ensureInstallable(manifestURL, {
+
+      onBadContentType: function() {
+        debug("incorrect content-type");
```
mykmelez: Nit: it'd be useful to add the bad content-type to this debug message, passing it from __ensureInstallable_ into this function.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+        actor._displayNotification("App not updated (not found)");
```
mykmelez: This message should be: "App not reinstallable (bad content-type " + contentType + ")".

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+    if (!origin || !manifestURL) {
+      this._displayNotification("App not updated (not found)");
+      return {};
+    }
+
+    let actor = this;
+    let debug = this.debug.bind(this);
+    this._ensureInstallable(manifestURL, {
+
+      onBadContentType: function() {
```
mykmelez: Nit: we're inconsistent about this, but it's useful to name functions, even when they're methods of objects, as in this case, because debuggers aren't always able to identify the function by method name, i.e.:

```
      onBadContentType: function onBadContentType() {
```

mykmelez: I also wonder why the callbacks are passed in an object. It seems like it would be simpler to just pass the two functions as the second and third arguments to __ensureInstallable_.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+        debug("incorrect content-type");
+        actor._displayNotification("App not updated (not found)");
+      },
+
+      onGoodContentType: function() {
+        // Uninstall
+        try {
```
mykmelez: This _try/catch_ block is humongous, which makes the code harder to read. And it doesn't need to be, since we don't expect setting _req.onerror_ and _req.onsuccess_ to fail. So it would be better to surround just the `mgmt.uninstall({origin: origin})` call with the _try/catch_ block.

However, I don't think the block is actually necessary at all, since [mozIDOMApplicationMgmt::uninstall](https://github.com/mozilla/mozilla-central/blob/c238bf4cfbd5f81fe4fe07a4da86f26334b87fc5/dom/apps/src/Webapps.js#L691-L699) never throws. Of course it might trigger an exception anyway because of a bug in the code, but we don't typically catch such errors and expose users to them. So I would remove the block and report uninstallation failure only in the _onerror_ callback.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+    }
+
+    let manifestURL = app.manifestURL;
+    let origin = app.origin;
+    if (!origin || !manifestURL) {
+      this._displayNotification("App not updated (not found)");
+      return {};
+    }
+
+    let actor = this;
```
mykmelez: This is ok, but in general we've been migrating from local variables like this one to function binding, i.e. 

```
      onBadContentType: (function() {
        this.debug("incorrect content-type");
        this._displayNotification("App not updated (not found)");
      }).bind(this),
```

Which also gives you _this.debug_ without having to define _debug_.

mykmelez: Hmm, although I see that _this_ is meaningful in _nsIDOMDOMRequest::onsuccess/onerror_ callbacks, where it references the request. That makes the _actor_ local more useful, although I might still bind those functions, since they close over scopes that bind _req_ and _req__ to the requests in question.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+                .getService(Components.interfaces.nsIApplicationCacheService)
+                .discardByAppId(appId, false);
+            } catch(e) {
+              // This error is always thrown even if cacheService.discardByAppId
+              // is working. See:
+              // https://github.com/mozilla/r2d2b2g/pull/556#issuecomment-18351200
+            }
+
+            // Re-install
+            let req_ = window.navigator.mozApps.install(manifestURL);
```
mykmelez: You can actually call both requests _req_, since the inner scope never accesses the outer scope's _req_. But it would be safer (and make the code easier to read) to give them more descriptive names, like _uninstallReq_ and _reinstallReq_.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+              // This seems to legitamtely fail
+              // (not actually flush the cache) if an app is:
+              //    1. Installed from the dashboard
+              //    2. Removed from the dashboard
+              //    3. Installed from outside of the dash
+              //    4. (then refreshed to get to this call)
+              Components.classes
+                ["@mozilla.org/network/application-cache-service;1"]
+                .getService(Components.interfaces.nsIApplicationCacheService)
+                .discardByAppId(appId, false);
```
mykmelez: Nit: use shortcuts and format such a multi-line statement as follows:

```
              Cc["@mozilla.org/network/application-cache-service;1"].
                getService(Ci.nsIApplicationCacheService).
                discardByAppId(appId, false);
```

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+              let appId_ = DOMApplicationRegistry._appId(this.result.origin);
+              actor._runApp.call(actor, appId_);
+            };
+          };
+        } catch(e) {
+          actor._displayNotification("App not updated (uninstallation failed)");
+          Cu.reportError(e);
+        }
+      }
+   });
```
mykmelez: Nit: it looks like this line needs to be indented one more space.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +267,119 @@ SimulatorActor.prototype = {
...
+    if (!app) {
+      this._displayNotification("App not updated (not found)");
+      return {};
+    }
+
+    let manifestURL = app.manifestURL;
+    let origin = app.origin;
+    if (!origin || !manifestURL) {
+      this._displayNotification("App not updated (not found)");
+      return {};
```
mykmelez: Nit: I think we should to return the same response object here and above as we do below, i.e.:

```
    return {
      message: "appNotFound received",
      success: true
    };
```

Perhaps we can factor it out to a variable we define at the top of the function and then reference as needed.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -4,7 +4,12 @@
 
 "use strict";
 
-Components.utils.import("resource://gre/modules/Services.jsm");
+let Cu = Components.utils;
+let Cc = Components.classes;
+let Ci = Components.interfaces;
```
mykmelez: It isn't actually necessary to define the C\* shortcuts, as this script is loaded into the debugger server's context, and [that context's global defines them](https://github.com/mozilla/mozilla-central/blob/e77a5a3fa104d724a927c4d48890a87d18856959/toolkit/devtools/debugger/server/dbg-server.js#L13-L17).

mykmelez:
I see a ton of these when I initiate a refresh, although the refresh proceeds and succeeds:

Timestamp: 2013/05/31 15:38:10
Error: msg is undefined
Source File: file:///Users/myk/Profiles/r2d2b2g/extensions/r2d2b2g@mozilla.org//resources/r2d2b2g/data/mac64/B2G.app/Contents/MacOS/components/Webapps.js
Line: 533

mykmelez:
> _ensureInstallable also needs to check for http:// in order to prevent non-hosted apps from returning installable.

This seems brittle. At the very least, it should also check for https:. But a better approach would be to check for app:, since I think all packaged apps will have an app: origin.

mykmelez:
Aside: we probably do want to refresh packaged apps, since, as @nickdesaulniers [noted](https://github.com/mozilla/r2d2b2g/issues/443#issuecomment-16617614), they might be installed from the Marketplace. But we can do that in a separate issue.

prosthesis/content/dbg-simulator-actors.js
```diff
@@ -261,10 +272,115 @@ SimulatorActor.prototype = {
...
+        let actor = this;
+        let mgmt = window.navigator.mozApps.mgmt;
+        let uninstallReq = mgmt.uninstall({origin: origin});
+        uninstallReq.onerror = function onerror() {
+          actor._displayNotification("App not updated (uninstallation failed)");
+        }
+        uninstallReq.onsuccess = function onsuccess() {
+          // Purge app cache
+          try {
+            // This seems to legitamtely fail
```
mykmelez: Nit: legitamtely -> legitimately

mykmelez:
Other than those last few issues, this looks good!

I'll want to do some iteration on the error messages, but that can be a separate issue.

mykmelez:
Oh: and ideally we wouldn't prompt the user to install the app. But I'm not sure how to bypass that prompt. We can tackle this separately.

bkase:
> Timestamp: 2013/05/31 15:38:10
> Error: msg is undefined
> Source File: file:///Users/myk/Profiles/r2d2b2g/extensions/r2d2b2g@mozilla.org//resources/r2d2b2g/data/mac64/B2G.app/Contents/MacOS/components/Webapps.js
> Line: 533

I wasn't able to reproduce this -- is there a specific app refresh that causes this?

mykmelez:
> I wasn't able to reproduce this -- is there a specific app refresh that causes this?

I see it consistently with Pasjans Solitaire, which I installed from Marketplace. The messages show up in the Error Console immediately after I initiate refresh, and there are dozens to hundreds of them. See also [bug 878388](https://bugzilla.mozilla.org/show_bug.cgi?id=878388), which reports the same problem (although on a different line) with app uninstalls.

mykmelez:
> I see it consistently with Pasjans Solitaire, which I installed from Marketplace. The messages show up in the Error Console immediately after I initiate refresh, and there are dozens to hundreds of them. See also bug 878388, which reports the same problem (although on a different line) with app uninstalls.

But I just tested manual app uninstall on the master branch, and I see the problem there too, so it isn't specific to this change, and we don't have to address it in this pull request.

bkase:
> But I just tested manual app uninstall on the master branch, and I see the problem there too, so it isn't specific to this change, and we don't have to address it in this pull request.

OK it should be ready then, I addressed all the other issues.
(I still don't see the error in the Error Console though, even when refreshing Pasjans Solitaire)

PR #574, opened 2013-05-30
Missing bind for two debug statements

bkase:
`this.debug` was called without rebinding `this` in a place where it was defined incorrectly.


mykmelez:
Normally we would bind the this-referencing functions directly, but that would mean binding their containing function too in this case, so this approach is reasonable.

PR #578, opened 2013-05-30
Update receipts

nickdesaulniers:
review? @mykmelez 
references #494 
test with `make test`

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -926,8 +926,11 @@ let simulator = module.exports = {
     for each (let tab in Tabs) {
       if (tab.url === url || (lax && tab.url.indexOf(url) === 0)) {
         tab.activate();
-        tab.attach({
-          onAttach: function attached(worker) {
+        // Hacky workaround to ensure the tab is ready before we call
+        // the callback.  Implement better approach after fixing SDK bug 879534!
+        let worker = tab.attach({
+          contentScript: "self.postMessage()",
+          onMessage: function() {
             worker.destroy();
             cb(tab);
           },
@@ -948,8 +951,11 @@ let simulator = module.exports = {
     if (this.worker) {
       let tab = this.worker.tab;
       tab.activate();
-      tab.attach({
-        onAttach: function attached(worker) {
+      // Hacky workaround to ensure the tab is ready before we call
+      // the callback.  Implement better approach after fixing SDK bug 879534!
+      let worker = tab.attach({
+        contentScript: "self.postMessage()",
+        onMessage: function() {
           worker.destroy();
           cb(tab);
         },
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -457,26 +457,21 @@ let simulator = module.exports = {
   },
 
   updateReceiptType: function updateReceiptType(appId, receiptType) {
-    let app = simulator.apps[appId];
-    let manifestURL = simulator.receiptManifestURL(app.type,
-                                                   app.origin || app.xkey);
-    let next = function next() {
-      simulator.sendListApps();
-      simulator.runApp(app);
-    };
-
+    let app = this.apps[appId];
+    let manifestURL = this.receiptManifestURL(app.type,
+                                              app.origin || app.xkey);
     if (receiptType === "none") {
       app.receipt = null;
       app.receiptType = receiptType;
-      simulator._updateApp(appId, simulator.sendListApps.bind(this));
+      this._updateApp(appId, this.sendListApps.bind(this));
     } else {
-      simulator.fetchReceipt(manifestURL, receiptType, function fetched(err, receipt) {
+      this.fetchReceipt(manifestURL, receiptType, function fetched(err, receipt) {
         if (err || !receipt) {
           console.error(err || "No receipt");
         } else {
           app.receipt = receipt;
           app.receiptType = receiptType;
-          simulator._updateApp(appId, simulator.sendListApps.bind(this));
+          this._updateApp(appId, this.sendListApps.bind(this));
         }
       }.bind(this));
     }
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -922,24 +922,38 @@ let simulator = module.exports = {
   },
 
   openTab: function(url, lax, cb) {
-    for each (var tab in Tabs) {
+    cb = cb || function () {};
+    for each (let tab in Tabs) {
       if (tab.url === url || (lax && tab.url.indexOf(url) === 0)) {
         tab.activate();
+        tab.attach({
+          onAttach: function attached(worker) {
+            worker.destroy();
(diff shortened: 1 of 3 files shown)


addon/lib/simulator.js
```diff
@@ -452,6 +452,36 @@ let simulator = module.exports = {
...
+  },
+
+  updateReceiptType: function updateReceiptType(appId, receiptType) {
+    let app = simulator.apps[appId];
+    let manifestURL = simulator.receiptManifestURL(app.type,
+                                                   app.origin || app.xkey);
+    let next = function next() {
+      simulator.sendListApps();
+      simulator.runApp(app);
+    };
```
mykmelez: This function doesn't seem to get used.

addon/lib/simulator.js
```diff
@@ -452,6 +452,36 @@ let simulator = module.exports = {
     }
   },
 
+  receiptManifestURL: function receiptManifestURL(appType, appOrigin) {
+    return appType === "local" ? "https://" + appOrigin + ".simulator" : appOrigin;
+  },
+
+  updateReceiptType: function updateReceiptType(appId, receiptType) {
+    let app = simulator.apps[appId];
```
mykmelez: Nit: methods of the Simulator singleton should use `this` to refer to the singleton.

addon/lib/simulator.js
```diff
@@ -904,17 +934,20 @@ let simulator = module.exports = {
...
+  openHelperTab: function(cb) {
     // Ensure opening only one simulator page
-    if (this.worker)
+    if (this.worker) {
       this.worker.tab.activate();
-    else
-      this.openTab(simulator.contentPage, true);
+    } else {
+      this.openTab(simulator.contentPage, true, cb);
+    }
```
mykmelez, line 965: We should call the callback no matter how we open/activate the tab, i.e. for all of the three code paths that lead to the tab being activated (two of which involve activating a tab that is already opened). But it isn't obvious how to do this, since you can register a "ready" handler for a tab that isn't ready yet (i.e. that you are opening), but you can't find out whether or not an existing tab is already ready.

We _should_ be able to do something like this at the end of _openHelperTab_ to call the call back unconditionally on ready:

```
    if (typeof cb === "function") {
      tab.attach({
        contentScript: "",
        onAttach: function() {
          openArgs.onReady = cb;
        }
      });
    }
```

However, I'm having trouble testing this. When I run tests, they hang with:

```
info: r2d2b2g: executing 'r2d2b2g/tests/test-receipts.test receipt update'
undefined: undefined
undefined undefined
```

Even more bizarrely, the tests passed when I ran them the first time (without making any changes to your code). But now they fail even if I revert my changes. Something strange is happening.

mykmelez, line 965: Erm:

```
-          openArgs.onReady = cb;
+          cb();
```

But I'm still unable to test.

addon/test/test-receipts.js
```diff
@@ -0,0 +1,105 @@
...
+            '<body><p>Hello World!</p></body></html>');
+};
+
+function appFirstListed({ target, event, onInstall, onUpdate, equal }) {
+  let appInstalled = false;
+  let appUpdated = false;
+  target.addEventListener(event, function onEvent(evt) {
+    let message = evt.detail;
+    if (message.name === "listApps") {
+      console.log("\n\n\n\n\n\nlistApps");
```
nickdesaulniers: oops!

addon/lib/simulator.js
nickdesaulniers, line 965: This behavior is because I don't call `srv.stop()` so subsequent runs of the test, and thus subsequent calls to `srv.start()` throw.  What is weird is that when I called `srv.stop()` in the past, that would throw...  Let me look into that again.

nickdesaulniers, line 965: Actually, looks like forgetting to call simulator.unload is the issue

nickdesaulniers:
review? @mykmelez

addon/lib/simulator.js
```diff
@@ -896,25 +921,42 @@ let simulator = module.exports = {
...
+    cb = cb || function () {};
+    for each (let tab in Tabs) {
       if (tab.url === url || (lax && tab.url.indexOf(url) === 0)) {
         tab.activate();
+        tab.attach({
+          onAttach: function attached(worker) {
+            worker.destroy();
+            cb(tab);
+          },
+        });
```
mykmelez, line 937: This doesn't work, because _tabs.attach_ doesn't actually have an _onAttach_ option:

> 15:30:06 - erikvold: hmm I don't think tab.attach({onAttach: function() {}}) is implemented
> 15:30:12 - erikvold: onAttach doesn't seem to fire
> 15:30:17 - erikvold: this is working tho:
> 15:30:40 - erikvold: https://gist.github.com/erikvold/5710196

@erikvold says to do something like this, which works in my tests:

```
diff --git a/addon/lib/simulator.js b/addon/lib/simulator.js
index 548df9f..fd74613 100644
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -921,18 +921,19 @@ let simulator = module.exports = {
     });
   },

   openTab: function(url, lax, cb) {
     cb = cb || function () {};
     for each (let tab in Tabs) {
       if (tab.url === url || (lax && tab.url.indexOf(url) === 0)) {
         tab.activate();
-        tab.attach({
-          onAttach: function attached(worker) {
+        let worker = tab.attach({
+          contentScript: "self.postMessage()",
+          onMessage: function() {
             worker.destroy();
             cb(tab);
           },
         });
         return;
       }
     }

@@ -943,18 +944,19 @@ let simulator = module.exports = {
   },

   openHelperTab: function(cb) {
     cb = cb || function () {};
     // Ensure opening only one simulator page
     if (this.worker) {
       let tab = this.worker.tab;
       tab.activate();
-      tab.attach({
-        onAttach: function attached(worker) {
+      let worker = tab.attach({
+        contentScript: "self.postMessage()",
+        onMessage: function() {
           worker.destroy();
           cb(tab);
         },
       });
     } else {
       this.openTab(simulator.contentPage, true, cb);
     }
   },
```

But we should also add a comment to these code blocks explaining that they're a workaround until someone fixes SDK [bug 879534](https://bugzilla.mozilla.org/show_bug.cgi?id=879534), something like:

```
        // Hacky workaround to ensure the tab is ready before we call
        // the callback.  Implement better approach after fixing SDK bug 879534!
        let worker = tab.attach({
          contentScript: "self.postMessage()",
          onMessage: function() {
            worker.destroy();
            cb(tab);
          },
        });
```

nickdesaulniers, line 937: Does the onMessage callback get passed a reference to worker?

nickdesaulniers, line 937: nvm that's what `let worker =` is for...

mykmelez:
Woot!

Issue #247, opened 2013-01-03
Simulator fails to start when path contains accented character


wildfier:
Hi, I'm running on Arch Linux 64 bits and when I launch the simulator the phone screen is black with the mozilla logo. Nothing else except that B2G uses 11% of the CPU (I have a i7). I have the following errors :

Error: NS_ERROR_ILLEGAL_VALUE: Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIXPCComponents_Utils.import]
Source File: file:///media/media/Partag%C3%83%C2%A9/Firefox/wildfier/extensions/r2d2b2g@mozilla.org/resources/r2d2b2g/data/linux64/b2g/components/HealthReportService.js
Line: 10

Error: ERROR addons.xpi-utils: Failed to remove database that could not be opened: [Exception... "Component returned failure code: 0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST) [nsILocalFile.remove]"  nsresult: "0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST)"  location: "JS frame :: resource://gre/modules/XPIProvider.jsm -> resource://gre/modules/XPIProviderUtils.js :: XPIDB_openDatabaseFile :: line 491"  data: no]
Source File: resource://gre/modules/XPIProvider.jsm -> resource://gre/modules/XPIProviderUtils.js
Line: 491

Error: ERROR addons.xpi-utils: Failed to open database (2nd attempt): [Exception... "Component returned failure code: 0x80520015 (NS_ERROR_FILE_ACCESS_DENIED) [mozIStorageService.openUnsharedDatabase]"  nsresult: "0x80520015 (NS_ERROR_FILE_ACCESS_DENIED)"  location: "JS frame :: resource://gre/modules/XPIProvider.jsm -> resource://gre/modules/XPIProviderUtils.js :: XPIDB_openDatabaseFile :: line 497"  data: no]
Source File: resource://gre/modules/XPIProvider.jsm -> resource://gre/modules/XPIProviderUtils.js
Line: 497

Error: [Exception... "'Method not implemented' when calling method: [nsIDOMEventListener::handleEvent]"  nsresult: "0x80004001 (NS_ERROR_NOT_IMPLEMENTED)"  location: "native frame :: <unknown filename> :: <TOP_LEVEL> :: line 0"  data: no]

[...]

PR #582, opened 2013-06-04
Disable addons update.

ochameau:
I've noticed that sometimes we see the addon update popup when launching b2g.
And then, most of the time, the profile seems to be corrupted and won't run anymore.
I'm wondering if it is the origin of the issue I mentioned here:
https://github.com/mozilla/r2d2b2g/issues/247#issuecomment-18541800
Where we see the following exception in jsconsole `nsIDOMEventListener::handleEvent]" nsresult: "0x80004001 (NS_ERROR_NOT_IMPLEMENTED)` and gaia won't ever start again. (stay on mozilla logo)

I think that this pref will disable this.

### DIFF
--- a/build/override-prefs.js
+++ b/build/override-prefs.js
@@ -1,0 +1,1 @@
+user-pref("extensions.update.enabled", false);


build/override-prefs.js
```diff
@@ -0,0 +1 @@
+user-pref("extensions.update.enabled", false);
```
mykmelez, line 1: This would need to be _user_pref_ (with an underscore rather than a dash).

Also, nit: append a newline to the file.

However, this change doesn't actually suppress that dialog, which appears when you run a version of B2G with a profile that was previously used with a different version of B2G (f.e. B2G 18 and then 24.0a1 or vice-versa). To suppress the dialog, you must set the _extensions.showMismatchUI_ preference:

```
user_pref("extensions.showMismatchUI", false);
```

But suppressing the dialog doesn't actually solve the problem of a broken profile. When I set the aforementioned pref in a Gaia profile, start the Simulator's build of B2G 18 with the profile, and then start a custom build of B2G 24.0a1 with the profile, the app starts, but its window never appears, and only the following messages appear in the console:

```
desktop-helper: +++ loading scripts for app: resource://gre-resources/hiddenWindow.html

AlarmHalService: Constructor
```

Going from 24.0a1 to 18 also fails, although differently. It displays the window, but the window remains black, and the following messages appear in the console:

```
desktop-helper: +++ loading scripts for app: resource://gre-resources/hiddenWindow.html

Cannot open Settings DB. Trying to open an old version?
 -*- Shell.js: patch RemoteDebugger.start
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
SettingsService failed to open DB!
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
Cannot open Settings DB. Trying to open an old version?
```

ochameau:
Oh! I was convinced this breakage was due to some dark magic around addons.
But with your comments I figured out various things.

In first place, I managed to get these update popups while always using custom b2g-18 builds.
**But** sometime, I'm running some experiments of simulator running on top of Firefox.
I think that's the scenario where I manage to see these popups. And it will most likely break when going back to b2g-18.

Otherwise, there is still some scenarios where profile gets corrupted while always using b2g-18 builds.
It fails when you go back and forth in b2g 18 versions. It fails if I go from last version to b2g-18.0.2013-03-26.en-US.win32. But if you do a regular upgrade from old version to the newest, it works fine.

Given that we use a brand new profile when we update the addon, I think we can safely ignore these issues.

PR #583, opened 2013-06-04
Fix touch events on <select>

ochameau:
I tried hard to make gaia build system configurable enough for the simulator, but it looks like it is still not enough.

So. I removed `BROWSER=1` when running gaia's make, as we really don't want this line to operate: https://github.com/mozilla-b2g/gaia/blob/master/build/webapp-manifests.js#L94-L96
(That's a very hacky way to make permissions work on Firefox)
It introduced issues with system app in the simulator, so I removed BROWSER=1 in this changeset: https://github.com/mozilla/r2d2b2g/commit/4e8e82447ead7778fcaefad7ef767e1664bdf121

But... we really want following prefs to be set:
https://github.com/mozilla-b2g/gaia/blob/master/build/preferences.js#L49-L76
Here, for touch events, we are missing the `dom.w3c_touch_events.enabled=1` pref.

I don't really know yet how to fix that in gaia, but we can easily fix that on simulator side for v4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,7 @@ profile:
 	cp build/override-prefs.js gaia/build/custom-prefs.js
 	cp build/override-settings.json gaia/build/custom-settings.json
 	NOFTU=1 GAIA_APP_SRCDIRS=apps make -C gaia
+	BROWSER=1 NOFTU=1 GAIA_APP_SRCDIRS=apps make -C gaia preferences
 	python build/override-webapps.py
 	cd gaia/tools/extensions/desktop-helper/ && zip -r ../../../profile/extensions/desktop-helper\@gaiamobile.org.xpi *
 	cd gaia/tools/extensions/activities/ && zip -r ../../../profile/extensions/activities\@gaiamobile.org.xpi *


ochameau:
It fixes issues for the <select> in the settings app > date & time > region.
It may also fix other issues with touches being ignored.

mykmelez:
> So. I removed BROWSER=1 when running gaia's make, as we really don't want this line to operate: https://github.com/mozilla-b2g/gaia/blob/master/build/webapp-manifests.js#L94-L96
> (That's a very hacky way to make permissions work on Firefox)
> It introduced issues with system app in the simulator, so I removed BROWSER=1 in this changeset: 4e8e824

What issues did it introduce with the system app in the simulator, and how are we going to address them?

> But... we really want following prefs to be set:
> https://github.com/mozilla-b2g/gaia/blob/master/build/preferences.js#L49-L76
> Here, for touch events, we are missing the dom.w3c_touch_events.enabled=1 pref.

Can we instead set those prefs in override-prefs.js? Although that would duplicate code, it would avoid reintroducing those issues with the system app.

ochameau:
> What issues did it introduce with the system app in the simulator, and how are we going to address them?

Setting BROWSER=1 for webapps-manifests gaia's makefile rule introduce issues reported in PR #555 (rotation issue). This patch doesn't reintroduce it as I take care of setting BROWSER=1 only for preferences rule.
Having said that, I'm not really against doing something else.

mykmelez:
> Setting BROWSER=1 for webapps-manifests gaia's makefile rule introduce issues reported in PR #555 (rotation issue). This patch doesn't reintroduce it as I take care of setting BROWSER=1 only for preferences rule.

Ah, ok, I see.

> Having said that, I'm not really against doing something else.

The only other thing I can think of is to copy those prefs into the Simulator's own prefs file. Whether or not that makes sense depends on how many of those prefs are actually needed/useful in the Simulator, since each change has the potential to regress behavior.

So if only dom.w3c_touch_events.enabled and a couple other prefs are needed/useful, then I'd rather copy them. But if most of the prefs are needed/useful, then let's reuse them via BROWSER=1.

Unfortunately, I don't know how to determine that. So let's do this: we'll merge this change, which does clearly fix the problem in my testing, and then be vigilant over the next week for regressions and other problematic behavior. Then, if we see too many problems, we'll revert the change in favor of copying over only specific prefs.

PR #584, opened 2013-06-04
Reuse running adb instead of killing existing one. Fixes #464

### LINKED ISSUE #464: Make bundled adb play nicely with running system ADB
When I start up the simulator, I run into issues where the device isn't seen, or my resident adb will no longer see connected devices. I believe they are fighting for the same ports/residency on my system. We should investigate how to either
- use a running ADB daemon, or
- possibly start on a different port (I don't know whether this is possible)

bkase:
This patch reuses a running `adb` process if one exists. (fixes #464)

Tested on Arch Linux, Ubuntu, Windows, and OSX. To test, I ran the simulator add-on with `adb logcat` running in another terminal. And again without `adb logcat`. When the module unloads and `adb` was running on startup, `adb` will be killed. When the module unloads and `adb` was not running on startup, `adb` won't be killed.

### DIFF
--- a/addon/lib/adb.js
+++ b/addon/lib/adb.js
@@ -235,7 +235,7 @@ this.ADB = {
       args = [];
     } else {
       args = ["aux"];
-      let psCommand = "psi";
+      let psCommand = "ps";
 
       let paths = env.PATH.split(':');
       let len = paths.length;
--- a/addon/lib/adb.js
+++ b/addon/lib/adb.js
@@ -145,11 +145,12 @@ this.ADB = {
   // We startup by launching adb in server mode, and setting
   // the tcp socket preference to |true|
   start: function adb_start() {
-    let onSuccessfulStart = function onSuccessfulStart() {
+    let onSuccessfulStart = (function onSuccessfulStart() {
       Services.prefs.setBoolPref("dom.mozTCPSocket.enabled", true);
       Services.obs.notifyObservers(null, "adb-ready", null);
-      self.ready = true;
-    };
+      this.ready = true;
+    }).bind(this);
+
     this._isAdbRunning().then(
       (function onSuccess(isAdbRunning) {
         if (isAdbRunning) {
--- a/addon/lib/adb.js
+++ b/addon/lib/adb.js
@@ -11,7 +11,7 @@
 // then it's a JavaScript Module.
 const COMMONJS = ("require" in this);
 
-let components;
+let components, subprocess, file, env;
 if (COMMONJS) {
   components = require("chrome").components;
 } else {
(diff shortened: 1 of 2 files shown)


addon/lib/adb.js
```diff
@@ -21,6 +21,30 @@ let Cc = components.classes;
...
+  let loader = Loader({
+    paths: {
+      '': 'resource://gre/modules/commonjs/sdk/'
+    },
+    globals: { },
+    modules: { }
+  });
+
+  // This variable needs to be named something other than require or else
+  // the addon-sdk loader gets confused
```
mykmelez: Can you elaborate about how it gets confused? It'd be useful to have additional information here about precisely what the issue is.

addon/lib/simulator.js
```diff
@@ -69,7 +69,9 @@ let simulator = module.exports = {
     // and so they close their filehandles if the user is updating the addon,
     // which we need to do on Windows to replace the files.
     this.kill();
-    ADB.kill(Runtime.OS == "WINNT" ? true : false /* sync */);
+    if (ADB.didRunInitially) {
+      ADB.kill(Runtime.OS == "WINNT" ? true : false /* sync */);
```
mykmelez, line 73: Calling _ADB.kill_ should set _didRunInitially_ back to `false`.

mykmelez:
On my Mac, opening the Dashboard with these changes applied prints the following messages:

> info: r2d2b2g: adb: Error: a task list executable not found on filesystem
> info: r2d2b2g: adb: Didn't find ADB process running, restarting

Whereas the ADB module should find and use the _ps_ executable at `/bin/ps`.

Moreover, "push to device" fails. Pressing the _Push_ button for an app prints:

> info: r2d2b2g: Simulator.onMessage pushAppToDevice
> info: r2d2b2g: Simulator.pushAppToDevice: /Users/myk/Mozilla/GeoLocTest/manifest.webapp
> info: r2d2b2g: adb: forwardPort 6000
> info: r2d2b2g: adb: runCommand host:forward:tcp:6000;tcp:6000

But the device never asks me to confirm an incoming debugging connection, and the app never gets pushed.

bkase:
I accidentally kept a typo in the "ps" file when I sent out the pull-request, that should be fixed now.

I'm looking into the push to device problem

mykmelez:
I just tested on Windows 7, and I see the "push to device" problem there too.

bkase:
I never tested the push-to-device functionality, I assumed that if the dashboard said the device was connected, everything would work. Sorry about that.
It definitely works on Mac now, I'll check on the other OSs quickly.

EDIT: Works for me on OSX, Ubuntu, and Arch Linux, but I don't have adb drivers working for Windows yet

mykmelez:
Looks good, thanks for the fixes!

PR #585, opened 2013-06-05
UI Revamp!

potch:
Cleaned up some UI - logo, prettied-up "Add Project", and some other UI tweaks.


mykmelez:
Nit: per [copy rules](http://www.mozilla.org/en-US/styleguide/communications/copy-rules/), the official abbreviation of "Firefox OS" is FxOS, so ffos-wordmark.png should be called fxos-wordmark.png.

addon/data/content/css/main.css
```diff
@@ -1,288 +1,9 @@
-/*
```
mykmelez, line 1: Now that this file contains only our own code, it could use a license header!

```
/* This Source Code Form is subject to the terms of the Mozilla Public
```

addon/lib/main.js
```diff
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
```
mykmelez: Nit: even better, apply the standard format of this header, per [Mozilla License Headers](http://www.mozilla.org/MPL/headers/), which will make it easier to update it automatically (in case we ever have to do that again):

```
/* This Source Code Form is subject to the terms of the Mozilla Public
```

mykmelez:
Woot!

mykmelez:
Erm, we forgot to change the name of the image in the CSS reference, so I made that change in 0a334060150bf2ab41fab1f5dbb1319dc7cbdbb0 as part of pushing new preview builds.

PR #586, opened 2013-06-06
Add UI for changing receipt types

### LINKED ISSUE #494: Add UI to dashboard to specify receipt type.

potch:
Puts a receipt type `<select>` on each app listing. Changing the type triggers the appropriate app update.

fixes #494.

### DIFF
--- a/addon/data/content/js/applist.js
+++ b/addon/data/content/js/applist.js
@@ -56,7 +56,7 @@ var AppList = (function() {
             app.prettyLastUpdate = timedelta(app.lastUpdate);
         }
         app.prettyType = Simulator.APP_TYPES[app.type];
-        
+
         // use a default icon
         var iconPath = "default.png";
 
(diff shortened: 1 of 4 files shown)


addon/data/content/js/applist.js
```diff
@@ -74,12 +74,24 @@ var AppList = (function() {
         }
 
         app.iconPath = iconPath;
+        app.receiptTypes = ['none', 'ok', 'expired', 'invalid', 'refunded'];
```
nickdesaulniers: while 'expired' is a valid option, the receipt verifier service treats them as valid, which is unexpected.  I forgot to mention this earlier, I should have.  But to avoid confusion, can you remove it from the list of possible types?

nickdesaulniers:
Works well, tests are green.  Fix up that one nit and it's good to go!  Thanks @potch !

nickdesaulniers:
Good work @potch !

PR #591, opened 2013-06-09
Clean up how the device is shown when connected to the simulator

JeremiePat:
Currently, when you connect a device to the simulator, the icon on the sidebar is misplaced. This pull request intent fix this.

### DIFF
--- a/addon/data/content/css/main.css
+++ b/addon/data/content/css/main.css
@@ -509,13 +509,27 @@ section header {
 
 /* Device-related UI */
 .device-dependent {
-    transition: .5s opacity;
+    transition: .5s opacity, .5s height;
     opacity: 0;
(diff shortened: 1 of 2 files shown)


mykmelez:
I think this is a regression from a recent UX revamp by @potch that appears to have moved the Help link back to the sidebar. That still seems like the wrong place for it, which is why I moved it to the top-right corner of the window. But it isn't clear that there's space for it there in the new design, so I'll do some thinking about where else to put it.

And this is a worthy change in any case. Thanks, @JeremiePat!

mykmelez:
Pushed with additional commit cb7fdcfc3663b6dbeb54a3da302bcec07772dcc1 to add @JeremiePat to contributor list.

PR #592, opened 2013-06-10
Introduce devtool toolbox per app.

ochameau:
So here is the work I was mentioning in PR #567 https://github.com/mozilla/r2d2b2g/pull/567#issuecomment-19115271

This is a more substantial patch, but with a lot more value.
It introduces a `connect` button next to each app that is displayed/hidden whenever the app is opened/closed. Once you click on this button, a remote devtool toolbox is being displayed as a regular web one, in the same window as a "bottom bar".
Then, if you close the simulation or close the app the toolbox is automatically closed.
If you click on another app `connect` button, a new one connected to the related app replace the eventual current toolbox.

In order to do that, instead of hacking `dbg-browser-actor.js` I decided to add new requests in `dbg-webapps-actor.js` in order to stop hacking around tabs and finally provide access to apps through dedicated `apps` keywords (So I copy pasted a piece of dbg-browser-actor, mainly `AppActor`)
Also, I tried to come up with modifications that would be easily upstreamed to firefox codebase. Moving apps stuff out of `listTabs` request is particularely important regarding firefox as we will most likely not always care about apps, so that building apps actors on each remote connection startup would be in most cases unecessary (listTabs is requested on each connection startup).

So, the webapps actors gained 3 methods:
- `listApps`, that returns a dictionary of all current opened app actors, indexed by they manifest URL.
- `watchApps`, order the webapps actor to start watching for app opening/closing. The actor is going to dispatch `webappsOpen`/`webappsClose` events.
- `unwatchApps, order the actor to stop listening for webapps opening/closing.

Finally, there is this challenging piece of code that aims to open the toolbox on all 4 current Firefox versions. It includes some hacks in order to make work the toolbox as a "bottom bar" and a bunch of conditional code in order to match every subtle differences around target and toolbox we have over firefox releases.

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -32,8 +32,8 @@ const Geolocation = Cc["@mozilla.org/geolocation;1"].getService(Ci.nsISupports);
 dbgClient.UnsolicitedNotifications.geolocationStart = "geolocationStart";
 dbgClient.UnsolicitedNotifications.geolocationStop = "geolocationStop";
 dbgClient.UnsolicitedNotifications.appUpdateRequest = "appUpdateRequest";
-dbgClient.UnsolicitedNotifications.webappsOpen = "webappsOpen";
-dbgClient.UnsolicitedNotifications.webappsClose = "webappsClose";
+dbgClient.UnsolicitedNotifications.appOpen = "appOpen";
+dbgClient.UnsolicitedNotifications.appClose = "appClose";
 
 // Log subprocess error and debug messages to the console.  This logs messages
 // for all consumers of the API.  We trim the messages because they sometimes
@@ -262,8 +262,8 @@ const RemoteSimulatorClient = Class({
 
     client.addListener("geolocationStart", this.onGeolocationStart.bind(this));
     client.addListener("geolocationStop", this.onGeolocationStop.bind(this));
-    client.addListener("webappsOpen", this.onWebappsOpen.bind(this));
-    client.addListener("webappsClose", this.onWebappsClose.bind(this));
+    client.addListener("appOpen", this.onAppOpen.bind(this));
+    client.addListener("appClose", this.onAppClose.bind(this));
 
     this._registerAppUpdateRequest(client);
 
@@ -317,14 +317,14 @@ const RemoteSimulatorClient = Class({
     }
   },
 
-  onWebappsOpen: function onWebappsOpen(type, packet) {
-    emit(this, "webappsOpen", {
+  onAppOpen: function onAppOpen(type, packet) {
+    emit(this, "appOpen", {
       manifestURL: packet.manifestURL
     });
   },
 
-  onWebappsClose: function onWebappsClose(type, packet) {
-    emit(this, "webappsClose", {
+  onAppClose: function onAppClose(type, packet) {
+    emit(this, "appClose", {
       manifestURL: packet.manifestURL
     });
   },
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -32,6 +32,8 @@ const Geolocation = Cc["@mozilla.org/geolocation;1"].getService(Ci.nsISupports);
 dbgClient.UnsolicitedNotifications.geolocationStart = "geolocationStart";
 dbgClient.UnsolicitedNotifications.geolocationStop = "geolocationStop";
 dbgClient.UnsolicitedNotifications.appUpdateRequest = "appUpdateRequest";
+dbgClient.UnsolicitedNotifications.webappsOpen = "webappsOpen";
+dbgClient.UnsolicitedNotifications.webappsClose = "webappsClose";
 
 // Log subprocess error and debug messages to the console.  This logs messages
 // for all consumers of the API.  We trim the messages because they sometimes
@@ -65,6 +67,16 @@ const RemoteSimulatorClient = Class({
   // (means that we can start using the client!)
   get isReady() !!this._remote,
 
+  get client() this._remote.client,
+
+  getActorForApp: function (manifestURL, callback) {
+    this.client.request({to: this._remote.webapps, type: "listApps"},
+      function (reply) {
+        let actor = reply.apps[manifestURL];
+        callback(actor);
+      });
+  },
+
   _hookInternalEvents: function () {
     // on clientConnected, register an handler to close current connection 
     // on kill and send a "listTabs" debug protocol request, finally
@@ -92,7 +104,11 @@ const RemoteSimulatorClient = Class({
     this.on("clientReady", function (remote) {
       console.debug("rsc.onClientReady");
       this._remote = remote;
-      emit(this, "ready", null);
+      // Start watching app open/close
+      this.client.request({to: this._remote.webapps, type: "watchApps"},
+        (function () {
+          emit(this, "ready", null);
+        }).bind(this));
     });
 
     // on clientClosed, untrack old remote target and emit 
@@ -246,6 +262,8 @@ const RemoteSimulatorClient = Class({
 
     client.addListener("geolocationStart", this.onGeolocationStart.bind(this));
     client.addListener("geolocationStop", this.onGeolocationStop.bind(this));
+    client.addListener("webappsOpen", this.onWebappsOpen.bind(this));
+    client.addListener("webappsClose", this.onWebappsClose.bind(this));
 
     this._registerAppUpdateRequest(client);
 
@@ -297,6 +315,18 @@ const RemoteSimulatorClient = Class({
       Geolocation.clearWatch(this._geolocationID);
       this._geolocationID = null;
     }
+  },
+
+  onWebappsOpen: function onWebappsOpen(type, packet) {
+    emit(this, "webappsOpen", {
+      manifestURL: packet.manifestURL
+    });
+  },
+
+  onWebappsClose: function onWebappsClose(type, packet) {
+    emit(this, "webappsClose", {
+      manifestURL: packet.manifestURL
+    });
   },
 
   // send a getBuildID request to the remote simulator actor
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -317,13 +317,13 @@ const RemoteSimulatorClient = Class({
     }
   },
 
-  onAppOpen: function onAppOpen(type, packet) {
+  onAppOpen: function (type, packet) {
     emit(this, "appOpen", {
       manifestURL: packet.manifestURL
     });
   },
 
-  onAppClose: function onAppClose(type, packet) {
+  onAppClose: function (type, packet) {
     emit(this, "appClose", {
       manifestURL: packet.manifestURL
     });
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -1212,9 +1212,9 @@ let simulator = module.exports = {
           if (gRunningApps.indexOf(app) != -1) {
             cb();
           } else {
-            simulator.remoteSimulator.on("webappsOpen", function listener({manifestURL}) {
+            simulator.remoteSimulator.on("appOpen", function listener({manifestURL}) {
               if (manifestURL == app.manifestURL) {
-                simulator.remoteSimulator.removeListener("webappsOpen", listener);
+                simulator.remoteSimulator.removeListener("appOpen", listener);
                 cb();
               }
             });
@@ -1287,7 +1287,7 @@ let simulator = module.exports = {
       }
     });
 
-    remoteSimulator.on("webappsOpen", (function ({ manifestURL }) {
+    remoteSimulator.on("appOpen", (function ({ manifestURL }) {
       let app = this._getAppByManifestURL(manifestURL);
 
       // Ignore apps not being tracked by the simulator
@@ -1298,7 +1298,7 @@ let simulator = module.exports = {
       gRunningApps.push(app);
     }).bind(this));
 
-    remoteSimulator.on("webappsClose", (function ({ manifestURL }) {
+    remoteSimulator.on("appClose", (function ({ manifestURL }) {
       let app = this._getAppByManifestURL(manifestURL);
 
       // Ignore apps not being tracked by the simulator
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -38,6 +38,7 @@ const TEST_RECEIPT_URL = "https://marketplace.firefox.com/api/v1/receipts/test/"
 
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
+const { gDevTools } = Cu.import("resource:///modules/devtools/gDevTools.jsm", {});
 
 // NOTE: detect if developer toolbox feature can be enabled
 const HAS_CONNECT_DEVTOOLS = xulapp.is("Firefox") &&
@@ -1030,8 +1031,9 @@ let simulator = module.exports = {
       },
       set: function (target, name, v) {
         // Prevent `this._transport = null;`
-        if (name == "_transport")
+        if (name == "_transport") {
           return false;
+        }
         target[name] = v;
       }
     });
@@ -1043,7 +1045,6 @@ let simulator = module.exports = {
         client: clientProxy,
         chrome: false
       };
-      let {gDevTools} = Cu.import("resource:///modules/devtools/gDevTools.jsm", {});
 
       // Devtools API changed at each 3 last FF version :'(
       // Either on how to load modules, or how to use the API.
@@ -1307,8 +1308,9 @@ let simulator = module.exports = {
       }
 
       let idx = gRunningApps.indexOf(app);
-      if (idx != -1)
+      if (idx != -1) {
         gRunningApps.splice(idx, 1);
+      }
 
       // Close the current toolbox if it targets the closed app
       if (gCurrentToolboxManifestURL == manifestURL) {
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -55,6 +55,7 @@ const MANIFEST_CONTENT_TYPE = "application/x-web-app-manifest+json";
 
 let worker, remoteSimulator;
 let deviceConnected, adbReady, debuggerReady;
+let gCurrentToolbox, gCurrentToolboxManifestURL;
 
 let simulator = module.exports = {
   QueryInterface: XPCOMUtils.generateQI([Ci.nsIObserver,
@@ -150,9 +151,11 @@ let simulator = module.exports = {
       console.log("Selected " + manifestFile);
 
       let apps = simulator.apps;
+      let xkey = UUID.uuid().toString().slice(1, -1);
       apps[manifestFile] = {
         type: "local",
-        xkey: UUID.uuid().toString().slice(1, -1),
+        xkey: xkey,
+        origin: "app://" + xkey,
         receipt: null,
         receiptType: receiptType,
       };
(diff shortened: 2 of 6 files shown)


addon/lib/simulator.js
```diff
@@ -148,9 +149,11 @@ let simulator = module.exports = {
       console.log("Selected " + manifestFile);
 
       let apps = simulator.apps;
+      let xkey = UUID.uuid().toString().slice(1, -1);
       apps[manifestFile] = {
         type: "local",
-        xkey: UUID.uuid().toString().slice(1, -1),
+        xkey: xkey,
+        origin: "app://" + xkey,
```
ochameau: I had to add this `origin` attribute over here in order to get a valid `manifestURL` later.

mykmelez:
I see the Connect button when I run the addon with a new temporary profile and add some apps to the Dashboard, but when I run it with an existing profile that already has some apps and then start the Simulator. the addon throws the following exception:

```
TypeError: app is null
resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js 1242
  File "resource://gre/modules/devtools/DevToolsUtils.js", line 61, in 
    return aHandler.apply(this, arguments);
  File "resource://gre/modules/devtools/server/transport.js", line 174, in 
    self.hooks.onPacket(parsed);
  File "resource://gre/modules/devtools/dbg-client.jsm", line 633, in DC_onPacket
    }, function (ex) {
  File "resource://gre/modules/commonjs/sdk/core/promise.js", line 155, in then
    result.then(resolve, reject);
  File "resource://gre/modules/commonjs/sdk/core/promise.js", line 155, in then
    result.then(resolve, reject);
  File "resource://gre/modules/commonjs/sdk/core/promise.js", line 45, in then
    return { then: function then(fulfill) { fulfill(value); } };
  File "resource://gre/modules/commonjs/sdk/core/promise.js", line 120, in resolve
    deferred.resolve(onFulfill ? onFulfill(value) : value);
  File "resource://gre/modules/commonjs/sdk/core/promise.js", line 120, in resolve
    deferred.resolve(onFulfill ? onFulfill(value) : value);
  File "resource://gre/modules/devtools/dbg-client.jsm", line 625, in 
    this.notify(aPacket.type, aPacket);
  File "resource://gre/modules/devtools/dbg-client.jsm", line 151, in EV_notify
    listener.apply(null, arguments);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/remote-simulator-client.js", line 322, in onWebappsOpen
    manifestURL: packet.manifestURL
  File "resource://gre/modules/commonjs/sdk/event/core.js", line 85, in emit
  File "resource://gre/modules/commonjs/sdk/event/core.js", line 112, in lazy
    if (~state.indexOf(listener)) yield listener.apply(target, args);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js", line 1242, in module.exports.remoteSimulator/<
    app.opened = true;
```

mykmelez:
For both a new profile and an existing one, starting the Simulator still displays the Connect… button for the Simulator as a whole, and pressing the button throws:

```
TypeError: simulator.openConnectDevtools is not a function
resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js 1298
    callback.apply(null, args);
  File "resource://gre/modules/commonjs/sdk/content/worker.js", line 102, in onContentEvent/<
    self._emit.apply(self, JSON.parse(args));
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 123, in _emit
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 153, in _emitOnObject
  File "resource://gre/modules/commonjs/sdk/content/worker.js", line 227, in postMessage
    self._addonWorker._emit('message', data);
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 123, in _emit
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 153, in _emitOnObject
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js", line 1298, in onMessage
    simulator.openConnectDevtools();
```

mykmelez:
> TypeError: app is null

Some debugging reveals that the manifestURL of the app in question is app://homescreen.gaiamobile.org/manifest.webapp. So it looks like webappsOpen is being sent when the homescreen is opened, and the homescreen is not in the addon's list of apps, but the webappsOpen handler assumes all manifest URLs it receives are in that list.

ochameau:
I pushed a new changeset to fix the second exception.
Unfortunately, it will be a bit harder to fix the first one.
I think that we are facing the corrupted manifestURL attribute introduced by the bug fixed over here: https://github.com/mozilla/r2d2b2g/pull/592#discussion_r4618113
We are having app's manifestURL looking like this: `undefined/manifest.webapp` for packaged apps,
and as we save apps dictionary, we also save this corrupted attribute.
Did you already faced such issue? Like introducing a new attribute for app objects?
We could eventually hook a cleanup code in an initialization method to ensure having valid manifestURL for existing apps??

mykmelez:
> We are having app's manifestURL looking like this: undefined/manifest.webapp for packaged apps,
> and as we save apps dictionary, we also save this corrupted attribute.
> Did you already faced such issue? Like introducing a new attribute for app objects?
> We could eventually hook a cleanup code in an initialization method to ensure having valid manifestURL for existing apps??

We do have some code like that already:

https://github.com/mozilla/r2d2b2g/blob/8cbd07b9d91b4218338ebeda5e263e7fdec74a31/addon/lib/main.js#L85-L91

ochameau:
Oh so it looks like we do want to re-execute this `ensureXkeysValid` again to fix the manifestURL key.

mykmelez:
Starting the Simulator still shows a Connect… button underneath the toggle switch, and pressing it still throws _TypeError: simulator.openConnectDevtools is not a function_.

mykmelez:
I also don't see any app-specific Connect buttons when I test this against my standard existing profile, although I do see them when I test it with a new profile.

ochameau:
Oh right, I removed it based on our previous conversations. But I can also keep it somehow alive for experts that want to connect to the shell.xul document.

Otherwise, the cleanup method wasn't called if you tried installing it on top of a 4.0pre5+ version as my branch was outdated. It should work fine now that I rebased the branch against master and updated the version check to 4.0pre7dev!

mykmelez:
> Oh right, I removed it based on our previous conversations. But I can also keep it somehow alive for experts that want to connect to the shell.xul document.

Actually I'd prefer that we remove it, as you've done. Simulator developers can always obtain the port from a log message like the following and use it to connect manually:

> debug: r2d2b2g: subprocess: subprocess started; got PID 15409

mykmelez:
> Otherwise, the cleanup method wasn't called if you tried installing it on top of a 4.0pre5+ version as my branch was outdated. It should work fine now that I rebased the branch against master and updated the version check to 4.0pre7dev!

I had previously tested with an earlier version on master. But note that `cfx run` (and by extension `make run`) never exercise this code, since it only runs on _loadReason_ `install`, `upgrade`, and `downgrade`; whereas `cfx run` sets _loadReason_ to `startup`.

So you need to test it by creating a XPI and then installing it in a profile with an earlier version of the addon. Alternately, modify the code to _ensureXkeysValid_ unconditionally for one run, since you only need a single run to fix invalid xkeys.

Nevertheless, even though my xkeys are all valid, I still don't see per-app _Connect_ buttons when I test these changes with my standard existing profile.

mykmelez:
I did more investigation and realized that I misunderstood the implementation. I thought Connect buttons would appear for all apps whenever the Simulator is running. But actually they appear for only those apps that are currently "open" in the Simulator.

I suspect that this will be confusing for other developers as well, especially since it isn't immediately obvious which apps are open in the Simulator (one has to open the Task Manager to determine this, which involves an unadvertised long-press on the Home button).

From a task-centric perspective, it doesn't really matter whether or not the app is open in the Simulator. Nor does it particularly matter if the Simulator is even running. When a developer intends to debug an app, their goal is to connect the developer tools to an open instance of the app in a running instance of the Simulator. And the addon should make that as easy and automatic as possible.

So I think the addon should always display a Connect button for each app. Then, when the user presses the button, the addon should do the following things:
1. If the Simulator isn't running, start the Simulator.
2. If the app isn't installed, install the app.
3. If the app isn't open, open the app.
4. If the app is in the background, bring it to the foreground.
5. If the toolbox isn't visible, display the toolbox. (This already happens.)
6. Hook up the toolbox to the app. (This already happens.)
7. Notify the user that the toolbox has been hooked up to the app (which isn't obvious when the toolbox is already visible, and it's hooked up to a different app, but there aren't any console messages for that app). I'm not sure what that notification should look like. Perhaps it should be a pop-up or toast notification. Or we could simply write a message to the Console tab, since changing the app always switches the user back to that tab.

In any case, wherever we display it, the message should be something like "The toolbox is now connected to [App Name]."

Finally, we should remove the Run button, which won't serve much of a purpose now that Connect runs an app (as does Update). But this last change can happen in a different pull request. (I've been intending to make it anyway, since it's easy enough to Run an app from within the Simulator itself.)

ochameau:
Right, your suggestions make a lot of sense. I think I went the overengineering way...

I tried to implement what you described in the last two revisions.
I ended up printing a message in the webconsole when we open a new toolbox.

gaia
```diff
@@ -1 +1 @@
-Subproject commit cdabd1e6e1fbf348ab21fe55e94ac7eed1c009a6
+Subproject commit f2e1981e5ac09b72e1e8ffcbc0f5f6c42cb92037
```
mykmelez: I don't see this revision in the [upstream Gaia repository](https://github.com/mozilla-b2g/gaia). Nor do I see it in your [public fork on GitHub](https://github.com/ochameau/gaia) (which hasn't been updated since last year). Is it in a private fork?

mykmelez:
If an app is not installed in the Simulator, and you press its Connect button, then nothing appears to happen, but the following messages appear in the log:

  info: r2d2b2g: Simulator.onMessage connectToApp
  info: r2d2b2g: simulator actor received a 'runApp' command:aa58c9d7-3bec-844f-9001-43c272eca956

But if you subsequently press the app's Update button, then the toolbox appears and says that it is now connected to the app.

Also, if you press the app's Connect button twice before pressing its Update button, then two toolboxes appear!

mykmelez:
Breakpoints don't seem to work in hosted apps on this branch. For example, if I install http://mykzilla.org/app/manifest.webapp, connect to it, set a breakpoint on line 13 of its app.js (just inside the _screamScreen_ function), and then press its "scream screen" button, then execution doesn't stop at the breakpoint.

Breakpoints do work in hosted apps on the master branch when I use its _Connect…_ button to connect to the shell.xul remote tab. And they work in packaged apps on this branch. So the problem seems specific to hosted apps on this branch.

addon/lib/remote-simulator-client.js
```diff
@@ -32,6 +32,8 @@ const Geolocation = Cc["@mozilla.org/geolocation;1"].getService(Ci.nsISupports);
 dbgClient.UnsolicitedNotifications.geolocationStart = "geolocationStart";
 dbgClient.UnsolicitedNotifications.geolocationStop = "geolocationStop";
 dbgClient.UnsolicitedNotifications.appUpdateRequest = "appUpdateRequest";
+dbgClient.UnsolicitedNotifications.webappsOpen = "webappsOpen";
+dbgClient.UnsolicitedNotifications.webappsClose = "webappsClose";
```
mykmelez: Nit: these notifications seem to refer to a single app, so I would use the singular "webapp" in their names, i.e. "webappOpen" and "webappClose".

mykmelez: (In fact I'd probably simplify further to "appOpen" and "appClose", since it looks like we use "app" in other unsolicited notification names and request types, like "appUpdateRequest", "listApps", and "watchApps".)

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -285,6 +291,135 @@ WebappsActor.prototype = {
...
+        this._watchActorPool.addActor(actor);
+
+        // XXX: workaround to be able to get the frame during appterminated evt
+        origin = event.detail.origin;
+        this._framesByOrigin[origin] = frame;
+
+        this.conn.send({ from: this.actorID,
+                         type: "webappsOpen",
+                         manifestURL: frame.getAttribute("mozapp"),
+                         actor: actor.grip()
```
mykmelez, line 400: Does _actor_ get used anywhere? _RemoteSimulatorClient_ doesn't seem to use it, and I don't see any other code that receives the `webappsOpen` notification.

addon/lib/simulator.js
```diff
@@ -934,28 +938,100 @@ let simulator = module.exports = {
...
+    }
+    // We need to workaround existing devtools TabTarget.destroy code,
+    // that tries to close the client when the related toolbox is closed
+    // whereas we want to keep the client alive for other usages!
+    // http://hg.mozilla.org/mozilla-central/annotate/cfcce7c5eb74/browser/devtools/framework/target.js#l427
+    // TODO: remove this workaround when bug 875104 reach release channel
+    let clientProxy = new Proxy(this.remoteSimulator.client, {
+      get: function (target, name) {
+        if (name == "close")
+          return function () {};
```
mykmelez: Nit: we typically follow the convention of unconditionally surrounding even one-line conditional blocks with braces.

addon/lib/simulator.js
```diff
@@ -934,28 +938,100 @@ let simulator = module.exports = {
...
+        return target[name];
+      }
+    });
+    let self = this;
```
mykmelez, line 1053: This doesn't actually get used. If a function does need it, however, then the new hotness is to bind the function to its `this` value.

addon/lib/simulator.js
```diff
@@ -934,28 +938,100 @@ let simulator = module.exports = {
...
+    this.remoteSimulator.getActorForApp(app.manifestURL, function (actor) {
+      let options = {
+        form: actor,
+        client: clientProxy,
+        chrome: false
+      };
+      let {gDevTools} = Cu.import("resource:///modules/devtools/gDevTools.jsm", {});
```
mykmelez: Nit: it would be more consistent with our module importation strategy generally to (lazily) import this at the top of the script.

addon/lib/simulator.js
```diff
@@ -934,28 +938,100 @@ let simulator = module.exports = {
...
+        }
+        let promise = devtools.TargetFactory.forRemoteTab(options).then(function (target) {
+          // We have to set tab as BottomHost expect a tab attribute on target whereas
+          // TabTarget ignores any tab being given as options attributes passed to forRemoteTab.
+          let browserWindow = Services.wm.getMostRecentWindow("navigator:browser");
+          Object.defineProperty(target, "tab", {value: browserWindow.gBrowser.selectedTab});
+
+          let promise = gDevTools.showToolbox(target, "webconsole", devtools.Toolbox.HostType.BOTTOM);
+          promise.then(toolboxDisplayed);
         });
```
mykmelez, line 1080: Nit: since this code never does anything with promises except immediately call their _then_ methods, it isn't necessary to assign them to a local variable, and it seems like the code would be easier to read if the function calls were chained consistently, i.e.:

```
        devtools.TargetFactory.forRemoteTab(options).then(function (target) {
          // We have to set tab as BottomHost expect a tab attribute on target whereas
          // TabTarget ignores any tab being given as options attributes passed to forRemoteTab.
          let browserWindow = Services.wm.getMostRecentWindow("navigator:browser");
          Object.defineProperty(target, "tab", {value: browserWindow.gBrowser.selectedTab});

          gDevTools.showToolbox(target, "webconsole", devtools.Toolbox.HostType.BOTTOM).
          then(toolboxDisplayed);
        });
```

addon/lib/simulator.js
```diff
@@ -1122,12 +1215,57 @@ let simulator = module.exports = {
...
+      let app = this._getAppByManifestURL(manifestURL);
+
+      // Ignore apps not being tracked by the simulator
+      if (!app) {
+        return;
+      }
+
+      let idx = gRunningApps.indexOf(app);
+      if (idx != -1)
+        gRunningApps.splice(idx, 1);
```
mykmelez, line 1324: Nit: add braces around single-line conditional blocks per convention previously mentioned.

addon/lib/simulator.js
```diff
@@ -1122,12 +1215,57 @@ let simulator = module.exports = {
...
+        // Close any still opened toolbox
+        if (gCurrentToolbox) {
+          gCurrentToolbox.destroy();
+          gCurrentToolbox = null;
+        }
       }
     });
 
+    remoteSimulator.on("webappsOpen", (function ({ manifestURL }) {
+      let app = this._getAppByManifestURL(manifestURL);
```
mykmelez, line 1304: Nit: this could be a one-liner:

```
      let app = this.apps.filter(function(a) manifestURL == a.manifestURL)[0];
```

addon/lib/simulator.js
```diff
@@ -1182,6 +1320,10 @@ let simulator = module.exports = {
           }
         });
         break;
+      case "connectToApp":
+        app = this.apps[message.id];
+        simulator.connectToApp(app);
+        break;
```
mykmelez, line 1402: It looks like you're assigned to the _app_ variable declared at the top of _onMessage_, which makes sense, since it's there. But it doesn't look like _app_ is actually being used by any other code in _onMessage_, although there's another variable named _app_ defined in a callback, which is confusing.

So it seems like it would be better to remove the _app_ declaration at the top of the function and just pass this.apps[message.id] here. Or pass _message.id_ to _connectToApp_ and let that function get the app from the ID, like many other _case_s do.

Also note that _onMessage_ is a method of _simulator_, so the code can call _this.connectToApp_.

addon/lib/simulator.js
```diff
@@ -934,28 +938,100 @@ let simulator = module.exports = {
...
-                         "  function() " +
-                         "    document.getElementById('port').value = " +
-                         "      '" + parseInt(port, 10) + "', " +
-                         "  true" +
-                         ");"
-        }).on('detach', function restoreOriginalRemotePort() {
-          // restore previous value (autosaved on submit by connect.xhtml)
-          Services.prefs.setIntPref("devtools.debugger.remote-port", originalPort);
+  connectToApp: function(app) {
+    simulator.runApp(app, simulator.openToolboxForApp.bind(simulator, app));
```
mykmelez, line 954: _connectToApp_ is a method of _simulator_, so its code can refer to the _simulator_ singleton as _this_.

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -285,6 +291,135 @@ WebappsActor.prototype = {
     }
 
     return { appId: appId, path: appDir.path }
+  },
+
+  _createAppActor: function wa__createAppActor(frame) {
```
mykmelez: Nit: @campd and @benvie said yesterday on IRC that SpiderMonkey's function name inference is good these days, and they no longer recommend naming function expressions (except when adding an event listener that the listener itself removes), so new function expressions that we add to the codebase should generally not have names.

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -285,6 +291,135 @@ WebappsActor.prototype = {
...
+    }).bind(this);
+
+    // Register the system app
+    let chromeWindow = Services.wm.getMostRecentWindow('navigator:browser');
+    let systemAppFrame = chromeWindow.shell.contentBrowser;
+    registerApp.call(this, systemAppFrame);
+
+    // Register apps hosted in the system app. (i.e. all regular apps)
+    let frames = systemAppFrame.contentDocument.querySelectorAll("iframe[mozapp]");
+    for(var i = 0; i < frames.length; i++) {
```
mykmelez: Nit: var -> let (and space between `for` and open parenthesis).

ochameau:
A quick status update here as I lost my irc server.
I spent the whole day on this and the issues with the debugger end up being quite challenging.
I identified a first issue where debugger wasn't tracking new document being loaded in the app frame.
That was causing the issue with hosted apps.

Although there is also something broken with breakpoints. Everything works fine the first time you open a toolbox.
But if you close it and open a new one for the same app, new breakpoint are going to work, but on the client, in Firefox debugger panel, nothing will be updated; so that you do not really notice that the breakpoint worked. If you take a look at your app you will see it is frozen and you can unfreeze it by clicking on the resume button.

That's the challenging issue. I enabled all logs, tried to print lots of runtime values but wasn't able to figured it out.
I would imagine that we miss some cleanups when we close the toolbox, most likely on server side.
One way to workaround that would be to spawn a new connection for each new toolbox, but it doesn't sound like a simple task.

prosthesis/content/dbg-webapps-actors.js
ochameau, line 400: No, it isn't used. We are only using manifestURL. But that actor will be handy if we make a smart client code.
For now, it list all apps and create actors for all apps. Instead we should watch appOpen and keep the actor, or save listApp result.
If we go the appOpen way, this actor will be usefull. That makes me wonder if we should create a getActor request and create actors only on explicit demand? As the usefullness of having the actor on listApp and/or appOpen really depend on the client implementation.

mykmelez:
This looks good and now works in my tests, with two exceptions:
1. Pressing the Connect button for an app that isn't already installed in the Simulator still doesn't successfully connect to the Simulator. It does successfully start the Simulator, but the toolbox never opens. I spent some time debugging this issue, and I developed a fix that I'll submit as a followup pull request.
2. If a toolbox is open on Firefox shutdown, a `this._client is null` exception sometimes gets thrown:

```
TypeError: this._client is null
resource://gre/modules/devtools/WebConsoleClient.jsm 324
  File "resource://r2d2b2g-at-mozilla-dot-org/subprocess/lib/subprocess.js", line 1561, in _done
    stderr: error,
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/remote-simulator-client.js", line 207, in 
    emit(this, "exit", result.exitCode);
  File "resource://gre/modules/commonjs/sdk/event/core.js", line 85, in emit
  File "resource://gre/modules/commonjs/sdk/event/core.js", line 112, in lazy
    if (~state.indexOf(listener)) yield listener.apply(target, args);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js", line 1280, in module.exports.remoteSimulator/remoteSimulator<.onExit
    gCurrentToolbox.destroy();
  File "resource:///modules/devtools/framework/toolbox.js", line 771, in TBOX_destroy
    outstanding.push(this._target.destroy());
  File "resource:///modules/devtools/framework/target.js", line 439, in TabTarget.prototype.destroy
    }.bind(this));
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js", line 1009, in clientClose
    continuation();
  File "resource://gre/modules/devtools/WebConsoleClient.jsm", line 355, in WCC_close
    this.stopListeners(null, aOnResponse);
  File "resource://gre/modules/devtools/WebConsoleClient.jsm", line 324, in WCC_stopListeners
    this._client.request(packet, aOnResponse);
info: r2d2b2g: adb: adb kill-server: 0
info: r2d2b2g: adb: trackDevices onclose
info: r2d2b2g: adb: adb kill-server: 0
Total time: 21.265524 seconds
Program terminated successfully.
```

I don't see that consistently, however, and it doesn't seem to cause any problem, so I don't think it should block this request.

Thus this is ready to merge, and I'll do so after reconciling conflicts with master.

PR #594, opened 2013-06-10
push receipt to device

nickdesaulniers:
review? @mykmelez
this is the patch the simulator needs to send receipt, first half of issue #496


mykmelez:
This certainly looks quite simple! Does it break "push to device" if the device in question doesn't yet support receiving receipts?

nickdesaulniers:
No sir, as [the receiving end just pulls out the appId](http://mxr.mozilla.org/mozilla-central/source/toolkit/devtools/server/actors/webapps.js#244).

mykmelez:
Then there's no reason to wait until the B2G change to land this one!

PR #602, opened 2013-06-11
FEATURE: clear sessionStorage, localStorage and cookie on "refreshClearCacheCmd"

### LINKED ISSUE #595: purge IndexedDB (or LocalStorage?) when user presses Shift-Cmd/Ctrl-R
Last Friday I demoed @rpl's Shift-Cmd/Ctrl-R feature to purge the AppCache while refreshing an app, and @cvan (or was it @mattbasta?) suggested we also purge another cache/database that I think was IndexedDB (although it could have been LocalStorage).

@cvan, @mattbasta, can y'all confirm which data you want to have purged during such a refresh?

rpl:
Should fix #595

Minimal test app:
- https://github.com/rpl/firefoxos-test-apps/tree/master/test-refreshClearCacheCmd-app

### DIFF
--- a/prosthesis/content/shell.js
+++ b/prosthesis/content/shell.js
@@ -100,11 +100,19 @@ function simulatorAppUpdate(clearCacheAndStorages) {
   if (clearCacheAndStorages) {
     debug("clear cache and storages for: " + origin);
 
-    let appWindow = simulatorGetAppWindow(origin);
+    let iframe = Array.prototype.filter.call(
+      shell.contentBrowser.contentDocument.getElementsByTagName("iframe"),
+      function(iframe) iframe.getAttribute("data-frame-origin") == origin
+    )[0];
+
+    if (!iframe) {
+      debug("ERROR: app iframe not found");
+      return;
+    }
 
     // clear localStorage/sessionStorage
-    appWindow.localStorage.clear();
-    appWindow.sessionStorage.clear();
+    iframe.contentWindow.localStorage.clear();
+    iframe.contentWindow.sessionStorage.clear();
 
     let localId = DOMApplicationRegistry.webapps[appId].localId;
 
@@ -133,14 +141,4 @@ function simulatorAppUpdate(clearCacheAndStorages) {
       appId: DOMApplicationRegistry._appId(origin)
     }
   }, "r2d2b2g:app-update", null);
-}
-
-function simulatorGetAppWindow(origin) {
-  // NOTE: some characters needs to be escaped when used in a query selector
-  // notes and code snippet from:
-  // http://www.mymindleaks.com/article/escape-special-characters-in-javascript-css-queryselector.html
-  let escapedOrigin = origin.replace(/[\[\]\.\:\*]/g,"\\$&");
-  let iframe = shell.contentBrowser.contentDocument.
-        querySelector("iframe[data-frame-origin='" + escapedOrigin + "']");
-  return iframe ? iframe.contentWindow : null;
 }
--- a/prosthesis/content/shell.js
+++ b/prosthesis/content/shell.js
@@ -105,7 +105,7 @@ function simulatorAppUpdate(clearCacheAndStorages) {
     )[0];
 
     if (!iframe) {
-      debug("ERROR: app iframe not found");
+      debug("error! app iframe not found");
       return;
     }
 
@@ -116,7 +116,7 @@ function simulatorAppUpdate(clearCacheAndStorages) {
     let localId = DOMApplicationRegistry.webapps[appId].localId;
 
     if (!localId) {
-      debug("ERROR: app localId not found");
+      debug("error! app localId not found");
       return;
     }
 
--- a/prosthesis/content/shell.js
+++ b/prosthesis/content/shell.js
@@ -98,7 +98,8 @@ function simulatorAppUpdate(clearAppCache) {
   let appId = DOMApplicationRegistry._appId(origin);
 
   if (clearAppCache) {
-    debug("clear all associated appCache entries: " + origin);
+    debug("clear cache and storages for: " + origin);
+    simulatorClearStorages(origin);
     simulatorClearAppCache(appId);
   }
 
@@ -111,7 +112,29 @@ function simulatorAppUpdate(clearAppCache) {
   }, "r2d2b2g:app-update", null);
 }
 
+function simulatorGetAppWindow(origin) {
+  let iframe = shell.contentBrowser.contentDocument
+        .querySelector("iframe[data-frame-origin='" + origin + "']");
+  return iframe ? iframe.contentWindow : null;
+}
+
+function simulatorClearStorages(origin) {
+  let appWindow = simulatorGetAppWindow(origin);
+
+  debug("clear sessionStorage entries");
+  appWindow.sessionStorage.clear();
+  debug("clear localStorage entries");
+  appWindow.localStorage.clear();
+  debug("clear cookies");
+  let s=(appWindow.document.cookie+';').
+        replace(/([^=]*)=[^;]*; */g,"$1=; expires=Thu, 01 Jan 1970 00:00:00 GMT;\n").split('\n');
+  for (var i in s) {
+    appWindow.document.cookie = s[i];
+  }
+}
+
 function simulatorClearAppCache(appId) {
+  debug("clear appCache entries");
   let localId = DOMApplicationRegistry.webapps[appId].localId;
 
   if (!localId) {
--- a/prosthesis/content/shell.js
+++ b/prosthesis/content/shell.js
@@ -91,62 +91,62 @@ document.getElementById("rotateButton").addEventListener("click", function() {
           .addEventListener("click", openWin);
 }
 
-function simulatorAppUpdate(clearAppCache) {
-  let wm = shell.contentBrowser.contentWindow.wrappedJSObject.
-           WindowManager;
-  let origin = wm.getCurrentDisplayedApp().origin;
-  let appId = DOMApplicationRegistry._appId(origin);
+let SimulatorAppUpdate = {
+  run: function appUpdateRun(clearCacheAndStorages) {
+    let wm = shell.contentBrowser.contentWindow.wrappedJSObject.
+          WindowManager;
+    let origin = wm.getCurrentDisplayedApp().origin;
+    let appId = DOMApplicationRegistry._appId(origin);
 
-  if (clearAppCache) {
-    debug("clear cache and storages for: " + origin);
-    let localId = DOMApplicationRegistry.webapps[appId].localId;
-    simulatorClearStorages(origin);
-    simulatorClearCookies(localId);
-    simulatorClearAppCache(localId);
-  }
-
-  debug("request app reinstall:" + origin);
-  Services.obs.notifyObservers({
-    wrappedJSObject: {
-      origin: origin,
-      appId: DOMApplicationRegistry._appId(origin)
+    if (clearCacheAndStorages) {
+      debug("clear cache and storages for: " + origin);
+      let localId = DOMApplicationRegistry.webapps[appId].localId;
+      this._doClearStorages(origin);
+      this._doClearCookies(localId);
+      this._doClearAppCache(localId);
     }
-  }, "r2d2b2g:app-update", null);
-}
 
-function simulatorGetAppWindow(origin) {
-  let iframe = shell.contentBrowser.contentDocument
-        .querySelector("iframe[data-frame-origin='" + origin + "']");
-  return iframe ? iframe.contentWindow : null;
-}
+    debug("request app reinstall:" + origin);
+    Services.obs.notifyObservers({
+      wrappedJSObject: {
+        origin: origin,
+        appId: DOMApplicationRegistry._appId(origin)
+      }
+    }, "r2d2b2g:app-update", null);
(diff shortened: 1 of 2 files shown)


prosthesis/content/shell.js
```diff
@@ -111,7 +112,29 @@ function simulatorAppUpdate(clearAppCache) {
...
+  debug("clear sessionStorage entries");
+  appWindow.sessionStorage.clear();
+  debug("clear localStorage entries");
+  appWindow.localStorage.clear();
+  debug("clear cookies");
+  let s=(appWindow.document.cookie+';').
+        replace(/([^=]*)=[^;]*; */g,"$1=; expires=Thu, 01 Jan 1970 00:00:00 GMT;\n").split('\n');
+  for (var i in s) {
+    appWindow.document.cookie = s[i];
+  }
```
ochameau: For the cookies, there is some dedicated platform method to cleanup cookies for a given app.
See this piece of code for example:
http://mxr.mozilla.org/mozilla-central/source/dom/apps/src/Webapps.jsm#396

rpl: much better!

prosthesis/content/shell.js
```diff
@@ -111,16 +114,32 @@ function simulatorAppUpdate(clearAppCache) {
...
+  let iframe = shell.contentBrowser.contentDocument
+        .querySelector("iframe[data-frame-origin='" + origin + "']");
+  return iframe ? iframe.contentWindow : null;
+}
 
-  if (!localId) {
-    return;
-  }
+function simulatorClearCookies(localAppId) {
+  debug("clear cookies");
```
mykmelez: Nit: the debug statements in each individual clearing function seem like overkill; the single such statement in simulatorAppUpdate should be sufficient to inform us that the data is being cleared.

prosthesis/content/shell.js
```diff
@@ -111,16 +114,32 @@ function simulatorAppUpdate(clearAppCache) {
   }, "r2d2b2g:app-update", null);
 }
 
-function simulatorClearAppCache(appId) {
-  let localId = DOMApplicationRegistry.webapps[appId].localId;
+function simulatorGetAppWindow(origin) {
+  let iframe = shell.contentBrowser.contentDocument
+        .querySelector("iframe[data-frame-origin='" + origin + "']");
```
mykmelez: Nit: a multi-line statement like this is usually formatted thusly:

```
  let iframe = shell.contentBrowser.contentDocument.
               querySelector("iframe[data-frame-origin='" + origin + "']");
```

Also, doesn't origin need to be escaped in some fashion when used in a querySelector call like this one?

prosthesis/content/shell.js
```diff
@@ -111,16 +114,32 @@ function simulatorAppUpdate(clearAppCache) {
   }, "r2d2b2g:app-update", null);
 }
 
-function simulatorClearAppCache(appId) {
-  let localId = DOMApplicationRegistry.webapps[appId].localId;
+function simulatorGetAppWindow(origin) {
```
mykmelez: Instead of prefixing each global symbol we add with "simulator", which will get unwieldy over time, we should create a Simulator singleton object that encapsulates the various Simulator-specific functions and values.

mykmelez:
One final note: the localStorage and sessionStorage clearing seems to work, but I can't quite get cookies to persist in your test app, so I haven't been able test clearing them.

prosthesis/content/shell.js
```diff
@@ -91,43 +91,62 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
-    debug("clear all associated appCache entries: " + origin);
-    simulatorClearAppCache(appId);
-  }
-
-  debug("request app reinstall:" + origin);
-  Services.obs.notifyObservers({
-    wrappedJSObject: {
-      origin: origin,
-      appId: DOMApplicationRegistry._appId(origin)
+let SimulatorAppUpdate = {
```
mykmelez: I'm not sure it's really necessary to turn this into an object and break up its code into multiple methods. It looks like none of the methods have multiple callers, and several of them are very short, including one that comprises only a single line of code; so this refactoring strikes me as [premature generalization](http://c2.com/cgi/wiki?PrematureGeneralization), which we should avoid!

prosthesis/content/shell.js
```diff
@@ -91,43 +91,62 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+  run: function appUpdateRun(clearCacheAndStorages) {
```
mykmelez: Nit: in general, when naming function expressions assigned to object properties, follow the convention of giving them the same name as the name of the property itself, i.e. call this function simply `run`. That makes it easy enough to identify the function in stack dumps while limiting the length and complexity of the function name.

mykmelez: I just double-checked the convention in #devtools, and @campd says that SpiderMonkey now has "pretty good name inference", so actually we should stop naming function expressions assigned to object properties.

prosthesis/content/shell.js
```diff
@@ -91,43 +91,62 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+    let wm = shell.contentBrowser.contentWindow.wrappedJSObject.
+          WindowManager;
```
mykmelez: Nit, align the property name on the second line with the object being dereferenced on the first line, i.e.:

```
    let wm = shell.contentBrowser.contentWindow.wrappedJSObject.
             WindowManager;
```

prosthesis/content/shell.js
```diff
@@ -91,43 +91,62 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+    let origin = wm.getCurrentDisplayedApp().origin;
```
mykmelez: Since this is the only use of _WindowManager_, assigning _WindowManager_ to a local variable is also premature generalization that we should generally avoid, since it makes the code more complicated to read and understand, and it doesn't reduce programming error.

We're better off waiting until we actually need to use a value twice before we factor it out into a local variable, even if it makes a subsequent change larger (because it contains the refactoring). Since we might never need to use the variable twice, and that way we only pay the cost of refactoring the variable when it becomes necessary. So this should be simply:

```
    let origin = shell.contentBrowser.contentWindow.wrappedJSObject.
                 WindowManager.getCurrentDisplayedApp().origin;
```

rpl:
@mykmelez I've just put together a simpler and readable single simulatorAppUpdate function

prosthesis/content/shell.js
```diff
@@ -91,43 +91,62 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+      // (NS_ERROR_UNEXPECTED) [nsIApplicationCacheService.discardByAppId]" \
+      // nsresult: "0x8000ffff (NS_ERROR_UNEXPECTED)" ...
+      // debug("EXCEPTION on localId " + localId + ": " + e);
+    }
+  },
+  _escapeForQuerySelector: function escapeNameForQuery(name) {
+    // NOTE: some characters needs to be escaped when used in a query selector
+    // notes and code snippet from:
+    // http://www.mymindleaks.com/article/escape-special-characters-in-javascript-css-queryselector.html
+    return typeof(name) == 'string' ? name.replace(/[\[\]\.\:\*]/g,"\\$&") : name;
```
mykmelez: That article notes that the characters it escapes are just a sample of the special characters, and there are actually many more characters to escape.

This [blog post](http://mathiasbynens.be/notes/css-escapes) describes them all, and this [unlicensed codebase](https://github.com/mathiasbynens/mothereff.in/tree/master/css-escapes) implements an escaper. But its code is quite large and complicated, so I think we're better off using a different API than _querySelector_ to retrieve the frame in question, even though _querySelector_ is an elegant way of doing so.

For example, we could use something like the following code:

```
    let iframe = Array.prototype.filter.call(
      shell.contentBrowser.contentDocument.getElementsByTagName("iframe"),
      function(iframe) iframe.getAttribute("data-frame-origin") == origin
    )[0];
```

rpl:
@mykmelez my fault, I've pushed a new commit to change it to the safer method and remove the get app window helper

prosthesis/content/shell.js
```diff
@@ -111,23 +135,12 @@ function simulatorAppUpdate(clearAppCache) {
...
-  try {
-    Cc["@mozilla.org/network/application-cache-service;1"].
-      getService(Ci.nsIApplicationCacheService).discardByAppId(localId, false);
-  } catch(e) {
-    // NOTE: currently cacheService.discardByAppId always raise an expection
-    // even if it's working correctly (and cache entries are refreshed on app
-    // window reload)
-    // EXCEPTION on localId NNNN: [Exception... "Component returned failure code: 0x8000ffff \
-    // (NS_ERROR_UNEXPECTED) [nsIApplicationCacheService.discardByAppId]" \
-    // nsresult: "0x8000ffff (NS_ERROR_UNEXPECTED)" ...
```
mykmelez, line 130: This code retrieves _appWindow_ and then calls some properties on it. But __getAppWindow_ doesn't assume that _appWindow_ exists, and it returns _null_ if it doesn't, which means the two code blocks disagree about whether or not _appWindow_ must exist. One way or another, they should agree, so either this code should check that _appWindow_ really exists, or the code in __getAppWindow_ shouldn't check its existence.

Leaving the code unrefactored, so the code that retrieves _appWindow_ and the code that uses it live right next to each other in the same block, would have avoided this discrepancy. Which is a good example of why it's useful to avoid premature  generalization! ;-)

mykmelez, line 130: Erm, this comment actually was meant to apply to an earlier version of your change! While I was writing the comment, you changed the code! But GitHub didn't tell me that, it just applied my comment to a different part of your code, for which the comment doesn't make any sense! So GitHub wasn't helpful in this case. :-/

mykmelez, line 130: In any case, I suspect that you probably already fixed the issue I identified! I'll take a look now and see…

rpl, line 130: Excuse me, I was not aware github updates pull request diffs "inplace" :-(

prosthesis/content/shell.js
```diff
@@ -91,15 +91,52 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
 }
 
-function simulatorAppUpdate(clearAppCache) {
-  let wm = shell.contentBrowser.contentWindow.wrappedJSObject.
-           WindowManager;
-  let origin = wm.getCurrentDisplayedApp().origin;
+function simulatorAppUpdate(clearCacheAndStorages) {
+  let origin = shell.contentBrowser.
+           contentWindow.wrappedJSObject.
+           WindowManager.getCurrentDisplayedApp().origin;
```
mykmelez: Nit: align the properties on subsequent lines with the reference on the first one, i.e.:

```
  let origin = shell.contentBrowser.
               contentWindow.wrappedJSObject.
               WindowManager.getCurrentDisplayedApp().origin;
```

Although in this case you have enough horizontal columns to shorten this to two lines:

```
  let origin = shell.contentBrowser.contentWindow.wrappedJSObject.
               WindowManager.getCurrentDisplayedApp().origin;
```

prosthesis/content/shell.js
```diff
@@ -91,15 +91,52 @@ document.getElementById("rotateButton").addEventListener("click", function() {
...
+  if (clearCacheAndStorages) {
+    debug("clear cache and storages for: " + origin);
+
+    let iframe = Array.prototype.filter.call(
+      shell.contentBrowser.contentDocument.getElementsByTagName("iframe"),
+      function(iframe) iframe.getAttribute("data-frame-origin") == origin
+    )[0];
+
+    if (!iframe) {
+      debug("ERROR: app iframe not found");
```
mykmelez: Nit: I would prefer to avoid having console messages that SHOUT (i.e. use all-caps), because they are distracting and make it difficult to read the log, especially when looking for other messages.

mykmelez:
> In any case, I suspect that you probably already fixed the issue I identified! I'll take a look now and see…

Yup, looks great! The only remaining issues are two minor nits.

> Excuse me, I was not aware github updates pull request diffs "inplace" :-( 

No problem, I didn't realize this either! I suppose it can be a useful feature, since it ensures you are looking at the latest version of the code. But GitHub should warn you when you're about to comment on a line of code that has changed out from under you!

prosthesis/content/shell.js
rpl: @mykmelez what about:

```
  debug("error! app iframe not found");
```

mykmelez:
>  debug("error! app iframe not found");

That's much better! It's still a little distracting, but not nearly as much, so it's fine.

(In the long term, we should report such messages as errors via something like _Components.utils.reportError_; and we might want to notify the user that something went wrong; but I haven't thought through the implications of this yet, so let's not worry about it now.)

mykmelez:
According to the [DOM Storage guide](https://developer.mozilla.org/en-US/docs/Web/Guide/DOM/Storage), sessionStorage isn't supposed to persist across browser restarts. And when I quit an app via FxOS's Task Manager and then restart it, sessionStorage has been cleared. So I would expect it to be cleared when I reinstall an app also. But that doesn't appear to happen, strangely.

In any case, that means it's useful for this code to clear sessionStorage, even if it seems like it shouldn't be necessary. And the code looks good and works as expected. Thanks for your patience through the many changes!

PR #615, opened 2013-06-13
Added rules to .gitignore such that repo is clean

bkase:
With this fix, there shouldn't be any untracked files after a `make` or `make package`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,12 @@ addon/data/win32/adb
 
 # locale repositories cloned by build/make-locales.py
 gaia-l10n
+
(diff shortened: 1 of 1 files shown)


mykmelez:
This is a common pull request! But we haven't yet accepted such changes, because we've been following a best practice to put such rules into a global .gitignore file, as suggested by GitHub's [Ignoring files](https://help.github.com/articles/ignoring-files#global-gitignore).

(The number of times it comes up makes me wonder… but let's stick to the current strategy for the time being; and perhaps add a note to _CONTRIBUTING.md_.)

PR #616, opened 2013-06-13
change Receipt Type label to Receipt

mykmelez:
The "Receipt Type" label is accurate, since it is the type of receipt that is being selected, not the receipt itself. But it isn't necessary to be this accurate, since users can grok the incongruence between the word "receipt" and the types listed in the menu; and it takes up extra space to include the word "type" in a space-constrained UI. So we should label this menu simply "Receipt", which is accurate enough and takes up less space.


mykmelez:
@nickdesaulniers does this seem reasonable?

PR #626, opened 2013-06-14
remove Run button fixes #619

### LINKED ISSUE #619: remove Run button
The Run button has limited utility, since Update runs an app, and a new Connect button will do so as well after #592. Plus it's easy enough to run an app from within the Simulator itself. So we should remove the Run button to simplify the array of choices we offer the user.

nickdesaulniers:


nickdesaulniers:
review? @potch

PR #627, opened 2013-06-14
Additional UI cleanup

potch:


potch:
@nickdesaulniers , r?

Issue #617, opened 2013-06-13
Manifest.webapp is not refreshing


breck7:
Here's my workflow:
- Open simulator
- Install app from url to the manifest.webapp file
- App installs great in simulator
- Update manifest.webapp
- Click "Update" in Dashboard

Expected: new manifest file is reflected.
Actual: cached(outdated) manifest file is used.

If I open a new tab in FF and redownload the manifest file, then refresh dashboard and click "Update", I get the new manifest file.

Perhaps this is already fixed, and it's something wrong with my setup.

mykmelez:
Is there a particular field or fields that you're changing? Any chance your manifest file is publicly available?

breck7:
I was changing the App title field and the icons field.

breck7:
Manifest file is here: http://firefly.nudgepad.com/manifest.webapp

mykmelez:
Hmm, I see this too on the latest dev version of the Simulator when changing the `name` field of a hosted app (http://mykzilla.org/app/manifest.webapp). And I confirm the workaround of downloading the manifest via a separate tab. Which suggests that this is a cache issue: updating the app uses a local cached copy of the manifest file instead of a fresh copy downloaded from its server.

breck7:
Yes, that was my guess as well. It seems like hitting update should always download the latest manfiest.webapp file

mykmelez:
> Yes, that was my guess as well. It seems like hitting update should always download the latest manfiest.webapp file

I agree!

breck7:
@mykmelez Much better code than my fiddling with the url.

Just updated. Using the new github interface so not quite sure if I submitted that updated request correctly.

mykmelez:
I'm on the new GitHub interface now too, and it looks like you submitted it correctly, so I have merged it along with https://github.com/mozilla/r2d2b2g/commit/b7027dce1d13a47041bffe99c6273e82b9b70fa8 which adds you to the list of contributors. Thanks for the issue and the fix!

PR #628, opened 2013-06-17
Always fetch latest manfiest.webapp when you click "Update"  on a Hosted App.

breck7:
Avoid hitting cache on issue #617.

There could be a better way to do avoid the cache hit. This is just the simplest I could think of.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -771,7 +771,7 @@ let simulator = module.exports = {
       break;
     case "hosted":
       Request({
-        url: id,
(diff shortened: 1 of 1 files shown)


mykmelez:
@breck7 You don't need to close this request and open a new one, you can just push another commit to the branch for which this requests a pull! I'm happy to merge it even if it contains changes that we later undo in favor of different solutions. (Alternately, if you prefer, you can squash the branch; for which you can still reuse the existing request.)

breck7:
Whoops! Okay won't happen again. Thanks for the tips

PR #630, opened 2013-06-17
notify user when push to device fails Fixes #593

### LINKED ISSUE #593: notify user if "push to device" fails
If a "push to device" fails for any reason, the Dashboard should tell the user about it, so they know that it failed (and ideally also why it failed). See issue #538 for examples of failures that we don't tell the user about.

nickdesaulniers:
review? @mykmelez

### DIFF
--- a/addon/lib/debugger.js
+++ b/addon/lib/debugger.js
@@ -8,6 +8,8 @@
 // (from an addon built using the Add-on SDK).  If it isn't a CommonJS Module,
 // then it's a JavaScript Module.
 const COMMONJS = ("require" in this);
+const { setTimeout, clearTimeout } = require("sdk/timers");
+const TIMEOUT_DURATION = 15000; // ms
 
 let components;
 if (COMMONJS) {
@@ -45,9 +47,16 @@ this.Debugger = {
     client = new DebuggerClient(transport);
 
     let deferred = Promise.defer();
-    let self = this;
 
+    let connectionTimer = setTimeout(function timedOut() {
+      let errorMsg = "Failed to connect to device: timeout";
+      dump(errorMsg + "\n");
+      deferred.reject(errorMsg);
+    }, TIMEOUT_DURATION);
+
+    // Not guaranteed to connect
     client.connect(function onConnected(aType, aTraits) {
+      clearTimeout(connectionTimer);
       client.listTabs(function(aResponse) {
         if (aResponse.webappsActor) {
           webappsActor = aResponse.webappsActor;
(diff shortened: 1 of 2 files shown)


addon/lib/debugger.js
```diff
@@ -45,9 +47,16 @@ this.Debugger = {
     client = new DebuggerClient(transport);
 
     let deferred = Promise.defer();
-    let self = this;
 
+    let connectionTimer = setTimeout(function timedOut() {
```
mykmelez: Nit: @campd and @Benvie pointed out to me today in #devtools that SpiderMonkey has good name inference these days, even for callback functions (although I'm not sure what it would call this one), so it isn't necessary to name function expressions anymore, and they generally recommend against doing so (except when adding a listener that must remove itself). So I would make this an anonymous function.

addon/lib/debugger.js
```diff
@@ -45,9 +47,16 @@ this.Debugger = {
...
+      let errorMsg = "Failed to connect to device: timeout";
+      dump(errorMsg + "\n");
+      deferred.reject(errorMsg);
+    }, TIMEOUT_DURATION);
+
+    // Not guaranteed to connect
```
mykmelez: Nit: this should cite the bug in question so a future reader of this code can find out why it wasn't guaranteed (and if that's been fixed since then).

nickdesaulniers:
review? @mykmelez

addon/lib/debugger.js
```diff
@@ -8,6 +8,8 @@
 // (from an addon built using the Add-on SDK).  If it isn't a CommonJS Module,
 // then it's a JavaScript Module.
 const COMMONJS = ("require" in this);
+const { setTimeout, clearTimeout } = require("sdk/timers");
```
mykmelez: debugger.js is shared with [b2gremote](https://github.com/fabricedesre/b2gremote), where it's loaded as a JSM; and we need to keep it working there. So use nsITimer to create/destroy the timer if !COMMONJS.

addon/lib/debugger.js
```diff
@@ -8,12 +8,32 @@
 // (from an addon built using the Add-on SDK).  If it isn't a CommonJS Module,
 // then it's a JavaScript Module.
 const COMMONJS = ("require" in this);
+const TIMEOUT_DURATION = 15000; // ms
 
 let components;
 if (COMMONJS) {
   components = require("chrome").components;
+  var { setTimeout, clearTimeout } = require("sdk/timers");
```
mykmelez: Instead of declaring these with _var_ here in the _if_ block and also in the _else_ block, declare them with the _let components_ declaration outside of the conditional blocks, then only assign them in the conditional blocks.

mykmelez:
Looks great, thanks @nickdesaulniers!

PR #632, opened 2013-06-18
Fix local app icon path on Windows. Fixes #548.

### LINKED ISSUE #548: Icon doesn't appear when registering a local app manifest
I don't know if I miss something, but with very simple testcase, the icon doesn't show up in dashboard when I register a local manifest:

```
/myapp/ico.png
/myapp/manifest.webapp:
{"name": "foo", "icons": {"128": "ico.png"}}
- or - 
{"name": "foo", "icons": {"128": "/ico.png"}}
```

ochameau:
Fixes https://github.com/mozilla/r2d2b2g/issues/548

### DIFF
--- a/addon/data/content/js/applist.js
+++ b/addon/data/content/js/applist.js
@@ -72,7 +72,11 @@ var AppList = (function() {
                 iconPath = app.origin + app.icon;
             } else if (app.type === "local") {
                 // use the file protocol for local apps
-                iconPath = "file://" + app.id.substring(0, app.id.lastIndexOf("/")) + app.icon;
+                // We need to support unix and win path separators
+                var lastPathSep = app.id.lastIndexOf("/");
+                if (lastPathSep == -1)
+                  lastPathSep = app.id.lastIndexOf("\\");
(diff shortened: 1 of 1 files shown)


mykmelez:
This would be easier if we had access to SDK APIs like `require('url').fromFilename` in the Dashboard.

addon/data/content/js/applist.js
```diff
@@ -72,7 +72,11 @@ var AppList = (function() {
                 iconPath = app.origin + app.icon;
             } else if (app.type === "local") {
                 // use the file protocol for local apps
-                iconPath = "file://" + app.id.substring(0, app.id.lastIndexOf("/")) + app.icon;
+                // We need to support unix and win path separators
+                var lastPathSep = app.id.lastIndexOf("/");
+                if (lastPathSep == -1)
+                  lastPathSep = app.id.lastIndexOf("\\");
```
mykmelez, line 78: Nit: put braces around the conditional block; also, indent it four spaces, which appears to be the standard in this file.

mykmelez:
Merged with nits fixed.

PR #638, opened 2013-06-19
renamed 'update' to 'Refresh', and new icons for everybody

### LINKED ISSUE #624: give Update icon and rename it to Push
The Update and Push buttons do about the same thing for their respective targets. But they have very different names, and one has an icon, while the other doesn't. We should give Update an icon and rename it for consistency with the Push button.

potch:
The "Connect" button is commented out for now, pending that work landing.

Fixes #624.

### DIFF
--- a/addon/data/content/index.html
+++ b/addon/data/content/index.html
@@ -43,7 +43,7 @@
                     </fieldset>
                 </div>
                 <h5 id="device-status" class="device-dependent">
-                    <img src="device.png" alt="Device"> Device connected.
+                    <img src="device.svg" alt="Device"> Device connected.
                 </h5>
                 <a id="help" class="item" href="https://developer.mozilla.org/en-US/docs/Tools/Firefox_OS_Simulator" target="_blank">Help</a>
             </header>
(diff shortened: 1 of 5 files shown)


addon/data/content/index.html
```diff
@@ -83,7 +83,10 @@ <h5 id="device-status" class="device-dependent">
                         <button class="action" data-action="undo">Undo</button>
                     {% else %}
                         <button class="action device-dependent" data-action="push">Push</button>
-                        <button class="action" data-action="update">Update</button>
+                        {#<button class="action connect device-dependent" data-action="connect">Connect</button>#}
```
mykmelez, line 86: cc: @ochameau so he's aware of this addition.

addon/data/content/reload.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mykmelez, line 1: Nit: I would prefer to call the file and class "refresh" as well for consistency with the button label (and App > Refresh menu item in the Simulator).

mykmelez:
Merged with nit fixed!

PR #640, opened 2013-06-19
prevent `this._remote is null` exception in position watcher (fix #579)

### LINKED ISSUE #579: geolocation watcher tries to update simulator that isn't running
After installing 4.0pre4, I'm noticing a bunch of these two messages showing up in pairs:

> info: r2d2b2g: Firefox sending geolocation response
> 
> Timestamp: 2013/05/30 16:27:18
> Error: this._remote is null
> Source File: resource://gre/modules/XPIProvider.jsm -> file:///Users/myk/Library/Application%20Support/Firefox/Profiles/j50wskkw.Default%20User/extensions/r2d2b2g@mozilla.org/bootstrap.js -> resource://gre/modules/commonjs/toolkit/loader.js -> resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/remote-simulator-client.js
> Line: 261

It looks like the geolocation watcher is trying to update the simulator with a new location even though the simulator isn't running.

mykmelez:


mykmelez:
@nickdesaulniers how does this look?

Two fixes:
1. check this._remote in the position watcher before trying to use it;
2. stop watching the position before emitting "clientClosed" (which is what causes _this._remote_ to get set to `null`).

PR #642, opened 2013-06-19
remove obsolete console & default app options (fixes #622)

### LINKED ISSUE #622: remove Console checkbox
Now that an app developer can connect remote devtools, and especially after they'll be able to connect them to a specific app via #592, the Error Console will no longer be particularly useful to them. It'll still be useful to Simulator developers, but that is not sufficient reason to expose the feature to app developers.

So we should remove the Console checkbox, and we should no longer open the Error Console if its corresponding preference is set. But we should still provide a mechanism to enable it (f.e. a different pref that Simulator developers can toggle via about:config).

mykmelez:
Originally I planned to remove just the console checkbox, but then I noticed there is other cruft in the code, i.e. functionality that we don't use anymore, in particular for specifying a "default app" on the b2g command line, so I removed that too. Which entrained additional code removal, like the (unused) ability to set a remote debugger client timeout in the options to _RemoteSimulatorClient.run_.

Originally I also planned to replace the checkbox with a hidden pref to enable/disable the Error Console, but then I realized that a remote toolbox connected to the "global process" will show chrome messages, including extension messages from prosthesis. So I don't think the Error Console is actually useful to anyone, not even Simulator developers, since we can connect remote tools to the global process to see these messages.

I also removed the code that displays the remote debugger port on the Dashboard if you're using Firefox 19 or older, since we only support Firefox versions from our four main distribution channels, which are currently 21-24.

There are probably some conflicts between these changes and the ones in the "toolbox per app" branch, but hopefully they are relatively minor and easy to resolve.

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -17,6 +17,7 @@ const Request = require('request').Request;
 const SStorage = require("simple-storage");
 const Gcli = require('gcli');
 const Simulator = require("simulator.js");
+const Prefs = require("preferences-service");
 
 Cu.import("resource://gre/modules/Services.jsm");
 
@@ -78,6 +79,12 @@ if (SStorage.storage.lastVersion != Self.version) {
 // - ensure apps xkeys are unique
 // - flag needsUpdateAll if there are active apps registered
 if (["install", "downgrade", "upgrade"].indexOf(Self.loadReason) >= 0) {
+  // Delete obsolete property and preference.
+  if (Services.vc.compare(lastVersion, "4.0pre7") < 0) {
+    delete SStorage.storage.defaultApp;
+    Prefs.delete("extensions.r2d2b2g.jsconsole");
+  }
+
   if (Simulator.apps) {
     let activeAppIds = Object.keys(Simulator.apps).
       filter(function (appId) !Simulator.apps[appId].deleted);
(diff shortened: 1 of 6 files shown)


mykmelez:
@ochameau can you review this?

addon/lib/main.js
```diff
@@ -78,6 +79,12 @@ if (SStorage.storage.lastVersion != Self.version) {
 // - ensure apps xkeys are unique
 // - flag needsUpdateAll if there are active apps registered
 if (["install", "downgrade", "upgrade"].indexOf(Self.loadReason) >= 0) {
+  // Delete obsolete property and preference.
+  if (Services.vc.compare(lastVersion, "4.0pre7") < 0) {
+    delete SStorage.storage.defaultApp;
+    Prefs.delete("extensions.r2d2b2g.jsconsole");
```
ochameau: `TypeError: Prefs.delete is not a function`
I think you want to use `Prefs.reset` instead.
https://addons.mozilla.org/en-US/developers/docs/sdk/latest/modules/sdk/preferences/service.html#reset%28name%29

ochameau:
Looks good.
Otherwise. I think we will have to keep an easy way to spawn a main process toolbox when we will get rid of the connect button, otherwise working on the simulator is going to be painfull. Having to look for the random port number in console and having to go through the connect page will make us loose tons of time when debugging.

mykmelez:
Sure, I'm open to adding an easy way to spawn a main process toolbox provided we don't expose it to app developers.

Issue #644, opened 2013-06-20
Instead of toggle logic, keep track of run state in dashboard


nickdesaulniers:
Instead of [toggling](https://github.com/mozilla/r2d2b2g/blob/7a656529d1c1190e39aca41b76e4b97f436e3914/addon/data/content/js/main.js#L147) which can get out of sync with the checkbox, we should actually check the state of the checkbox and post a start or stop message (which would have to be implemented).

PR #645, opened 2013-06-20
don't attach toggle handler until window.load Fixes #634

### LINKED ISSUE #634: Simulator starts on Firefox startup with session restore of Dashboard if it was running on Firefox shutdown
If the Simulator is running, and the Dashboard is open, and the user quits Firefox and then restarts it, and Firefox is configured to restore the user's session on startup, then it not only restores the Dashboard, but it also restarts the Simulator. That's quite unexpected and seems unlikely to be what the user wants.

I think it's also a recent regression, as I don't recall it happening until very recently (late last week?).

@potch @nickdesaulniers perhaps one of last week's Dashboard changes triggered this?

nickdesaulniers:
review? @potch
I removed the fixme comment and opened Issue #644 instead

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -10,11 +10,6 @@ var Simulator = {
   init: function() {
 
     this.toggler = $('#command-toggle')[0];
-    $(this.toggler).prop('checked', false).on('change', function(evt) {
-      // FIXME: Change to actual checkbox state
-      Simulator.toggle();
-    });
-
     var currentUrl;
     $('#add-app-url, #new-from-manifest').on('keyup change input', function(evt) {
       var url = $(this).val();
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
@potch 

> What if I run the simulator, close the simulator tab, and then open the simulator tab?

The simulator stays running after tab is closed.  When tab is reopened, the toggler jumps to running.  You can toggle it to stopped and it will kill the simulator.

You could always checkout my branch to test my patches `git remote add --track [patch's branch name] nick https://github.com/nickdesaulniers/r2d2b2g.git && git checkout nick/[patch's branch name]`  I _think_ that's right, but I usually use [sourcetree](http://www.sourcetreeapp.com/).

PR #647, opened 2013-06-20
Simulator unloads on last tab, so adb can unload. Fixes #641

### LINKED ISSUE #641: *test-receipts.test receipt update* error binding to port 8099
If I `make test` more than once in a row, then _test-receipts.test receipt update_ fails the second and subsequent times:

```
info: r2d2b2g: executing 'r2d2b2g/tests/test-receipts.test receipt update'
error: r2d2b2g: fail: Error binding to port 8099, did you forget to call simulator.unload? - true == false
info: r2d2b2g: Traceback (most recent call last):
    callback.apply(null, args);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 288, in null
    timer.setTimeout(function() { onDone(self); }, 0);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 458, in null
    options.onDone(self);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 440, in runNextTest
    self.start({test: test, onDone: runNextTest});
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 471, in start
    this.test.testFunction(this);
  File "resource://gre/modules/commonjs/sdk/test.js", line 64, in null
    test(assert, function() {
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/tests/test-receipts.js", line 88, in receiptUpdate
    assert.ok(false, "Error binding to port 8099, did you forget to call " +
  File "resource://gre/modules/commonjs/sdk/test/assert.js", line 103, in ok
    operator: "=="
  File "resource://gre/modules/commonjs/sdk/test/assert.js", line 89, in fail
    this._log.fail(message);
  File "resource://gre/modules/commonjs/sdk/deprecated/unit-test.js", line 88, in fail
    this.console.trace();
  File "resource://gre/modules/commonjs/sdk/lang/functional.js", line 71, in null
    return function() fn.apply(this, args.concat(Array.slice(arguments)));
```

@nickdesaulniers any idea what's going wrong here? I don't remember seeing this when we first landed the receipt tests, although it's possible that I missed it.

bkase:
Thanks for help narrowing down this bug @nickdesaulniers!

This patch makes `adb` be killed like it's supposed to after running `make test`.
`adb.js` was being loaded twice because:
`1) require("main") -> require("simulator") -> require("adb") // in test-main.js`
and
`2) require("simulator") -> require("adb") // in test-receipts`

This seemed reasonable, but it looks like we have to move those two top-level requires inside of the test cases so that the second loading executes completely after the first one. Also, we weren't killing `adb` after the last tab was closed (parameter-less unload will do `Simulator.kill` and will kill `adb` if necessary)

### DIFF
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -112,7 +112,7 @@ Tabs.on('ready', function() {
 Tabs.on('close', function() {
   // Kill process when the last tab is gone
   if (!Tabs.length) {
-    Simulator.kill();
+    Simulator.unload();
   }
   if (Simulator.worker) {
     Simulator.sendListTabs();
--- a/addon/test/test-main.js
+++ b/addon/test/test-main.js
@@ -1,6 +1,5 @@
-
 exports.test_run = function(test) {
-  var main = require("main");
+  const main = require("main");
   test.pass("Unit test running!");
 };
 
--- a/addon/test/test-main.js
+++ b/addon/test/test-main.js
@@ -1,6 +1,6 @@
-var main = require("main");
 
 exports.test_run = function(test) {
+  var main = require("main");
   test.pass("Unit test running!");
 };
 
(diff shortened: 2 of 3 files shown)


mykmelez:
test-main.js should have a license header. I'll add it while merging this change!

mykmelez:
Thanks @bkase and @nickdesaulniers!

PR #648, opened 2013-06-20
nicer remove button

potch:


nickdesaulniers:
That _is_ a nicer remove button!

PR #650, opened 2013-06-21
disable DOM battery API (fixes #536)

### LINKED ISSUE #536: Simulator crashes on startup when battery absent on Linux
When I attempt to start the simulator (4.0pre2) on my ArchLinux x86_64 laptop with Firefox 21, the simulator window appears, shows the "based on mozilla technology" image, and immediately shutdowns without showing the dashboard (it shows the animations with the three circles, and the window closes).

However, with almost the same setup, the simulator runs fine on my desktop computer in my office. So I went through some investigation and reinstalled my system package after package (it needed some spring cleaning anyway :-) ) until I could spot the one responsible for the problem.
With a system having only the bare minimum to run Firefox, the simulator started and was running fine. However, as soon as I installed the 'upower' package, the issue appeared again.
Upower is responsible for monitoring AC and battery power, among other things. But as my laptop has no battery present inside the slot, upower always reports 0% remaining battery capacity (while still properly reporting the battery as not present) which, I think, is very likely to be the reason why the simulator shuts down.

A workaround for this bug is to disable the battery API (changing the "dom.battery.enabled" to false in greprefs.js worked for me).

A way to fix the bug would be to always report 100% battery capacity when no battery is actually _present_.

One other possible enhancement would be to let the user set the battery capacity value the same way as it can set the GPS coordinates. It would be easier to test applications reacting on battery level without having to wait for your laptop battery to discharge (if you have a laptop and a battery that is).

Hope that helped.

mykmelez:
This simple change sets a pref that disables the DOM battery API (navigator.battery). With this change, Gaia no longer shuts down on startup if a battery is absent on Linux (#536). However, the Settings app doesn't handle the situation perfectly, as it correctly exits early when it can't access the API, but then it tries to access it anyway and triggers an exception:

```
[22:35:47.467] Could not get window.navigator.battery
[22:35:47.468] TypeError: battery is undefined @ app://settings.gaiamobile.org/js/battery.js:68
```

And the Battery section of Settings doesn't open when I click on it. That seems ok, but it looks like I also have trouble opening Device Information after clicking Battery. No matter how many times I click on it, it doesn't open if I previously clicked on Battery.

I also have trouble getting Gaia to respond to mouse click events after dragging to scroll the app (it takes 2-3 clicks to get Gaia to register a click), and I see a ton of these errors in the console:

```
[22:35:54.697] NotSupportedError: Operation is not supported @ chrome://desktop-helper.js/content/touch-events.js:167
```

But that happens both with and without this change. So it seems like a separate issue.

In any case, it might be hard to set up a system that can reproduce #536, but it's easy to test this change, since it disables the API on all systems. I've tested it on both the Linux VM that can reproduce #536 and a Mac installation. It behaves the same on both.

Note that this change might prevent people from testing the battery API. But hooking up the Simulator to a laptop's real battery is a poor way of "simulating" the API anyway, because some users have desktops without batteries, and even users who have laptops with batteries will find it cumbersome to drain their real battery just to test their app in a low battery state.

A good simulation of the battery API would actually simulate a battery and enable the user to set the battery to any state (just like the geolocation simulation lets you set the position). We don't have that simulation yet, and in the meantime, I'm ok with disabling the API. But I'm concerned about side effects like that Device Information issue I noticed.

@ochameau What do you think about this way of fixing #536?

(Another way of fixing it would be to override the API with a custom implementation. The API is defined in WebIDL, and I'm not sure if it's possible to override such APIs. But I see that Gaia's tools/extensions/activities/components/SystemMessageManager.js implements nsIDOMNavigatorSystemMessages, so perhaps we could use a component to override nsIDOMBatteryManager.)

### DIFF
--- a/build/override-prefs.js
+++ b/build/override-prefs.js
@@ -2,10 +2,3 @@
 // prosthesis/defaults/preferences/prefs.js, but these load earlier, so this
 // file is useful for prefs that are accessed before addons are loaded.
 
-// Disable the battery API so Gaia doesn't shut down the Simulator on machines
-// with a dead or missing battery that it misinterprets as a low power state.
-//
-// This has to be here to affect the state of BatteryManager, which checks it
-// before addons have loaded.
-//
-user_pref("dom.battery.enabled", false);
--- a/build/override-prefs.js
+++ b/build/override-prefs.js
@@ -1,0 +1,11 @@
+// Prefs that override GRE/Gaia prefs.  Note that other such prefs are in
+// prosthesis/defaults/preferences/prefs.js, but these load earlier, so this
+// file is useful for prefs that are accessed before addons are loaded.
+
+// Disable the battery API so Gaia doesn't shut down the Simulator on machines
+// with a dead or missing battery that it misinterprets as a low power state.
+//
+// This has to be here to affect the state of BatteryManager, which checks it
+// before addons have loaded.
+//
+user_pref("dom.battery.enabled", false);


ochameau:
I'd prefer identifying which code in gaia or gecko shut b2g down. Is it a crash or a deliberate app shutdown?
If that's deliberate, we should patch that code to prevent this. If that's a crash, that may be also worth fixing it!
Otherwise, we can also  mock battery API, we may be able to easily override it by using `JavaScript-navigator-property` explained over here:
  http://blog.techno-barje.fr//post/2013/02/14/how-to-write-a-webapi/
  https://developer.mozilla.org/en-US/docs/Developer_Guide/Adding_APIs_to_the_navigator_object

Fixing issues in Settings apps (and other places that expect to always have a working battery API) doesn't look like a valuable work, but I can agree to go that way if we can unblock people with dead batteries before other options.

mykmelez:
> I'd prefer identifying which code in gaia or gecko shut b2g down. Is it a crash or a deliberate app shutdown?

I explain this in https://github.com/mozilla/r2d2b2g/issues/536#issuecomment-19792127: it's a deliberate shutdown by the Gaia system app, which shuts down whenever the battery reaches a very low level, regardless of whether or not the device is plugged in.

> If that's deliberate, we should patch that code to prevent this. If that's a crash, that may be also worth fixing it!

I agree that we should fix this, and not only for desktop devices! I've seen this bug manifest on mobile devices too when the battery runs out. At that point, I plug in the device, start it back up, and it immediately shuts itself back down! So I can't use the device at all until the battery charges up, even though the device is plugged in and could be used.

I haven't filed a Gaia bug about this yet, but I will do so. Then perhaps we can take the change in the Gaia branch used by the Simulator, if you'd rather fix the Gaia bug than disable the Battery API. Otherwise, it might take a while to get the fix to the Simulator, since we'll have to get Gaia review and then v1-train approval.

> Fixing issues in Settings apps (and other places that expect to always have a working battery API) doesn't look like a valuable work, but I can agree to go that way if we can unblock people with dead batteries before other options.

It's worth fixing the Settings app too, since one part of it correctly checks for the presence of a battery, while another part of it incorrectly assumes the battery will be there. But it isn't more valuable than fixing the System app, and we can probably fix both in the same changeset.

mykmelez:
@ochameau Ok, I filed [bug 887634](https://bugzilla.mozilla.org/show_bug.cgi?id=887634) with the Gaia fix, cherry-picked the fix to my v1-train-integration branch, and updated this pull request to use that branch instead of disabling the battery API. Does this work for you?

mykmelez:
Here's the comparison between the v1-train and v1-train-integration branches (the latter of which also contains your touch events fix):

https://github.com/mykmelez/gaia/compare/mozilla-b2g:v1-train...mykmelez:v1-train-integration

ochameau:
That looks like the perfect way to fix this issue!

mykmelez:
Ok, thanks @ochameau, I'll merge it then.

PR #653, opened 2013-06-24
send ping to initialize SimulatorActor Fixes #652

### LINKED ISSUE #652: "use my current coordinates" doesn't work before push from Dashboard
If I start the Simulator and then tell it to use my current coordinates for the geolocation simulation, it doesn't actually do so. Apparently this is because it notifies "r2d2b2g:geolocation-start", but SimulatorActor doesn't observe the notification because it hasn't been instantiated yet.

Once I push an app from the Dashboard, though, SimulatorActor gets instantiated, and a subsequent request to "use my current coordinates" succeeds.

I suppose we could work around this problem by having RemoteSimulatorClient send a ping to the server onClientReady. Although I'd like to figure out why SimulatorActor isn't instantiated until DebuggerServer receives a message for it. Perhaps that's intentional, perhaps it's a bug, but either way it would be good to know.

nickdesaulniers:
review? @ochameau

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -92,6 +92,8 @@ const RemoteSimulatorClient = Class({
     // listeners and emit an high level "ready" event
     this.on("clientReady", function (remote) {
       console.debug("rsc.onClientReady");
+      // Needed for Debugger Server to initialize SimulatorActor
+      remote.client.request({to: remote.simulator, type: "ping"});
       this._remote = remote;
       emit(this, "ready", null);
     });


bkase:
This would also fix #571 

Refreshing hosted apps installed outside the dashboard will now work without explicitly waking up the simulator actors (tested this with @nickdesaulniers ) -- see [pull 573](https://github.com/mozilla/r2d2b2g/pull/573)

ochameau:
Looks good, but I'm not sure it fixes all geolocation issues. I get random results for unknown reasons.
But it may be related to buggy geolocation behavior in firefox itself.

PR #660, opened 2013-06-26
auto open dashboard for `make run` Fixes #657

### LINKED ISSUE #657: auto run the simulator when executing make run
After successful _make_ of **r2d2b2g** , the command

```
$ make run
```

ends up into running an blank Firefox window with an information bar below the URL Bar telling, 

```
Mozilla Firefox is free and open source software from the non-profit Mozilla-Foundation     <button>Know your rights...</button>
```

The status of the **make run** in the terminal stops at

```
:~/B2G/r2d2b2g$ make run
cd addon-sdk && . bin/activate && cd ../addon && cfx run --templatedir template/  
Welcome to the Add-on SDK. Run 'cfx docs' for assistance.
Using binary at '/usr/bin/firefox'.
Using profile at '/tmp/tmpRgwBd_.mozrunner'.
info: r2d2b2g: adb: init
```

Shouldn't the init also start the process of starting the simulator automatically, when seperately invoked from the blank firefox window through Web Developer, it ends up running this. I believe its supposed to happen automatically.

```
info: r2d2b2g: Simulator.onMessage getIsRunning
info: r2d2b2g: Simulator.onMessage getDeviceConnected
info: r2d2b2g: Simulator.onMessage listApps
info: r2d2b2g: Simulator.sendListApps
info: r2d2b2g: Simulator.onMessage listTabs
info: r2d2b2g: adb: Found ADB process running, not restarting
info: r2d2b2g: simulator.observe: adb-ready
info: r2d2b2g: adb: trackDevices
```

nickdesaulniers:
review? @mykmelez

Is there a good reason why we don't do this?

Now you shouldn't need the pref to auto open tabs and windows on reopen.

What sucks is that there's still a new tab open.  Not sure what the best way to fix is?  Maybe @ochameau has some advice?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ locales:
 	python build/make-locales.py
 
 run:
-	cd addon-sdk && . bin/activate && cd ../addon && cfx run --templatedir template/ $(BIN_ARG) $(PROFILE_ARG)
+	cd addon-sdk && . bin/activate && cd ../addon && cfx run --templatedir template/ $(BIN_ARG) $(PROFILE_ARG) --static-args="{ \"autoRun\": true }"
 
 package:
 	cd addon-sdk && . bin/activate && cd ../addon && cfx xpi --templatedir template/
--- a/addon/lib/main.js
+++ b/addon/lib/main.js
@@ -18,6 +18,7 @@ const SStorage = require("simple-storage");
 const Gcli = require('gcli');
 const Simulator = require("simulator.js");
 const Prefs = require("preferences-service");
+const Args = require("sdk/system").staticArgs;
 
 Cu.import("resource://gre/modules/Services.jsm");
 
@@ -264,4 +265,8 @@ if (PermissionSettings) {
         return "unknown";
     }
   };
+}
+
+if (Args.autoRun) {
+  Simulator.openHelperTab();
 }


mykmelez:
The purpose of `make run` is to give a developer a similar experience to what a user will get. But there are actually three different possible experiences on addon load, depending on why the addon is loading: initial install, upgrade, or Firefox startup.

And only one of those experiences--initial install--opens the Dashboard. On upgrade, the Dashboard closes when the old version unloads, and it doesn't reopen when the new version loads (although perhaps it should, if it was open in the old version). And when a user starts up Firefox with the addon installed, the Dashboard doesn't open unless it was open when the user quit Firefox and the browser is configured to reopen tabs from the previous session.

But the Add-on SDK currently sets the load reason to _startup_ when you invoke `cfx run`, on which `make run` is based. And there's no way to set it to another value, as noted in this [Jetpack discussion forum thread](https://groups.google.com/forum/#!topic/mozilla-labs-jetpack/Mw2_Mr5WivY).

I'd like to see that change. In particular, I think the reason should be _install_ when you invoke `cfx run` with a new profile and _upgrade_ when you invoke it with an existing one. Or perhaps there should be a command-line flag for specifying the reason, or discrete _install_ / _upgrade_ / _startup_ commands to _cfx_. And/or a way to install/upgrade an addon to a running instance of Firefox, which would be even better at simulating the actual user experience, since users install/upgrade the addon while Firefox is running; they don't do it by starting Firefox.

Regarding that last idea, @canuckistani has pointed me at @palant's [Extension Auto-Installer](https://addons.mozilla.org/en-US/firefox/addon/autoinstaller/), about which more in [this blog post](http://adblockplus.org/blog/faster-extension-development-cycle-install-changes-automatically).

Finally, @Mossop notes in the aforementioned thread that [bug 607818] would fix the problem, although I think it'll just change the load reason from _startup_ to _install_ (since `cfx run` uninstalls the addon on shutdown, so it always installs the addon anew) without making it possible to specify _upgrade_ (or, indeed, _startup_).

In any case, in the meantime, I'm not sure we should work around the issue for the Dashboard specifically via a command-line flag. I think I'd rather find some way to let you load the addon with the _upgrade_ and _install_ reasons in addition to the _startup_ reason, whether that is an upstream fix to the SDK, a workaround in the Simulator code, or a third-party tool like Extension Auto-Installer.

mykmelez:
Per that previous comment, I don't think we should do this, but I do think we should address this issue with the aforementioned upstream change to `cfx run`.

PR #662, opened 2013-06-26
install app on connect to app if app isn't already installed

mykmelez:
This changeset makes the new "connect to app" functionality install the app while connecting to it if it isn't already installed.

It does this by making _simulator.runApp_ a fully round-trip operation that calls its _next_ callback with an error if the app isn't already installed. Then it makes _simulator.connectToApp_ call _simulator.runApp_ with a callback that calls _simulator.updateApp_ and then _simulator.runApp_ again if it receives an _app-not-installed_ error.

Along the way I simplified some code in _simulator.runApp_ and the actor script. But mostly I made the code more complex, because the code in question uses callbacks to chain asynchronous functions, which quickly becomes complex when you actually start to handle error cases instead of silently ignoring them. In the long run, we should move this code to promises to make error handling of complex asynchonous function chains simpler.

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -926,7 +926,23 @@ let simulator = module.exports = {
     gConnectingToApp = true;
 
     let app = this.apps[id];
-    simulator.runApp(app, simulator.openToolboxForApp.bind(simulator, app));
+    this.runApp(app, (function(error) {
+      if (error) {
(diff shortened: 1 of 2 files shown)


mykmelez:

ochameau:
So yes, I can't say it looks good, as it looks like the existing callback noodles!
But it works fine and does what we have to do.

PR #667, opened 2013-06-27
show throbber by receipt label when updating receipt (fixes #613)

### LINKED ISSUE #613: no feedback on status of Receipt Type selection
When I select a receipt type from the Receipt Type selector for an existing app, nothing appears to happen. This is disconcerting, even if the selection is actually working as expected, because there's no way to know if it really did without pushing the app to the Simulator and seeing if it behaves the way you would expect it to behave with the given receipt.

A Dashboard app entry should provide information about the status of a Receipt Type selection, i.e. inform the user via some kind of status indication that a selection is being processed and then that it has been completed successfully (or failed for some reason).

mykmelez:
@potch does this seem reasonable?

I'm not entirely happy with setting _app.updateReceipt_, since the _app_ object comes from SimpleStorage and persists across browser sessions, whereas an "update receipt" request always terminates with the session. But there is no session-specific storage at the moment, so setting a property of the _app_ object is the best thing we can do. And it should be ok, as the callback deletes the property when the request completes (which should happen even if the user terminates the session in the middle of the request).

In addition to setting _app.updateReceipt_, which ensures the throbber will appear if the user reloads the Dashboard while an update request is underway, the changeset messages the Dashboard when the request starts/stops, so it can update the specific app record. I don't think we have code that does this yet; currently we regenerate the entire app list anytime anything changes. But it seems preferable to update the specific record than to regenerate the whole list, especially if regeneration takes non-negligible time, so the list doesn't flash in a distracting way.

Ultimately we'll want a more general solution, but that can wait until another development cycle.

### DIFF
--- a/addon/data/content/css/main.css
+++ b/addon/data/content/css/main.css
@@ -348,6 +348,15 @@ figure {
 .receipt-type {
     padding: 1px 0;
 }
+.receipt-label {
+    padding-left: 23px;
+}
+.app.updateReceipt .receipt-label {
+    background-image: url(../throbber.svg);
+    background-repeat: no-repeat;
+    background-position: 6px center;
+    background-size: auto 14px;
+}
 #apps-list .options button, #apps-list .options a {
     padding: 4px 10px 3px;
     font-weight: normal;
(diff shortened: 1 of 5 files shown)


potch:
I agree with the notion of updating a single app in AppList. Will file an enhancement against myself for v5.

PR #670, opened 2013-06-27
notify user and reset receipt menu on fetch receipt failure

mykmelez:
When an "fetch receipt" request fails, the user isn't notified, and the Receipt Type dropdown menu in the Dashboard doesn't change back to the value it originally had, so it looks like the Dashboard successfully fetched a new receipt.

The Dashboard should notify the user that it failed to fetch the receipt, and the dropdown menu should change back to the original receipt type.

cc: @nickdesaulniers

@potch: seem ok?

Note that the change exposes the specific error message, which might be a bit too technical. But I think it's better than giving a vague, generic message; it'll help us debug user issues.

Also note that the change calls _sendListApps_, which regenerates the whole list. That's overkill, but it's the current convention, and since this is an edge case, it doesn't seem worth making another one-off to update just the specific app record. We can address that in the next cycle.

Finally, note that I add a _return_ to the _onComplete_ callback, which is an unrelated change that makes that callback always return a value (otherwise my JavaScript strict syntax checker complains that the function doesn't always do that).

### DIFF
--- a/addon/lib/simulator.js
+++ b/addon/lib/simulator.js
@@ -435,7 +435,8 @@ let simulator = module.exports = {
         delete app.updateReceipt;
         this.postUpdateReceiptStop(appId);
         if (err || !receipt) {
-          console.error(err || "No receipt");
+          this.error("Error getting receipt: " + (err || "unknown error"));
+          this.sendListApps();
         } else {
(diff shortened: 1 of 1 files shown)


potch:
Nice clean patch.

PR #672, opened 2013-06-28
clear indexedDB on "refresh app + clear cache and storages"

### LINKED ISSUE #666: clear IndexedDB when doing data-clearing refresh
Apropos #595, @digitarald says it would also be useful to clear the IndexedDB when doing an app refresh that clears persistent data.

cc: @rpl, @mattbasta, @cvan

rpl:
added a new "clear indexedDB" step to the "refresh + clear" hidden shortcut (fix #666)

NOTE: tested using https://github.com/mozilla/high-fidelity

### DIFF
--- a/prosthesis/content/shell.js
+++ b/prosthesis/content/shell.js
@@ -123,6 +123,12 @@ function simulatorAppUpdate(clearCacheAndStorages) {
     // clear cookies
     Services.cookies.removeCookiesForApp(localId, false);
 
+    // clear indexedDB
+    // NOTE: extracted from "resource://gre/modules/ForgetAboutSite.jsm"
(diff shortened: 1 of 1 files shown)


mykmelez:
Brilliant! Note that the interface has changed on mozilla-central, per [bug 767944](https://bugzilla.mozilla.org/show_bug.cgi?id=767944), to:

```
Cc["@mozilla.org/dom/quota/manager;1"].
getService(Ci.nsIQuotaManager).
clearStoragesForURI(iframe.contentDocument.documentURIObject);
```

So once we start supporting a newer version of B2G, we'll need to update this code to handle both versions.

mykmelez:
Thanks for the super-quick fix!

PR #677, opened 2013-06-29
window.locationbar/menubar/personalbar/statusbar/toolbar should be 'visible == false' 

### LINKED ISSUE #675: window.locationbar.visible is true, should be false
Apps are checking if they run chromeless using window.locationbar.visible. In B2G and on Desktop this values is consistently false when the app runs in chromeless mode.

In Simulator it returns true, not sure where it gets that from. It is probably similar to the screen size issue we had before.

Affects a lot of apps, as its still a recommendation mentioned in the MDN.

rpl:
added a new component to fake all the bars properties that are "visible == false" on the device (window.locationbar fix #675)

NOTE: tested using https://github.com/rpl/firefoxos-test-apps/tree/master/test-bars


prosthesis/chrome.manifest
```diff
@@ -13,3 +13,4 @@ overlay chrome://prosthesis/content/shell.xul chrome://prosthesis/content/shell-
 
 manifest components/FakeGeolocationProvider.manifest
 manifest components/SimulatorScreen.manifest
+manifest components/SimulatorBars.manifest
```
mykmelez: Nit: everything else is named "SimulatorBar" or "simulator-bar" singular, so this file's name should be singular too for consistency.

prosthesis/components/SimulatorBars.js
```diff
@@ -0,0 +1,50 @@
...
+let Cc = Components.classes;
+let Cu = Components.utils;
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+
+let DEBUG = true;
+let DEBUG_PREFIX = "prosthesis: SimulatorBar.js - ";
+let debug = DEBUG ?
+  function debug() dump(DEBUG_PREFIX + Array.slice(arguments) + "\n") :
+  function() {};
```
mykmelez: The _DEBUG_ flag is unnecessary, because logging of messages the Simulator dumps is controlled at runtime by the log level preferences. So this should simply dump unconditionally.

Also, automatic prefixes like the one in DEBUG_PREFIX tend to get expensive in terms of the amount of space they take up in the log and the amount of time it takes to read them. And they usually don't help identify the source of a log message very much, since the message is generally unique enough to be identifiable on its own (or with a simple code search).

In this case the only message being logged is "loading component definition", which is one of the rare messages that is actually ambiguous. But it's easy to fix that by changing the message to "loading SimulatorBar", and that's much shorter and easier to read than "prosthesis: SimulatorBar.js - loading component definition."

So I would remove _DEBUG_PREFIX_ in favor of short, sweet, and recognizable log messages.

rpl:
@mykmelez I've rebased the pull request and applied the suggested tweaks

mykmelez:
Looks great; thanks, @rpl!

PR #682, opened 2013-07-01
Updated Custom Profiles section. fixes #665

### LINKED ISSUE #665: CONTRIBUTING.md talks about step 9 when there isn't any
The current file says 

```
## Using Custom Profiles ##

If you need to use a custom Firefox profile, you can skip step 9 and run via
the following command

`PROFILE=/path/to/arbitrary/profile make run`
```

Updating it by either removing the step 9 part and changing the portion as

If you wish to use a custom Firefox profile which is the best way to simulate Gaia in the browser, the URL specified in this command is actually intercepted and routed to a local web server that's started while using this profile (there's a custom Firefox extension that runs the local server, while using this profile).

```
/path/to/firefox -profile /Volumes/Mozilla/b2g/B2G/gaia/profile http://system.gaiamobile.org:8080
```

On Mac OS X

```
/Applications/FirefoxNightly.app/Contents/MacOS/firefox -profile /path/to/B2G/gaia/profile http://system.gaiamobile.org:8080
```

Or anything else appropriate.

sudheesh001:


nickdesaulniers:
Looks good @sudheesh001 !  I'll add you to the authors file shortly!  Thanks for your contribution!

PR #683, opened 2013-07-01
Remove toggle logic from dashboard Fixes #644

### LINKED ISSUE #644: Instead of toggle logic, keep track of run state in dashboard

nickdesaulniers:
review? @mykmelez or @potch

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -10,6 +10,7 @@ var Simulator = {
   init: function() {
 
     this.toggler = $('#command-toggle')[0];
+    $(this.toggler).on('change', this.toggle.bind(this));
     var currentUrl;
     $('#add-app-url, #new-from-manifest').on('keyup change input', function(evt) {
       var url = $(this).val();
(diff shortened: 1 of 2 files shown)


mykmelez:
With this patch applied, when I start Firefox and it reopens a Dashboard tab that was open in my last session, the Dashboard gets stuck on "Loading your apps…" and never shows me my apps.

mykmelez:
It looks like #645 is what regressed this. And the latest commit seems to fix it.

mykmelez:
> With this patch applied, when I start Firefox and it reopens a Dashboard tab that was open in my last session, the Dashboard gets stuck on "Loading your apps…" and never shows me my apps.

This is unrelated [Firefox bug 888972](https://bugzilla.mozilla.org/show_bug.cgi?id=888972).

PR #689, opened 2013-07-02
Allow the simulator to update a single app row.

potch:
Sometimes, we don't need to re-render all apps. Actually, most times we don't. I added an `sendSingleApp` method to the simulator, which works with `AppList.update()` to re-render a single row of the dashboard. I also migrated all the `sendListApps` calls that were only updating a single app to use the new method.

### DIFF
--- a/addon/data/content/js/applist.js
+++ b/addon/data/content/js/applist.js
@@ -25,23 +25,48 @@ var AppList = (function() {
     var apps = {};
     var appIds = [];
 
-    function update(data) {
+    function updateAll(data) {
+        console.log('updating all apps');
+        if (!data) return;
         apps = data;
         appIds = Object.keys(apps).sort();
         render();
     }
 
-    function render() {
-        listEl.empty();
+    function update(id, data) {
+        console.log('updating', id);
+        var app = apps[id];
+        if (!app) return;
+        if ('key' in app) {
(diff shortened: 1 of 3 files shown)


addon/data/content/js/applist.js
```diff
@@ -25,23 +25,48 @@ var AppList = (function() {
...
         appIds = Object.keys(apps).sort();
         render();
     }
 
-    function render() {
-        listEl.empty();
+    function update(id, data) {
+        console.log('updating', id);
+        var app = apps[id];
+        if (!app) return;
```
mykmelez, line 39: Nit: we've been following the emerging Mozilla convention to always surround conditional blocks with parentheses, even one-liners.

mykmelez:
Other than that nit, this looks good, but `make test` times out, apparently on _test receipt update_.

mykmelez:
cc: @nickdesaulniers for his thoughts on the test timeout.

nickdesaulniers:
I added a patch to fix up the test in #706 .  @potch 's patch simplified the testing logic.  Most of my patch is just the un-nesting of a block from a removal of a now unnecessary conditional.

nickdesaulniers:
I'm reusing this patch in #706 .  Thanks @potch !

Issue #685, opened 2013-07-02
Packaged apps CSP is not being enforced in the simulator in alignment with the device in regards to inline JS


jds2501:
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=889356. This could be either a simulator bug or a desktop build bug.

The packaged app CSP is not being enforced correctly per bug 889356. We had an app developer build an app that was in clear violation of it by using inline JS, but the simulator still allowed execution of inline JS. We should be blocking execution of inline JS in the simulator in alignment with the CSP for packaged apps.

ochameau:
https://github.com/mozilla/r2d2b2g/pull/705 should fix the `redirects` support for simulator.

mykmelez:
#705 has fixed `redirects` support, but I'm going to leave this issue open to address the CSP problem.

PR #705, opened 2013-07-09
Fix app manifest `redirects` property support.

ochameau:
The webapps actors isn't synced with webapps.jsm new feature.
We are (at least) missing `redirect` property support when installed a privileged app.
https://developer.mozilla.org/en-US/docs/Web/Apps/Manifest#redirects

Here is a simple fix for the simulator, but we should upstream this fix (and the other features we have added like receipts) and use the upstream version!

I also identified similar issue for CSP, but this time, we are missing upstream fixes... so I'll try to sync with upstream!

This patch only fixes the original issue being reported in #685's bugzilla ticket.

### DIFF
--- a/prosthesis/content/dbg-webapps-actors.js
+++ b/prosthesis/content/dbg-webapps-actors.js
@@ -76,13 +76,14 @@ WebappsActor.prototype = {
                                requestID: "bar"
                              });
 
+        reg.broadcastMessage("Webapps:AddApp", { id: aId, app: aApp });
         reg.broadcastMessage("Webapps:Install:Return:OK",
                              { app: aApp,
                                oid: "foo",
                                requestID: "bar"
                              });
         delete aApp.manifest;
-        reg.broadcastMessage("Webapps:AddApp", { id: aId, app: aApp });
+
         self.conn.send({ from: self.actorID,
                          type: "webappsEvent",
                          appId: aId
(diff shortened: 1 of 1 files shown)


ochameau:
You can test that patch with this packaged app:
 http://test.techno-barje.fr/manifests/redirect/redirect.zip
That tries to load `http://test.techno-barje.fr/manifests/redirect/index.html`
Where I have the following htaccess being registered:

```
Redirect permanent /manifests/redirect/index.html http://test.techno-barje.fr/manifests/redirect/redirected.html
```

mykmelez:
With this change, I still see "Failed, redirected to the online version! (failed to catch the redirect and open the local file instead)" when I start the app, and also when I click the link in the app. Shouldn't I see "Redirect to local file!"?

ochameau:
Oh sorry. I thought it didn't depend on uplifting m-c fixes, but it actually does.
So I ended up cherry picking bug 843802, but it was full of conflicts :(
I tested hosted and packaged apps, but I don't really know how to test the receipt feature?

ochameau:
While we are at it, I also applied the only one patch left we miss on the webapps actor - https://bugzilla.mozilla.org/show_bug.cgi?id=842725

prosthesis/content/dbg-webapps-actors.js
```diff
@@ -109,49 +110,76 @@ WebappsActor.prototype = {
...
-            self._registerApp(app, aId, aDir);
+            let appType = self._getAppType(aManifest.type);
+
+            // In production builds, don't allow installation of certified apps.
+//#ifdef MOZ_OFFICIAL
+//            if (appType == Ci.nsIPrincipal.APP_STATUS_CERTIFIED) {
+//              self._sendError("Installing certified apps is not allowed.", aId);
+//              return;
+//            }
+//#endif
```
mykmelez, line 149: Commenting out this code seems like the best way to disable it while preserving the original structure (and line numbers) of the file as much as possible.

PR #709, opened 2013-07-11
ignore editor backups and downloaded files.

vicamo:


mykmelez:
@vicamo Thanks for the pull request! In the Simulator project, we follow the recommendations of GitHub's [Using Git - Ignoring files](https://help.github.com/articles/ignoring-files), which is to put package extensions like _.tar.gz_ and _.zip_ into a global _.gitignore_ file and "files created by your editor" into the local per-repository file _.git/info/exclude_ (although those too could go into the global _.gitignore_ file).

So I recommend you create a global .gitignore file per the instructions in that document and then put these rules there. Does that seem reasonable?

vicamo:
sure! thank you.

PR #716, opened 2013-07-12
Added addon/r2d2b2g.xpi to .gitignore

bkase:


mykmelez:
Normally this package extension is the kind of thing developers should put into a global .gitignore file, as recommended in [Using Git - Ignoring files - Global .gitignore](https://help.github.com/articles/ignoring-files#global-gitignore). But the _.xpi_ extension is an unusual one, little used outside Mozilla; and it isn't in the list of package extensions referenced by that document. So it's reasonable to include it in this Simulator-specific .gitignore.

Issue #715, opened 2013-07-12
Start the simulator with a custom profile path


kumar303:
It would be great to have a hidden pref that developers can toggle to set a custom path for [-profile](https://github.com/mozilla/r2d2b2g/blob/master/addon/lib/remote-simulator-client.js#L473) when starting the simulator. This would allow one to check out the gaia source, set custom preferences, build a custom profile, and get all the benefits of the simulator (dashboard, etc). This is an essential feature for developing the Firefox Marketplace because we need custom prefs to allow Dev payments.

mykmelez:
@kumar303 and I discussed this on IRC, and I wouldn't want to expose this feature to regular users of the addon, but it seems like a reasonable hidden feature to provide to expert users who have this particular need. You can currently point the B2G executable inside the Simulator at another profile via the _-profile_ command line option, but that doesn't give you access to the Dashboard, so it isn't as helpful as a hidden pref would be.

cvan:
:+1: this would help immensely with payments development!

ochameau:
We need special flags when building gaia and also needs prothesis to be installed in this profile.
I tend to think it will be easier to do that from simulator sources.

You can already use a custom profile by running the simulator from source:

```
$ git clone https://github.com/mozilla/r2d2b2g.git
$ cd r2d2b2g
$ cd gaia
// do whatever necessary modification to gaia sources
// by default the simulator uses v1-train branch, but you can checkout master
$ cd ..
// We have similar files to what exists in gaia/build folder, to set custom settings/prefs
$ vi build/override-settings.json
$ vi build/override-prefs.js
$ make profile run
// And in order to avoid always using a new empty profile
$ BIN_ARG="-p /tmp/profile" make profile run
```

kumar303:
@ochameau I can already start b2g from the addon source (e.g. `~/Library/Application Support/Firefox/Profiles/.../r2d2b2g@mozilla.org`) with a custom profile but that doesn't let me hook into the dashboard and all the nice simulator features. This issue is just asking to make the `-profile` path configurable here: https://github.com/mozilla/r2d2b2g/blob/master/addon/lib/remote-simulator-client.js#L473

ochameau:
I crafted that patch, but then you miss prosthesis addon and all dashboard features are lost.

ochameau:
See PR #722.

mykmelez:
> I crafted that patch, but then you miss prosthesis addon and all dashboard features are lost.

[...]

PR #722, opened 2013-07-15
Introduce a pref to select a custom profile folder.

ochameau:
I'm not sure that patch is really usefull as we need a simulator profile, with precise gaia build flags and prosthesis addon being installed.

### DIFF
--- a/addon/lib/remote-simulator-client.js
+++ b/addon/lib/remote-simulator-client.js
@@ -18,6 +18,7 @@ const Self = require("self");
 const URL = require("url");
 const Subprocess = require("subprocess");
 const { setTimeout, clearTimeout } = require("sdk/timers");
+const Prefs = require("preferences-service");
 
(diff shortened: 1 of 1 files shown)


kumar303:
I started building a custom addon per issue #715 but got side tracked and haven't finished yet. However, I think issue #715 explains that adding a custom pref is useless because the profile won't work. So it sounds like this is a wontfix issue, right?

mykmelez:
> I started building a custom addon per issue #715 but got side tracked and haven't finished yet. However, I think issue #715 explains that adding a custom pref is useless because the profile won't work. So it sounds like this is a wontfix issue, right?

That's basically right, but if building a custom addon didn't work for you, then we'd think about other ways to meet your needs, which might include both this pref and some other work.

In any case, it sounds like the next step is to figure out if building a custom addon can meet your needs, and we can close this request in the meantime, since any alternative will involve more than just this change in any case.

PR #730, opened 2013-07-18
set prefs to enable push notifications (fixes #724)

### LINKED ISSUE #724: Configure server for push notifications
To test push notifications, the Simulator should have push notifications enabled and configured.

```
user_pref("services.push.enabled", true);    
// Via JR from services/ops team
user_pref("services.push.serverURL", "wss://push.services.mozilla.com");
```

mykmelez:
@digitarald: can you confirm that this enables push notifications for you?

### DIFF
--- a/build/override-prefs.js
+++ b/build/override-prefs.js
@@ -2,3 +2,5 @@
 // prosthesis/defaults/preferences/prefs.js, but these load earlier, so this
 // file is useful for prefs that are accessed before addons are loaded.
 
+user_pref("services.push.enabled", true);
+user_pref("services.push.serverURL", "wss://push.services.mozilla.com");


digitarald:
Gotta find or create a working app with push notifications. I should have a green light on this next week.

mykmelez:
@digitarald have you had a chance to test this yet?

mykmelez:
@nikhilm pointed out that we don't actually need these prefs, because they're already set in B2G Desktop. However, we do need the fix for [bug 856495](https://bugzilla.mozilla.org/show_bug.cgi?id=856495), so I've backported it to b2g18 in this branch:

Nikhil pointed out that B2G18 needs the fix in bug 856495, so I backported it to b2g18 in the https://github.com/mykmelez/mozilla-central/tree/b2g18-856495-push-notifications branch, then I merged into the https://github.com/mykmelez/mozilla-central/tree/b2g18-simulator-integration branch (which is currently identical to the upstream b2g18 branch, except for this fix).

I'm now building B2G from b2g18-simulator-integration and will update the Simulator to use those builds once they're done, after which I'll spin and publish preview builds of the Simulator with the new B2G builds.

PR #731, opened 2013-07-19
purge permissions added by obsolete hidden feature

mykmelez:
As @digitarald and I discussed on IRC today, we should purge the permissions we add to Firefox's permissions database for hosted apps, since that's an obsolete hidden feature that we never exposed to users, and they won't expect those permissions to have been added.

Here's an initial take that should work, although I haven't tested it yet. @ochameau, I'll want you to review this once it's ready, but it isn't ready yet!


mykmelez:
Closing this obsolete pull request.

PR #732, opened 2013-07-19
validate manifest file name

ccarruitero:


addon/lib/validator.js
```diff
@@ -127,6 +127,16 @@ exports.validateNameIcons = function(errors, warnings, manifest, app) {
   }
 }
 
+exports.validateManifestFile = function(errors, warnings, app){
+  let appType = app.manifest.type;
+  if (appType !== "web") {
```
mykmelez: The [_type_ property](https://developer.mozilla.org/en-US/docs/Web/Apps/Manifest#type) is optional, and many apps don't set it, so checking _app.manifest.type_ isn't a reliable way to determine that an app is packaged. Instead, check _app.type_, which is either `local`, `hosted`, or `generated`, where `local` means "packaged".

Also, it isn't necessary to add a _fileName_ property to the app record, because the app ID is the file path for packaged apps. So you just need to pass the ID into the validator function and use _File.basename_ to extract the name of the file from it.

Finally, I would move the _app.type_ check up the caller ( _validateApp_ ), so it only calls _Validator.validateManifestFile_ when the app is packaged, i.e.:

```
  validateApp: function(id, next) {
    let app = simulator.apps[id];
    …
    if (app.type == "local") {
      Validator.validateManifestFile(app.validation.errors,
                                     app.validation.warnings,
                                     id);
    }
```

Then, in validator.js:

```
// at the top of the file:
const File = require("file");
…
exports.validateManifestFile = function(errors, warnings, path) {
  if (File.basename(path) !== "manifest.webapp") {
    errors.push("the manifest file must be called 'manifest.webapp'");
  }
}
```

addon/lib/simulator.js
```diff
@@ -129,8 +129,10 @@ let simulator = module.exports = {
     fp.appendFilters(Ci.nsIFilePicker.filterAll);
 
     let ret = fp.show();
-    if (ret == Ci.nsIFilePicker.returnOK || ret == Ci.nsIFilePicker.returnReplace) {
+    if (ret == Ci.nsIFilePicker.returnOK ||
+        ret == Ci.nsIFilePicker.returnReplace) {
```
mykmelez, line 138: Nice catch!

.gitignore
```diff
@@ -18,3 +18,9 @@ addon/r2d2b2g.xpi
 
 # locale repositories cloned by build/make-locales.py
 gaia-l10n
+
+# files in use
+*.swp
+
+# b2g build
+b2g-*
```
mykmelez: When it comes to the project's .gitignore file, we generally ask developers to follow the recommendations of [Using Git > Ignoring Files](https://help.github.com/articles/ignoring-files), which is to put most ignore rules into either a global .gitignore file or local per-repository .git/info/exclude files. b2g-\* is covered by the package rules in the global .gitignore file recommended by that doc, while *.swp seems like the kind of editor-generated file that also makes sense to add to your global .gitignore file.

mykmelez:
@ccarruitero Sorry about the delay reviewing, and thanks for the useful patch! It just has a few minor issues. Any chance you can fix it up and prepare it for re-review? With the requested fixes, I will be very happy to merge it!

ccarruitero:
@mykmelez don't worry about delay. I will resend patch

mykmelez:
Thanks for the fixes! I merged them with two minor code style nits.

PR #735, opened 2013-07-22
Add a page-mod against marketplace to ease reviewer life.

ochameau:
Thanks @diox, who registered me some test account on marketplace dev server and introduced me to the reviewer UI!

That page-mod adds a button in reviewer UI, to easily register an hosted or packaged app, meant to be reviewed, in the Simulator. I choosed to register the app to the dashboard, as it is simplier to implement that way, but may be we want to bypass the dashboard and only push the app to b2g? I can easily see a reviewer dashboard made of tons of apps being reviewed, whereas he don't really want them to be registered...

@myk What do you think about a page-mod? Do you think this code is meant to be hosted in the simulator or should we make another addon that depends on it? Or do something completely different? Is marketplace the only one website where we would like to integrate with the simulator?

### DIFF
--- a/addon/data/marketplace-script.js
+++ b/addon/data/marketplace-script.js
@@ -1,0 +1,30 @@
+/*
+ * Hack the marketplace reviewer UI to add a "Run in simulator" button
+ * to ease reviewer app testing
+ */
+
+let installButton = document.querySelector("button.product.install");
+if (installButton) {
+  let simulatorButton = installButton.cloneNode();
+  simulatorButton.textContent = "Run in simulator";
+  simulatorButton.style.marginTop = "1em";
+  simulatorButton.removeAttribute("disabled");
+  simulatorButton.classList.remove("disabled");
+  // Remove this class to prevent the button listener click to trigger regular app install
+  simulatorButton.classList.remove("product");
+  installButton.parentNode.insertBefore(simulatorButton,
+                                        installButton.nextSibling);
+  simulatorButton.addEventListener("click", function () {
+    if (installButton.dataset.is_packaged === "true") {
+      self.postMessage({
+        type: "packaged",
+        miniManifestURL: installButton.dataset.manifest_url
+      });
+    } else {
+      self.postMessage({
+        type: "hosted",
+        manifestURL: installButton.dataset.manifest_url
+      });
+    }
+  });
+}
--- a/addon/lib/marketplace-mod.js
+++ b/addon/lib/marketplace-mod.js
@@ -1,0 +1,90 @@
+const { PageMod } = require("sdk/page-mod");
+const Self = require("self");
+const File = require("file");
+const Request = require('./request').Request;
+const simulator = require("./simulator");
+
+const { CC, Cu, Cc, Ci } = require("chrome");
+// use addon manager internal helper to extract a zip to a folder
+const { extractFiles } = Cu.import("resource://gre/modules/XPIProvider.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+const WebBrowserPersist = CC("@mozilla.org/embedding/browser/nsWebBrowserPersist;1",
+                               "nsIWebBrowserPersist");
+
+function downloadFile(url, path, onDone) {  
+  let persist = WebBrowserPersist();
+  persist.persistFlags = persist.PERSIST_FLAGS_REPLACE_EXISTING_FILES;
+  let uri = Services.io.newURI(url, null, null);
+  let file = Cc["@mozilla.org/file/local;1"].createInstance(Ci.nsIFile);
+  file.initWithPath(path);
+  persist.progressListener = {
+    onStateChange: function(aProgress, aRequest, aStateFlag, aStatus) {
+      if (aStateFlag & Ci.nsIWebProgressListener.STATE_STOP) {
+        onDone();
+      }
+    }
+  };
+  persist.saveURI(uri, null, null, null, null, file, null);
+}
+
+PageMod({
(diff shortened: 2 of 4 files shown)


mykmelez:
> That page-mod adds a button in reviewer UI, to easily register an hosted or packaged app, meant to be reviewed, in the Simulator. I choosed to register the app to the dashboard, as it is simplier to implement that way, but may be we want to bypass the dashboard and only push the app to b2g? I can easily see a reviewer dashboard made of tons of apps being reviewed, whereas he don't really want them to be registered...

Yes, it seems preferable to push the app directly to B2G, since the reviewer and developer workflows are quite different, and the reviewer UI is itself a kind of dashboard (in the sense that it presents a list of apps in review and gives you options for interacting with them).

If I was a reviewer, I would want to have a button in the reviewer UI that installs the app, starts the Simulator, and then starts the app in the Simulator, so I could go directly from the reviewer UI to the Simulator without having to open the Dashboard and press a button in it. I might also want a button that pushes the app to a device, if one is attached.

> @myk What do you think about a page-mod?

At first glance, a page-mod seems like a fine way to implement this!

> Do you think this code is meant to be hosted in the simulator or should we make another addon that depends on it? Or do something completely different?

As for hosting it in the Simulator, in the long run I think it makes sense to put this into a separate addon, because the audience for the functionality is different (a handful of reviewers vs. thousands of developers), and it will be easier to document the feature and market it to just the right set of users via a separate addon than if it is a hidden feature of the Simulator addon.

For the same reason, "push to device" will probably become a separate addon (or integrated into Firefox): it only matters for users who have devices and want to develop on them, which is a different set of users from those who want to develop on the Simulator (although there is some overlap).

And the Dashboard itself will be replaced by the App Manager integrated into Firefox.

Ultimately, I expect the Simulator addon to comprise just Gaia + B2G + the Simulator window that displays B2G/Gaia and gives you access to hardware simulation configuration (i.e. the home button along with interfaces for configuring geolocation, device orientation, network status, battery status, etc.) and developer tools (if we can build developer tools directly into the Simulator window).

Everything else, like the App Manager, Device Manager, "push to device", "push to Android", "push to desktop runtime", and app templates for creating an app (like those that @potch is adding to the Simulator), will either live in core Firefox or a separate addon.

Nevertheless, I'm fine with landing this functionality into the Simulator addon in the short run, if it's difficult to put it into a separate addon while enabling it to communicate with the Simulator.

> Is marketplace the only one website where we would like to integrate with the simulator?

I can't think of another website, but I could imagine also integrating the [Developer Hub area of the Marketplace](https://marketplace.firefox.com/developers/) with the Simulator, f.e. enabling developers to test the app installation process while their app is still under review, or letting developers create and register an app from a template on the site.

But I wouldn't worry too much about extending this addon to cover those use cases. Better to start with a single use case, create the best possible flow for satisfying it, and then figure out what else you might want to do. :-)

ochameau:
My main concern here is that it looks hard to be able to use simulator codebase to only open b2g and push the app to it. Otherwise, having a distinct addon doesn't sound really complicated as soon as we expose some communication channel in the simulator addon.

So it should be easy to move this feature to another addon requirering the simulator to be installed,
but it will be harder to make it so that it doesn't register apps to the dashboard.

mykmelez:
> My main concern here is that it looks hard to be able to use simulator codebase to only open b2g and push the app to it. Otherwise, having a distinct addon doesn't sound really complicated as soon as we expose some communication channel in the simulator addon.
> 
> So it should be easy to move this feature to another addon requirering the simulator to be installed,
> but it will be harder to make it so that it doesn't register apps to the dashboard.

Hmm, ok. Then let's start by integrating this into the Simulator addon and sticking with the current approach of registering apps to the Dashboard. It isn't a big problem to add the app to the Dashboard, since it's relatively simple for the reviewer to remove the app from the Dashboard after reviewing it. And it seems reasonable to integrate this into the Simulator at this stage, even if we eventually move it to its own addon (in the same way that we ship ADB in the Simulator, even though in the long run it'll be integrated into Firefox or its own addon).

addon/lib/marketplace-mod.js
```diff
@@ -0,0 +1,90 @@
...
+const File = require("file");
+const Request = require('./request').Request;
+const simulator = require("./simulator");
+
+const { CC, Cu, Cc, Ci } = require("chrome");
+// use addon manager internal helper to extract a zip to a folder
+const { extractFiles } = Cu.import("resource://gre/modules/XPIProvider.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+const WebBrowserPersist = CC("@mozilla.org/embedding/browser/nsWebBrowserPersist;1",
+                               "nsIWebBrowserPersist");
```
mykmelez, line 12: Nit: it looks like the indentation is off by a couple spaces here.

addon/data/marketplace-script.js
```diff
@@ -0,0 +1,30 @@
...
+  simulatorButton.addEventListener("click", function () {
+    if (installButton.dataset.is_packaged === "true") {
+      self.postMessage({
+        type: "packaged",
+        miniManifestURL: installButton.dataset.manifest_url
+      });
+    } else {
+      self.postMessage({
+        type: "hosted",
+        manifestURL: installButton.dataset.manifest_url
```
mykmelez, line 26: Nit: here and above, we've been adding trailing commas to such property lists lately so fewer lines appear to change when we add a property later.

mykmelez:
I see a couple trivial nits, but otherwise the code looks great, and I'm going to merge it as-is.

PR #740, opened 2013-07-24
Changed condition variable API so that XP would work

bkase:
The condition variable shim used Windows APIs that were only supported by Vista+. This patch rewrites the condition variable shim using APIs that are also supported on XP.

In order to fix a bug with using the Windows Event APIs, reading from USB is now done using the `AdbReadEndpointAsync` function.

### DIFF
--- a/android-tools/adb-bin/usb_windows.cpp
+++ b/android-tools/adb-bin/usb_windows.cpp
@@ -43,9 +43,9 @@
 #define AdbCloseHandle(...) (false)
 #endif
 
-#define D_ D
-#undef D
-#define D printf
+//#define D_ D
+//#undef D
+//#define D printf
 
 
 /** Structure usb_handle describes our connection to the usb device via
@@ -83,9 +83,6 @@ static usb_handle handle_list = {
   /* .prev = */ &handle_list,
   /* .next = */ &handle_list,
 };
-
-/// Locker for the should_kill_cond wait
-ADB_MUTEX_DEFINE( should_kill_cond_lock );
 
 /// Locker for the should_kill signal
 ADB_MUTEX_DEFINE( should_kill_lock );
@@ -145,10 +142,6 @@ static struct dll_bridge * bridge;
 extern struct dll_io_bridge * i_bridge;
 extern struct dll_io_bridge * o_bridge;
 
-//#define D_ D
-//#undef D
-//#define D printf
-
 int known_device_locked(const char* dev_name) {
   usb_handle* usb;
 
@@ -203,7 +196,7 @@ int register_new_device(usb_handle* handle) {
 
 
 static int should_kill = 0;
-static adb_cond_t should_kill_cond;
+ADB_COND_DEFINE(should_kill_cond);
 static void set_should_kill(int val) {
   adb_mutex_lock(&should_kill_lock);
   D("Set should_kill to %d\n", val);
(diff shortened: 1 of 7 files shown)


mykmelez:
@ochameau can you review this?

android-tools/adb-bin/usb_windows.cpp
```diff
@@ -398,10 +394,11 @@ int usb_write(usb_handle* handle, const void* data, int len) {
 
 extern THREAD_LOCAL int (*getLastError)();
 int usb_read(usb_handle *handle, void* data, int len) {
-  unsigned long time_out = 100;
+  unsigned long time_out = 10;
```
ochameau: `time_out` doesn't seem to be used anymore.

android-tools/adb-bin/usb_windows.cpp
```diff
@@ -227,22 +220,25 @@ void notify_should_kill(int k, char who) {
   }
 
   set_should_kill(k+1);
```
ochameau, line 222: I don't fully understand the purpose of this mutex value, but I tend to think we should do the `k+1` inside of the mutex, otherwise we can have two concurrent code colliding and one `+1` will be ignored.
Can you add a comment explaining the goal of this mutex?

ochameau, line 222: Also, do we really need a mutex for reading, i.e. do we need to use a mutex in get_should_kill?

ochameau, line 222: Oh if the mutex is only about to protect `should_kill++` collision and that we can get rid of get_should_kill completely (I don't see why we would need a mutex to read the value) Then the code would be quite explicit.

ochameau:
Unless I miss something really important, I see various useless mutex. The same applies to `get_io_pump_status`. I really don't see why we would need one for just setting/reading a boolean flag.

If that's the case, we should have two flags here, `should_kill` that won't need mutex and will only be set to true here:
`set_should_kill(1);`  --> `should_kill = true;`  in kill_threads
and read without mutex in device_poll_thread

```
int k = get_should_kill();
if (k) {
  D("Cleaning in timer handler\n");
```

->

```
if (should_kill) {
  D("Cleaning in timer handler\n");
```

Then, other usages should use a distinct value, the number of killed thread. This time, the number of kill thread will be read without mutex, but its increment in notify_should_kill will be protected with a mutex.

android-tools/adb-bin/usb_windows.cpp
```diff
@@ -410,35 +407,55 @@ int usb_read(usb_handle *handle, void* data, int len) {
...
+        saved_errno = getLastError();
+        if (complated_handle == NULL) {
+          D("AdbReadEndpointAsync, errno: %d\n", saved_errno);
+          return -1;
+        }
+        D("Read async started");
+
+      int hasComplated = FALSE;
+      do {
+        hasComplated = o_bridge->AdbHasOvelappedIoComplated(complated_handle);
```
ochameau: Is `complated` a realy english word?
If not I'm not sure it is really worth fixing adb typo in `AdbHasOvelappedIoComplated`,
but still worth not spreading it in our code.

android-tools/adb-bin/usb_windows.cpp
```diff
@@ -410,35 +407,55 @@ int usb_read(usb_handle *handle, void* data, int len) {
...
+      ret = o_bridge->AdbGetOvelappedIoResult(complated_handle, NULL, &read, true);
+      if (!ret) {
+        D("AdbGetOvelappedIoResult(read %u) failure (%u). Error %u, read %u\n",
+            xfer, ret, getLastError(), read);
+        return -1;
       }
+      D("usb_read got: %ld, expected: %d, errno: %d\n", read, xfer, saved_errno);
+
+      data_ += read;
+      len -= read;
```
ochameau, line 453: So if I follow correctly, we do not care about how much bytes we read during the call to `AdbReadEndpointAsync` as it looks like we ignore its `read` argument. Then as it is async we call `AdbHasOvelappedIoComplated` to see when it is done reading.
But I don't fully understand the purpose of `AdbGetOvelappedIoResult`. This time, we do care about its `read` argument. It looks like it testify the number of byte being read during AdbReadEndpointAsync, but aren't we introducing some collusion by reusing the same `read` variable for both?

ochameau:
If I'm right about mutex, we can address that in another pull request as it isn't really related to the work you are doing in this one.
Otherwise other comments should be clarified/fixed before landing

bkase:
Thanks for the feedback @ochameau . I think you're right about the mutexes-- I'll fix that up after this pull-request lands.

I don't know why the API is constructed in this way, but `AdbGetOvelappedIoResult` needs to be called before reading bytes out of the buffer. The value that `AdbReadEndpointAsync` writes to `&read` is ignored because it is reset by `AdbGetOvelappedIoResult` (I think the `AdbReadEndpointAsync` one is the amount read so far but not the complete amount which we don't need).

ochameau:
> The value that AdbReadEndpointAsync writes to &read is ignored because it is reset by AdbGetOvelappedIoResult (I think the AdbReadEndpointAsync one is the amount read so far but not the complete amount which we don't need).

It depends how `AdbReadEndpointAsync` is implemented. I was considering `AdbReadEndpointAsync` to keep a ref to `read` and eventually modify its value while `AdbGetOvelappedIoResult` is running. Looking at both of these method's implementation, they both immediatly set `read` to `0`, but later set it to some value. If there is any thread involved in this code, that would be a concrete issue, but it doesn't looks like they are using thread behind the scene?

PR #744, opened 2013-07-25
Added sources and changed Makefiles to build AdbWin*.dll

bkase:
The changeset adds the AdbWinApi.dll and AdbWinUsbApi.dll sources and modifies the Makefiles so that (after setting up the proper environment), these DLLs can be built from source.

Downloading WinDDK 7.1 is a pre-requisite for building these DLLs. The default settings install the DDK to C:\WinDDK\7600.16385.1. If you install it somewhere else, change the first line of _android-tools\adb-win-api\make_driver.bat_. WinDDK 7.1 can be installed here:
https://www.microsoft.com/en-us/download/details.aspx?id=11800

I wasn't sure if I should make a new target or not, so for now, the top level Makefile attempts to build AdbWin*.dll if it is configured to build adb from source and it is being executed from Windows. Let me know if this should be in a separate target or organized in a different way.


Makefile
```diff
@@ -213,7 +216,11 @@ adb:
 	fi;
 	if [ "$(LIBADB_LOCATION)" = "local" ]; then \
 	  make -C android-tools lib; \
-	  cp $(ADB_OUT_DIR)libadb$(LIB_SUFFIX) addon/data/$(B2G_PLATFORM)/adb; \
+	  cp $(ADB_OUT_DIR)libadb$(LIB_SUFFIX) $(DATA_ADB_PATH); \
+    make -C android-tools driver; \
+    cp $(ADB_DRIVERS_DIR)api/objfre_wxp_x86/i386/AdbWinApi$(LIB_SUFFIX) $(DATA_ADB_PATH) 2> /dev/null; \
+    cp $(ADB_DRIVERS_DIR)winusb/objfre_wxp_x86/i386/AdbWinUsbApi$(LIB_SUFFIX) $(DATA_ADB_PATH) 2> /dev/null; \
+		true; \
```
mykmelez: This works, but there's some funkiness here. Because these are separate statements, the last of which is `true;`, this always appears to succeed, even if one or more commands fails. And it isn't clear why it pipes STDERR to /dev/null for the last two file copies, since developers will want to know about any problems with the copies. Finally, these lines should begin with a tab character, even though they don't have to, in order to make it more obvious where the recipe begins and ends.

Here's how I would do this (fixing another indentation issue in the process):

```
Index: c:\Users\myk\Mozilla\r2d2b2g\Makefile
index 99fe924..81335f4 100644
--- a/Makefile
+++ b/Makefile
@@ -208,19 +208,18 @@ adb:
    cd addon/data/$(B2G_PLATFORM) && rm -rf adb $(ADB_BINARIES)
    mkdir addon/data/$(B2G_PLATFORM)/adb
    if [ ! -f $(ADB_PACKAGE) ] && \
-        ( [ "$(LIBADB_LOCATION)" = "remote" ] || [ $(DEPS) ] ); then \
+      ( [ "$(LIBADB_LOCATION)" = "remote" ] || [ $(DEPS) ] ); then \
      $(DOWNLOAD_CMD) $(ADB_URL); \
    fi;
    if [ -f $(ADB_PACKAGE) ]; then \
      unzip $(ADB_PACKAGE) -d addon/data/$(B2G_PLATFORM)/adb; \
    fi;
    if [ "$(LIBADB_LOCATION)" = "local" ]; then \
-     make -C android-tools lib; \
-     cp $(ADB_OUT_DIR)libadb$(LIB_SUFFIX) $(DATA_ADB_PATH); \
-    make -C android-tools driver; \
-    cp $(ADB_DRIVERS_DIR)api/objfre_wxp_x86/i386/AdbWinApi$(LIB_SUFFIX) $(DATA_ADB_PATH) 2> /dev/null; \
-    cp $(ADB_DRIVERS_DIR)winusb/objfre_wxp_x86/i386/AdbWinUsbApi$(LIB_SUFFIX) $(DATA_ADB_PATH) 2> /dev/null; \
-       true; \
+     make -C android-tools lib && \
+     cp $(ADB_OUT_DIR)libadb$(LIB_SUFFIX) $(DATA_ADB_PATH) && \
+     make -C android-tools driver && \
+     cp $(ADB_DRIVERS_DIR)api/objfre_wxp_x86/i386/AdbWinApi$(LIB_SUFFIX) $(DATA_ADB_PATH) && \
+     cp $(ADB_DRIVERS_DIR)winusb/objfre_wxp_x86/i386/AdbWinUsbApi$(LIB_SUFFIX) $(DATA_ADB_PATH); \
    fi;

 locales:
```

This makes the following changes:
- remove the `true;` statement and chain the rest of the statements together, so they all have to succeed;
- stop piping STDERR to /dev/null;
- prefix all recipe lines with tab characters (which isn't obvious from the HTML representation of the patch, but you should be able to see it if you edit this comment).

Makefile
```diff
@@ -137,6 +138,8 @@ ifdef TEST
   TEST_ARG = -f $(TEST)
 endif
 
+DATA_ADB_PATH = addon/data/$(B2G_PLATFORM)/adb
```
mykmelez: Nit: the other ADB-related variables all start with "ADB" or "LIBADB", so I'd make this one do the same (i.e. call it ADB_DATA_PATH).

mykmelez:
Other than those minor build script issues, this looks good. The only other change I would make is to `git rm` the DLLs that we don't need anymore now that we're rolling our own:
- android-tools/include/AdbWinApi.dll
- android-tools/include/AdbWinUsbApi.dll

Also, what about zlib1.dll; is that needed?
- android-tools/zlib/zlib1.dll

Makefile
bkase: The reason that I piped STDERR to /dev/null (and the true command) was to make sure that the Makefile worked on Mac and Linux (which is definitely hacky).

It felt weird doing a Windows check in this part of the Makefile, but it also seemed like it would be inappropriate to copy the AdbWin\* binaries to the right spot in _addon/data_ in any Makefile other than this one. What do you think is the best way to handle this?

bkase: I'll put in a `test -e` for the dlls for now-- I think that might be a pretty good solution

mykmelez: In the platform-specific defines, instead of declaring ADB_OUT_DIR, specify the files to copy, i.e. for Windows:

```
ADB_LIBS = \
  android-tools/win-out/libadb$(LIB_SUFFIX) \
  $(ADB_DRIVERS_DIR)api/objfre_wxp_x86/i386/AdbWinApi$(LIB_SUFFIX) \
  $(ADB_DRIVERS_DIR)winusb/objfre_wxp_x86/i386/AdbWinUsbApi$(LIB_SUFFIX)
```

And for Mac/Linux:

```
ADB_LIBS = android-tools/adb-bin/libadb$(LIB_SUFFIX)
```

Then you can copy those files to DATA_ADB_PATH without having to write platform-specific code in the recipe:

```
make -C android-tools lib && \
make -C android-tools driver && \
cp $(ADB_LIBS) $(DATA_ADB_PATH);
```

bkase: ok cool, this is better than what I did. I'll switch to this.

Makefile
```diff
@@ -75,6 +75,11 @@ ifeq (win32, $(B2G_PLATFORM))
   LIB_SUFFIX = .dll
 
   ADB_OUT_DIR = android-tools/win-out/
+  ADB_DRIVERS_DIR = android-tools/adb-win-api/
+  ADB_LIBS = \
+    android-tools/win-out/libadb$(LIB_SUFFIX) \
+    $(ADB_DRIVERS_DIR)api/objfre_wxp_x86/i386/AdbWinApi$(LIB_SUFFIX) \
+    $(ADB_DRIVERS_DIR)winusb/objfre_wxp_x86/i386/AdbWinUsbApi$(LIB_SUFFIX)
```
mykmelez, line 82: Two final nits:
- remove trailing slashes from the values of the directory variables, so paths containing them are easier to read;
- use $(ADB_OUT_DIR) in the libadb path (here and in the Mac/Linux declarations).

I'll make these changes when resolving conflicts with master in the process of merging the changeset!

mykmelez:
In the process of merging this, I also fixed some build bustage when building libadb on Windows (4a42d38).

PR #747, opened 2013-07-26
Created js-message bridge between C and JS

bkase:
This changeset adds a generic communication bridge from C to JS.

Why make a communication bridge?
Each time we need to add a C to JS communication we have to create a new function pointer, initialize it properly in JS (which means declaring the types for the function pointer), modify or create a new function in the native code that takes the function pointer as a parameter, and then assign the function pointer to some variable that is accessible in the scope that it will be called.
This bridge makes it so that we reuse the same function call to JS and we can switch on a channel name and extract arbitrary data out of the arguments.

This shouldn't actually be merged-- it's just an example of this function being called during ADB initialization. @mykmelez suggested that I create a pull-request just for this bridge before using it everywhere.

C communicates to JS with `MSG("channel", [type1, value1, ...])`
JS uses `JsMessage.prototype.unravel(type1, [type2, ...])` to extract data.

See https://github.com/bkase/r2d2b2g/compare/js-message?expand=1#L4R185 for an example of `MSG`
The `MSG` macro creates an anonymous structure for the given types and populates it with the proper values before passing it as a parameter to the generic `void * js_msg(void *)` function.
The macro is complex but it makes it really easy to pass a struct to JS without having to construct a struct manually in C. If JS gets an actual struct, we can use `ctypes.cast` and not have to worry about pointer offsets and sizes of the types passed in. `MSG` is implemented in _js-message.h_.

See https://github.com/bkase/r2d2b2g/compare/js-message?expand=1#L0R35 for an example of `JsMessage`'s `unravel`. This method takes the types of each of the elements (which we know based on the channel name) and returns an N-tuple of the data extracted from an anonymous struct constructed from the type information. `JsMessage` is implemented in _adb/js-message.js_

If this lands, I can clean up the old native code in a few places as well (since we can share one function pointer now).

@ochameau Can you look at this?

### DIFF
--- a/addon/lib/adb/adb-server-thread.js
+++ b/addon/lib/adb/adb-server-thread.js
@@ -10,11 +10,16 @@ const INSTANTIATOR_URL = URL_PREFIX + "ctypes-instantiator.js";
 const EVENTED_CHROME_WORKER_URL = URL_PREFIX + "evented-chrome-worker.js";
 const CONSOLE_URL = URL_PREFIX + "worker-console.js";
 const ADB_TYPES = URL_PREFIX + "adb-types.js";
+const JS_MESSAGE = URL_PREFIX + "js-message.js";
 
 const WORKER_URL_IO_THREAD_SPAWNER = URL_PREFIX + "adb-io-thread-spawner.js";
 const WORKER_URL_DEVICE_POLL = URL_PREFIX + "adb-device-poll-thread.js";
 
-importScripts(INSTANTIATOR_URL, EVENTED_CHROME_WORKER_URL, CONSOLE_URL, ADB_TYPES);
+importScripts(INSTANTIATOR_URL,
+              EVENTED_CHROME_WORKER_URL,
+              CONSOLE_URL,
+              ADB_TYPES,
+              JS_MESSAGE);
 
 const worker = new EventedChromeWorker(null);
 const console = new Console(worker);
@@ -24,6 +29,16 @@ let libadb = null;
 let libPath_;
 let restartMeFn = function restart_me() {
   worker.emitAndForget("restart-me", { });
+};
+let w;
+let jsMsgFn = function js_msg(channel, args) {
+  switch (channel.readString()) {
+    default:
+      console.log("Unknown message");
+  }
+
+  w = ctypes.uintptr_t(10);
+  return ctypes.cast(w, ctypes.uintptr_t.ptr);
 };
 
 worker.once("init", function({ libPath }) {
@@ -58,7 +73,14 @@ worker.once("init", function({ libPath }) {
                   args: [ CallbackType.ptr ]
                 }, libadb);
 
+  let install_js_msg =
+      I.declare({ name: "install_js_msg",
+                  returns: ctypes.void_t,
+                  args: [ JsMsgType.ptr ]
+                }, libadb);
+
   install_thread_locals(CallbackType.ptr(restartMeFn));
+  install_js_msg(JsMsgType.ptr(jsMsgFn));
 });
 
 worker.once("start", function({ port, log_path }) {
--- a/addon/lib/adb/js-message.js
+++ b/addon/lib/adb/js-message.js
@@ -1,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+ * JsMessage
+ *
+ * Takes a (void *)(struct *) from C and unpacks it properly
+ */
+
+'use strict';
+
+;(function(exports, module) {
+
+  if (module) {
+    const { Cu } = require("chrome");
+    Cu.import("resource://gre/modules/ctypes.jsm");
+  }
+
+  // the lifetime of the value returned from the JsMessage if it is allocated in
+  //   JS lasts until MSG is called again (this will prevent leaks as much as
+  //   possible)
+  let w;
+  exports.JsMessage = {
+    pack: function pack(val, type) {
+      if (type(0) === Number(0)) {
+        w = ctypes.intptr_t(val);
+        return ctypes.cast(w, ctypes.intptr_t.ptr);
+      } else if (type(0) === String(0)) {
+        w = ctypes.char.array()(val);
+        return w;
+      } else {
+        w = ctypes.intptr_t(-404);
+        return ctypes.cast(w, ctypes.intptr_t.ptr);
+      }
+    },
+
+    unpack: function unpack(struct_ptr /*, types */) {
+      let types = Array.slice(arguments, 1);
+      let struct_body = types.map(function (type, i) {
+        let o = {};
+        o["s_" + i] = type;
+        return o;
+      });
+
+      const struct_type =
+        new ctypes.StructType("anon", struct_body);
+
+      let as_struct_type = ctypes.cast(struct_ptr, struct_type.ptr);
+      return types.map(function(unused, i) as_struct_type.contents["s_" + i]);
+    }
+  };
+
+}).apply(null,
+  typeof module !== 'undefined' ?
+       [exports, module] : [this]);
+
--- a/android-tools/adb-bin/js_message.h
+++ b/android-tools/adb-bin/js_message.h
@@ -1,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef JS_MESSAGE_H
+#define JS_MESSAGE_H
+
+#define MSG(save, channel, instance) do {\
+    *save = js_msg(channel, (void *)&instance);\
+  } while(0)
+
+#endif
+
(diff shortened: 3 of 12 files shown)


bkase:
If it helps: 

``` c
  MSG(&res, "a channel", int, 10, char *, "hello");
```

Turns into

``` c
do { 
  struct _anon_0 { int g;char * h; }; 
  struct _anon_0 _anon_0_instance; 
  do { 
     struct _anon_0 _tmp = _anon_0_instance; 
     _tmp.g = 10; _tmp.h = "hello"; 
     _anon_0_instance = _tmp;
  } while(0); 
  *&res = js_msg("a channel", (void *)&_anon_0_instance); 
} while(0);
```

(whitespace modified so that code is more readable)

bkase:
Visual Studios' pre-processor has different semantics from gcc/clang. Working on a fix.

android-tools/adb-bin/js_message.h
```diff
@@ -0,0 +1,85 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef JS_MESSAGE_H
+#define JS_MESSAGE_H
+
+// Utils
+#define ARGIFY(x) x,
+#define CONCAT1(x, y) x##y
+#define CONCAT(x, y) CONCAT1(x, y)
```
ochameau: Hum aren't CONCAT and CONCAT1 strictly equivalent?

ochameau:
I like the idea of simplifying calling JS code from C, but I'm not sure the current implementation is efficient.
Isn't it costly to define a new struct each time we want to send a new message?

I think a big part of the complexity of calling JS from C is to pass a callback to C each time we want to be called,
but I'm not convinced that struct type definition is that painfull on C side, it is daily C job...

What do you think about just registering a global channel function, but still do manual struct definition on C side?
That would avoid loosing hairs on figuring out subtle compiler differences and prevent having to maintain magical macros.

Something like:

```
struct my_msg {
  int a;
  char *b;
}

my_msg m;
m.a = 10;
m.b = "foo";
MSG(&res, "a channel", msg);
```

```
let o = new JsMessage(args).unravel({a:ctypes.int, b:ctypes.float64_t});
o.a, o.b
```

android-tools/adb-bin/js_message.h
bkase: It turns out if you do:
`#define CONCAT(x, y) x##y` the `x##y` won't first evaluate the parameters before concatenating. But it will if you do that CONCAT1 trick.
So `CONCAT1(ARGIFY(a), b)` is `ARGIFY(a)b` while `CONCAT(ARGIFY(a), b)` is `a,b`

bkase:
You're right, that sounds good -- it's definitely not worth maintaining the different magic macro implementations. I'll change it to something closer to what you suggested.

> Isn't it costly to define a new struct each time we want to send a new message?

I was under the impression that it was actually free to do this, so I did a small example to make sure:
C:

``` c
volatile int x;
volatile int y;
...
int main() {
  void * res;
  MSG(&res, "a channel", int, 10, char *, "hello");
  x = (int)(res);
  MSG(&res, "a channel", int, 10, char *, "hello");
  y = (int)(res);
  return x + y;
}
```

After `clang -O1 test.c && objdump -d a.out`

``` asm
0000000100000ee0 <_main>:
   100000ee0:   55                      push   %rbp
   100000ee1:   48 89 e5                mov    %rsp,%rbp
   100000ee4:   53                      push   %rbx
   100000ee5:   48 83 ec 28             sub    $0x28,%rsp
   100000ee9:   48 8d 1d 47 00 00 00    lea    0x47(%rip),%rbx
   100000ef0:   48 89 5d f0             mov    %rbx,-0x10(%rbp)
   100000ef4:   c7 45 e8 0a 00 00 00    movl   $0xa,-0x18(%rbp)
   100000efb:   48 8d 75 e8             lea    -0x18(%rbp),%rsi
   100000eff:   e8 bc ff ff ff          callq  100000ec0 <_js_msg>
   100000f04:   89 05 f6 00 00 00       mov    %eax,0xf6(%rip)       
   100000f0a:   48 89 5d e0             mov    %rbx,-0x20(%rbp)
   100000f0e:   c7 45 d8 0a 00 00 00    movl   $0xa,-0x28(%rbp)
   100000f15:   48 8d 75 d8             lea    -0x28(%rbp),%rsi
   100000f19:   e8 a2 ff ff ff          callq  100000ec0 <_js_msg>
   100000f1e:   89 05 e0 00 00 00       mov    %eax,0xe0(%rip)        
   100000f24:   8b 05 d6 00 00 00       mov    0xd6(%rip),%eax        
   100000f2a:   03 05 d4 00 00 00       add    0xd4(%rip),%eax
   100000f30:   48 83 c4 28             add    $0x28,%rsp
   100000f34:   5b                      pop    %r
```

So it's only 4-5 instructions with `-O1` to prepare the parameters and call `js_msg`, even though it redefines a new struct every time (and all that other magic).

bkase:
This works on Windows if you comment out the debug statements (because it tries to print before initializing the log file)

bkase:
I forgot to change the JS behavior from `[a, b, c]` to `{ a: _, b: _, c: _ }`, I'll do that now

bkase:
`js-ctypes` needs to know the order of the parameters in the struct because the offsets of the elements change depending on the size.
So something like `new JsMessage(args).unravel({a:ctypes.int, b:ctypes.float64_t})` would actually need to be `new JsMessage(args).unravel({a:ctypes.int}, {b:ctypes.float64_t})`. I prefer omitting the names because of this and keeping the old API. Does that sound good?

addon/lib/adb/js-message.js
```diff
@@ -0,0 +1,52 @@
...
+      return a.map(function(x) [x, i++]);
+    },
+
+    unravel: function unravel(/* types */) {
+      let types = Array.prototype.slice.call(arguments);
+      let struct_body = this._zipWithIndex(types).map(function ([type, i]) {
+        let o = {};
+        o["s_" + i] = type;
+        return o;
+      });
```
ochameau, line 44: `Array.map` already exposes `i` to the callback, you should be able to drop zipWithIndex by doing:
`types.map(function (type, i) { ... });`

addon/lib/adb/js-message.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  JsMessage.prototype = {
+    _zipWithIndex: function zipWithIndex(a) {
+      let i = 0;
+      return a.map(function(x) [x, i++]);
+    },
+
+    unravel: function unravel(/* types */) {
+      let types = Array.prototype.slice.call(arguments);
```
ochameau: nit: `Array.slice(arguments)` should do the same job. (I think that's a mozilla specific thing)

addon/lib/adb/adb-server-thread.js
```diff
@@ -25,6 +30,25 @@ let libPath_;
 let restartMeFn = function restart_me() {
   worker.emitAndForget("restart-me", { });
 };
+let w;
+let jsMsgFn = function js_msg(channel, args) {
+  console.log("*** Channel: " + channel.readString());
+  switch (channel.readString()) {
+    case "test1":
+      let [x, y, z] = new JsMessage(args).unravel(ctypes.int, ctypes.char.ptr, ctypes.int);
```
ochameau: What about replacing JsMessage and unravel by just a simple function?
I don't see any usecase where we would need to just call JsMessage(args) first and then later call unravel,
nor call unravel multiple times for the same JsMessage with/without same arguments.
But I can imagine doing the opposite: First define a struct, and then cast multiple args sets matching this struct.

A simple function would be something like:
`let [x, y, z] = JsMessage(args, [ctypes.int, ctypes.char.ptr, ctypes.int]);`
or
`let [x, y, z] = JsMessage(args, ctypes.int, ctypes.char.ptr, ctypes.int);`

Or if we want to convert multiple args matching the same struct:

```
let test1Struct = JsMessage(ctypes.int, ctypes.char.ptr, ctypes.int);
let [x, y, z] = test1Struct.unravel(args);
```

ochameau:
With some polish on the JS side and unit test instead of hardcoded tests in main codepath,
I think we would be able to land that patch and let you start converting existing code to use that new feature in followups.

bkase:
@ochameau I went with `let [x, y, z] = JsMessage.unravel(args, ctypes.int, ctypes.char.ptr, ctypes.int)` because it seemed like it would be nice to deal with casting the primitive values in _JsMessage_ as well. See the new _JsMessage.pack_ method. How is that?

ochameau:
Sounds good, then what do you think about calling them pack/unpack?

addon/lib/adb/js-message.js
```diff
@@ -0,0 +1,57 @@
...
+        w = ctypes.char.array()(val);
+        return w;
+      } else {
+        w = ctypes.intptr_t(-404);
+        return ctypes.cast(w, ctypes.intptr_t.ptr);
+      }
+    },
+
+    unravel: function unravel(struct_ptr /*, types */) {
+      let types = Array.slice(arguments).slice(1);
```
ochameau: nit: I think `Array.slice(arguments, 1)` would also work.

ochameau:
I have to go now, but I think you can go ahead and land this patch.

bkase:
Thanks! I think @myk should look at the file structure the unit tests first since we may need to move stuff around. See fb9c82b

mykmelez:
> Thanks! I think @myk should look at the file structure the unit tests first since we may need to move stuff around. See fb9c82b

Looks good to me. My only nit is the one about the names of the Make targets.

PR #751, opened 2013-07-29
Fix app addition/removal on the dashboard. Fixes #726

### LINKED ISSUE #726: added app doesn't appear on Dashboard until reload
After adding a packaged app to the Dashboard, the Simulator starts and installs the app, but the app doesn't appear on the Dashboard itself until I reload it.

The problem doesn't exist in 4.0. @potch perhaps this is a regression from updating individual entries?

ochameau:
The escaping of `\` on app ids (that are manifest paths) allows to make work the app removal on windows.
And the rest of the patch fixes app registration on the dashboard.

### DIFF
--- a/addon/data/content/js/applist.js
+++ b/addon/data/content/js/applist.js
@@ -16,6 +16,10 @@ function timedelta(d1) {
     return ~~(d / 365) + ' years ago';
 }
 
+function escapeSelector(str) {
+  return str.replace(/[!"#$%&'()*+,.\/:;<=>?@\[\\\]^`{|}~]/g, '\\$&');
+}
+
 var AppList = (function() {
 
     window.appTemplate = new nunjucks.Template($('#app-template').html());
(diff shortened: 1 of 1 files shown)


addon/data/content/js/applist.js
```diff
@@ -37,26 +37,36 @@ var AppList = (function() {
...
         }
-        render(id);
     }
 
     function render(id) {
         if (id) {
             // re-render a single app in place.
-            var row = $('[data-id="'+id+'"]');
+            // On windows with need to escape backslashes used in paths
+            var row = $('[data-id="'+id.replace(/\\/g,"\\\\")+'"]');
```
mykmelez: Hrm, we've run into this before. :-/

The [jQuery Selectors doc](http://api.jquery.com/category/selectors/) says, "to use any of the meta-characters ( such as !"#$%&'()*+,./:;<=>?@[]^`{|}~ ) as a literal part of a name, it must be escaped with two backslashes", and it points to a lengthy blog post detailing [CSS character escape sequences](http://mathiasbynens.be/notes/css-escapes). Meanwhile, several people have written functions to do general escaping of such strings:
- http://alexandregiannini.blogspot.com/2011/05/escaping-strings-for-jquery-selectors.html
- http://totaldev.com/content/escaping-characters-get-valid-jquery-id
- http://stackoverflow.com/questions/739695/jquery-selector-value-escaping

Let's add a function like those to our script and use it to do the escaping here.

cc: @potch

ochameau:
I addressed both comments.

PR #752, opened 2013-07-29
Simplified mutex-related code in Windows shutdown routine

bkase:
This changeset applies @ochameau's suggestions from #740. I didn't create a separate boolean variable because I thought the code would be simpler without it, but let me know if I should add that in.

### DIFF
--- a/android-tools/adb-bin/usb_windows.cpp
+++ b/android-tools/adb-bin/usb_windows.cpp
@@ -196,7 +196,14 @@ int register_new_device(usb_handle* handle) {
 
 
 static int should_kill = 0;
+extern int is_io_pump_on;
 ADB_COND_DEFINE(should_kill_cond);
+static void inc_should_kill() {
+  adb_mutex_lock(&should_kill_lock);
+  D("Set should_kill to %d\n", should_kill+1);
+  should_kill++;
+  adb_mutex_unlock(&should_kill_lock);
+}
 static void set_should_kill(int val) {
   adb_mutex_lock(&should_kill_lock);
   D("Set should_kill to %d\n", val);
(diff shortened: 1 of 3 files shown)


android-tools/adb-bin/usb_windows.cpp
```diff
@@ -196,32 +196,22 @@ int register_new_device(usb_handle* handle) {
...
-void notify_should_kill(int k, char who) {
-  if (k < 0) {
-    k = get_should_kill();
-    if (k == 0) {
-      return;
-    }
+void notify_should_kill() {
+  if (!should_kill) {
+    return;
   }
```
ochameau, line 217: We are checking `should_kill` twice unecessary, we should either check it here or before calling `notify_should_kill`.

android-tools/adb-bin/usb_windows.cpp
```diff
@@ -230,10 +220,9 @@ void notify_should_kill(int k, char who) {
 // signals the device_loop and device_output_thread
 void should_kill_threads() {
   ADB_COND(should_kill_cond);
-  set_should_kill(1);
+  inc_should_kill();
```
ochameau: Shouldn't we set it to `1` in order to be able to instanciate/destroy these thread mutliple times?

android-tools/adb-bin/usb_windows.cpp
```diff
@@ -196,32 +196,22 @@ int register_new_device(usb_handle* handle) {
...
+void notify_should_kill() {
+  if (!should_kill) {
+    return;
   }
 
-  set_should_kill(k+1);
-  int is_io_pump_on = get_io_pump_status();
-  if (k >= 2 + is_io_pump_on) {
+  inc_should_kill();
+  if (should_kill >= 2 + is_io_pump_on) {
```
ochameau: nit: what do you think about always calling adb_cond_broadcast and let should_kill_threads decide wheter the cleanup is done or not?

ochameau:
This patch is good to go with comments being addressed

android-tools/adb-bin/usb_windows.cpp
bkase: Good point

android-tools/adb-bin/usb_windows.cpp
bkase, line 217: We need the check in `notify_should_kill` in case it is called when the device input thread is destroyed but we're not actually ready to shutdown (in the case of the phone being unplugged for example). But then we also need to check for `should_kill` here: https://github.com/mozilla/r2d2b2g/pull/752/files#L2R240 because we need to `return null` if `should_kill` is true. I'll make `notify_should_kill` return `should_kill`; I think that's the best solution.

android-tools/adb-bin/usb_windows.cpp
bkase: sounds good

Issue #748, opened 2013-07-27
[Receipts] fix manifest_url API parameter for packaged apps.


nickdesaulniers:
See [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=867282#c1).

bkase:
`make test TEST=test-receipts` hangs right now. Is it because of this bug?

nickdesaulniers:
could be, haven't investigated yet

nickdesaulniers:
Just tested this today, nothing broken.  Might not be live yet in production.

nickdesaulniers:
I don't think this is an issue.

PR #753, opened 2013-07-30
The simulator requires old adb module if needed

bkase:
The Simulator module now dynamically requires adb/adb (the new _adb.js_) or adb/adb-fallback (the _adb.js_ before my huge pull-request).
Now we can continue to break compatibility with adb's socket IPC interface without worrying about the case where _adb_ is already running on a user's computer.

All tests except for _test-receipts_ passes (_test-reciepts_ was already broken see #748) and everything seems to work when _adb_ is either running or not running before starting up the simulator 

Thanks @nickdesaulniers for noticing that the simulator wasn't detecting devices when _adb_ was running in the latest preview build.


addon/lib/adb/adb.js.yeah
```diff
@@ -0,0 +1,355 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mykmelez: Umm, yeah?

addon/lib/simulator.js
```diff
@@ -32,6 +37,10 @@ const Debugger = require("debugger");
 // The b2gremote debugger port.
 const DEBUGGER_PORT = 6000;
 
+const isAdbRunningPromise = require("adb/adb-running-checker").check();
+
+let ADB = { val: null };
```
mykmelez: Why must _ADB_ be an object with a _val_ property? We should be to assign the return value of a module `require()` statement to it directly, and the object doesn't seem to store any other value.

addon/lib/simulator.js
```diff
@@ -32,6 +37,10 @@ const Debugger = require("debugger");
 // The b2gremote debugger port.
 const DEBUGGER_PORT = 6000;
 
+const isAdbRunningPromise = require("adb/adb-running-checker").check();
```
mykmelez: This line causes the module to check for ADB when it loads, whereas we want to delay doing so until the Dashboard actually opens (i.e. the worker is set). So we shouldn't do this check until then.

addon/lib/simulator.js
```diff
@@ -52,6 +61,10 @@ let deviceConnected, adbReady, debuggerReady;
 let gCurrentConnection, gCurrentToolbox, gCurrentToolboxManifestURL;
 let gRunningApps = [];
 
+// HACK: For some reason the isAdbRunning promise still resolves even after
+//       the module unloads. If ADB is initialized, but not closed FF gets
+//       in a bad state. To prevent this, we set isDead to true in unload
+let isDead = false;
```
mykmelez: I suspect a reference cycle involving this module and the Promise module, although it isn't exactly clear how. I suppose you could try declaring _Promise_ as a variable and deleting it when this module unloads. But if that doesn't work, then there should still be a more elegant way to ignore the success callback invocation after the module unloads.

addon/lib/adb/adb.js.yeah
bkase: oops, sorry

bkase:
I can't reproduce the weird unloading bug anymore, so one of the other changes must have fixed it. I just removed the isDead and the comments.

addon/lib/simulator.js
```diff
@@ -113,9 +119,15 @@ let simulator = module.exports = {
         worker = null;
       });
 
-      if (!ADB.ready) {
-        ADB.start();
-      }
+      const isAdbRunningPromise = require("adb/adb-running-checker").check();
+      isAdbRunningPromise.then(function onSuccess(isAdbRunning) {
```
mykmelez: Nit: unless you have to reference a function from inside itself (f.e. to remove an event listener from inside the listener), there's no longer any good reason to name a function, as the JS debugger has gotten really good at identifying anonymous functions. So this callback need not have the name _onSuccess_.

Also, nit: you shouldn't need to declare the _isAdbRunningPromise_ const here; just call _then_ directly against the function call that returns a promise, i.e.:

```
      require("adb/adb-running-checker").check().then(function(isAdbRunning) {
```

mykmelez: Erm, to be precise: there's no longer a reason to name a _function expression_; there's still plenty of reason to name a function declaration, of course, since otherwise you wouldn't be able to reference it. :-)

PR #754, opened 2013-07-30
add new app to AppList if updated app is not presently in the list.

### LINKED ISSUE #726: added app doesn't appear on Dashboard until reload

potch:
Finally chased down the bug that causes newly added apps to not render. Fixes #726.


mykmelez:
Closing as the issue was fixed by another pull.

PR #755, opened 2013-07-30
 Device tracker calls callback instead of sockets

bkase:
This pull request rips out the device tracker implementation (based on socket communication) in native code and replaces it with one that calls JS callback (sends a js-message) every time some device state changes.

### DIFF
--- a/addon/lib/adb/adb-device-tracker.js
+++ b/addon/lib/adb/adb-device-tracker.js
@@ -21,98 +21,69 @@ function debug() {
   console.debug.apply(console, ["ADB: "].concat(Array.prototype.slice.call(arguments, 0)));
 }
 
-let waitForFirst = true;
 let devices = { };
-let socket;
 let hasDevice = false;
+let listenId = null;
+let worker = null;
 module.exports = {
   get hasDevice() {
     return hasDevice;
   },
 
-  reset: function reset() {
-    waitForFirst = false;
-    devices = { };
-    socket.close();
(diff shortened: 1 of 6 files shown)


bkase:
@mykmelez This should be ready for review now (since #747 landed)

mykmelez:
Running `make test` on Mac is crashing for me:

https://crash-stats.mozilla.com/report/index/7e0c4cb9-6f5b-4d52-ac3d-179ee2130806

addon/lib/adb/adb-device-tracker.js
```diff
@@ -21,98 +21,69 @@ function debug() {
...
+  stop: function stop() {
+    if (listenId !== null) {
+      worker.freeListener("device-update", listenId);
+    }
+  },
+
+  reset: function reset() {
+    devices = { };
+    hasDevice = false;
+    listenId = null;
```
mykmelez, line 87: Shouldn't _reset_ also set _worker_ back to `null`? I don't see any code path that would lead to _worker_ being accessed after the tracker is reset, but someone could add such a path in the future, and it seems better for that to fail with a "TypeError: worker is null" exception than be able to access the previously-registered worker.

mykmelez:
Other than the crasher and that _reset_ issue, this patch looks great, very straightforward.

bkase:
The crash is fixed if this diff is rebased on top of #775. So we should probably wait to merge this until that lands.

PR #756, opened 2013-07-31
Start integrating with app manager

ochameau:
Work in progress...

It mainly requires review on the platform side part before being taken seriously here,
but feedback much appreciated for both patches on how we are going to integrate with the simulator addon.

Here, we register to a firefox JSM that will be queried by the app manager code to know if there is some simulator being installed, and if yes, launch/close a new simulator:
https://github.com/ochameau/mozilla-central/commit/234071900839a5f1986e63cacbb2a3a8eead0554

Otherwise, on simulator side, I basically forked remote-simulator-client in order to only keep process code (and strip all DebuggerClient logic). I think it will be handy to land this code in the simulator addon to easily test the new firefox app manager. But later on, we should spin another addon, that do not ship the dashboard.

With this, we should be able to install multiple simulator addon, so there is some refinement to be defined, where we would have one new simulator addon id per Firefox OS big release/branch: 1.1, 1.2, ...


mykmelez:
Feedback+!

Overall this looks like a sensible refactoring and API for registering Simulators with the App Manager. I didn't do a code review, but here are a few general thoughts about the design:
- We should make the version string be configurable at build time, by preprocessing the script that specifies it or putting the value into a configuration file.
- We'll want to be able to create Simulator builds with non-release versions, i.e. nightly builds, preview builds, etc. I'm not sure what those version strings would look like; perhaps some combination of branch version and date (f.e. 1.1-20130807).
- Will we ever want to support component-specific versions (where the components in question are B2G, Gaia, and possibly also the Simulator itself)?
- We should convert SimulatorProcess to use promises instead of the event/callback model.
- The way we implement the remoteDebuggerPort property as a kind of memoizing getter whose value can nevertheless be changed by others has always seemed fairly complex and error-prone. It also looks like we don't use it anymore, since _launch_ in register-simulator.js solicits and sets the port itself.  So it seems like we should remove it from SimulatorProcess. Or, if we still need it sometimes, make _SimulatorProcess.run_ set it, either to the value of a _port_ parameter or by choosing the port itself; and don't expose a setter for changing it afterward.

ochameau:
> - We should make the version string be configurable at build time, by preprocessing the script that specifies it or putting the value into a configuration file.

What about reusing addon version, with something like `require("self").version`?
- We'll want to be able to create Simulator builds with non-release versions, i.e. nightly builds, preview builds, etc. I'm not sure what those version strings would look like; perhaps some combination of branch version and date (f.e. 1.1-20130807).

I tend to think we should put the device branch version in the adon name/id, so that we can have one addon for each branch (1.1, 1.2, ...). Then the addon version itself would be a combination of m-c+gaia+simulator revisions. 
May be a simple build date version like 20130807 would work with such pattern.
Do you think that the right way to go?

Otherwise, regarding device releases, I haven't heard about any kind of rc/beta/alpha releases for devices. I've only seen branches and there isn't even tags for official releases :/
For now it seems to be only nightly style for each branch v1.0.1, v1.1, v1.2.
The version string being used is: 1.2.0.0-prerelease.
- Will we ever want to support component-specific versions (where the components in question are B2G, Gaia, and possibly also the Simulator itself)?

I think we should keep the device/connection manager in firefox simple. Also, even if I heard multiple times about supporting disctinct branches between gecko and gaia, like running gaia master on gecko18. It has been recently officially described as "not supported". See "Do we need to support Gaia master + b2g18?" thread on dev-gaia/dev-b2g. 
Having said that, we may do something about this in future, but I'd prefer doing this complex components management in the addon itself and may be tweak the device manager to let the addon display more custom content. For example, one usecase that I find really interesting would be to have a simulator addon shipping gecko master and where you would like to plug a local clone of gaia master and hack on it. That would be soon be possible to build gaia via a firefox addon (gaia is soon going to have its own "cfx in js" addon, "gaia build system in js"). It will open sane ways to offer such feature and it clearly is what you are describing here. I think that at the end, we want to control the gaia version (local path) in some UI controlled by the simulator addon, not a UI in firefox itself.

ochameau:
See following bug for gecko part: https://bugzilla.mozilla.org/show_bug.cgi?id=902910

Mossop:
You probably want to expose something like nsIXULAppInfo for the simulator build. That can include potentially useful info like the changeset used to make the build and the versions of the main components. It may seem like overkill but I bet it'll be useful down the road, plus we can hopefully just generate a JSON blob with that info automatically.

ochameau:
@mossop you should definitely take a look at https://bugzilla.mozilla.org/show_bug.cgi?id=895360
Your comment makes also a lot of sense regarding that bug's task.
So with your suggestion in mind we would more or less expose same data than this new device actor, but from the addon itself, i.e. without having to launch b2g runtime and connect to it via remote protocol.
I like that, as it would allow us to display various information controled by Firefox,
but I'm also wondering if at the end it isn't the addon that really knows how to describe itself with a simple string??
Let's say we release a firefox OS for kids or a simular matching a precise operator customization, we could easily put a "for kids" or "zte-open" in one app info property,
but wouldn't it better to just let the addon describe itself with custom word ordering?

Here I assume the simulator gives just a version like "1.1", but the more I think about this, the more I think it should give a string like "Firefox OS 1.2", "Firefox OS 1.2 zte-open version", "Firefox OS 1.4 for kids", "Firefox OS 1.1 with gaia 1.2", ...

Mossop:
Yes, replicating that data in a way that doesn't require starting the b2g runtime sounds perfect. I think that information will be useful for all sorts of things, like deciding if a certain debug capability exists in the simulator etc. But I agree that a more human friendly name is probably also useful. Maybe that is just the add-on name or something in addition to the more detailed info. Do Firefox OS devices have a name like this at all? Seems like that would be a useful thing to have for them too

mykmelez:
To recap all the comments here: it sounds like we want the Simulator to expose both:
- a single string that expresses succinctly a human-friendly name that helps humans distinguish multiple simulators in application interfaces (`Firefox OS 1.2`, `Firefox OS 1.2 zte-open version`, etc.);
- an API for accessing the precise changesets/versions of the components comprising a Simulator build (B2G, Gaia, Simulator addon wrapper, perhaps libadb, etc.).

The simple string seems like it could be simply the name of the addon, which would be useful in addon-centric interfaces like the Add-ons Manager as well.

> Do Firefox OS devices have a name like this at all? Seems like that would be a useful thing to have for them too

_Settings_ > _Device information_ has a _Software_ field with a value like `Boot2Gecko 1.1.0.0-prerelease` (from my Unagi), and its _More Information_ button then takes you to a screen with additional software (and hardware) fields. Here are the software fields and their values on my Unagi:
- _OS version_: `1.1.0.0-prerelease`
- _Platform version_: `18.1`
- _Build Identifier_: `20130812041203`
- _Update channel_: `unagi/1.1.0/beta`
- _Git commit info_: `2013-08-11 13:53:25 [revision ID]`

ochameau:
I crafted a python script (build/make-appinfo.py) to generate during build a static file (appinfo.json),
that is shipped in data's addon folder. Then we just forward this info to the simulator jsm.
Otherwise, I now send addon's name and version concatenation.
I also tried to address comments about SimulatorProcess  and remoteDebuggerPort.
Finally, I haven't made addon name configurable during build as it isn't really easy to do so and I'm not sure we will immediatly need such feature?

I think this patch is ready for review, I'll ask for review for the mozilla-central patch as well.

build/make-appinfo.py
```diff
@@ -0,0 +1,52 @@
...
+  "buildid": config.get("App", "BuildID"),
+  "id": config.get("App", "ID"),
+  "b2g_version": None
+}
+prefs_file = os.path.join(options.gecko_path,
+                          "defaults", "pref", "b2g.js")
+prefs = open(prefs_file, "r")
+for line in prefs:
+  if "b2g.version" in line:
+    appinfo["b2g_version"] = line.split('"')[3]
```
mykmelez, line 38: Hrm, I wish we had a better way to do this.

build/make-appinfo.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+# Then fetch some usefull data from gaia folder
```
mykmelez: Nit: usefull -> useful

addon/lib/simulator-process.js
```diff
@@ -0,0 +1,190 @@
...
+        emit(this, "stderr", data);
+      }).bind(this),
+
+      // on b2g instance exit, reset tracked process, remoteDebuggerPort and
+      // shuttingDown flag, then finally emit an exit event
+      done: (function(result) {
+        console.log(this.b2gFilename + " terminated with " + result.exitCode);
+        this.process = null;
+        emit(this, "exit", result.exitCode);
+      }).bind(this)
```
mykmelez, line 108: shuttingDown doesn't actually seem to get reset here, which suggests that it'll remain `true` forever after the first time the simulator process is killed via the _kill_ function.

addon/lib/simulator-process.js
```diff
@@ -0,0 +1,190 @@
...
+    if (this.process && !this.shuttingDown) {
+      let deferred = Promise.defer();
+      emit(this, "kill", null);
+      this.shuttingDown = true;
+      this.once("exit", (exitCode) => {
+        this.shuttingDown = true;
+        deferred.resolve(exitCode);
+      });
+      this.process.kill();
+      return deferred.promise;
```
mykmelez, line 125: It isn't entirely clear: is this.process.kill() actually asynchronous? Here you return a promise that gets resolved once("exit"), which happens on "done". But _run_ sometimes calls the same function to kill the process and then immediately starts up a new process.

If this.process.kill() is actually synchronous, then presumably exitCode is available by the time it returns, so it isn't necessary to wait for "exit" nor to return a promise here.

If, on the other hand, this is actually asynchronous, then _run_ should also wait for _exit_ after calling this.process.kill(); and it should probably reuse this function to kill the process.

addon/lib/simulator-process.js
```diff
@@ -0,0 +1,190 @@
...
+    } else {
+      return Promise.resolve(-1);
```
mykmelez: Hmm, I'm not sure this is the right exit code in all cases. It seems ok if `!this.process`, although it might be better to return the undefined value or NaN to make it clear that there was no exit code. But if `this.shuttingDown`, then I imagine that the caller will want the function to wait until the process finishes shutting down to resolve the promise, and it'll want to know the actual error code that was returned. So I would register an _exit_ handler in that case and resolve the promise in that handler.

addon/lib/register-simulator.js
```diff
@@ -0,0 +1,48 @@
...
+// so do not throw on addon startup when we miss it
+try {
+  Simulator = Cu.import("resource://gre/modules/devtools/Simulator.jsm").Simulator;
+} catch(e) {}
+
+if (Simulator) {
+  Simulator.register(Self.name + " " + Self.version, {
+    appinfo: appinfo,
+    launch: launch,
+    close: close
```
mykmelez: It doesn't look like _launch_ and _close_ do much; they seem like very basic wrappers around SimulatorProcess functions. Perhaps it would be simpler to just expose the SimulatorProcess constructor here and let consumers instantiate it and call its _run_ and _kill_ functions directly.

mykmelez:
> Otherwise, I now send addon's name and version concatenation.
> …
> Finally, I haven't made addon name configurable during build as it isn't really easy to do so and I'm not sure we will immediatly need such feature?

I think we're going to need it eventually, but we don't need it immediately, so this seems ok.

mykmelez:
> Hrm, I wish we had a better way to do this.

I suppose [Mozilla::Prefs::Simple](http://search.cpan.org/~rrwo/Mozilla-Prefs-Simple-0.01/lib/Mozilla/Prefs/Simple.pm) is that better way, although I'm not sure it's worth pulling in a Python module for this one use case. But perhaps it's worth referencing it in a comment, in case we run into parsing problems in the future.

build/make-appinfo.py
ochameau, line 38: I'd rather fix that on b2g side by exposing it through application.ini or something...
But so far, we aren't even using these appinfo on Firefox UI side, so I don't really want to put more effort until we find a real usecase for these data.

addon/lib/simulator-process.js
ochameau, line 108: Oops!

ochameau, line 108: I meant to move this `shuttingDown=false` from here to `kill` so that this flag is only set/read from this function.

ochameau:
I just pushed a bunch of necessary fixes for b2g/gaia master, 26 branch.
Some more fixes are needed, like removing the software home button.
I also addressed some review comments, but I need to test them once app manager stuff gets landed and b2g desktop build get shell.html. (this patch supports only b2g with shell.html and today's build still have shell.xul)

ochameau:
@mykmelez I addressed various comments, removed the dashboard and adb and started using very last gecko and gaia.
It is working very great with last required patch for firefox, that should land soon - https://bugzilla.mozilla.org/show_bug.cgi?id=912892

Should I push this branch to a new upstream branch?

Would you have some free time to spawn xpi builds out of this branch?

Makefile
```diff
@@ -45,21 +45,8 @@ B2G_TYPE ?= specific
...
-# binaries from the FTP server.
-LIBADB_LOCATION ?= remote
-
-# This variable determines whether or not ADB is built with crypto support
-# The values are "off" (crypto is disabled) or "dynamic" (crypto is enabled
-# and linked dynamically) or "static" (crypto is enabled and linked statically 
-# on linux)
-ADB_AUTH ?= off
-export ADB_AUTH
+# Use the current last know revision that sucessfully builds on windows
```
mykmelez: Nits: know -> known, sucessfully -> successfully, windows -> Windows.

Makefile
```diff
@@ -203,37 +145,12 @@ profile:
 # an alias to that target now.
 prosthesis: profile
 
+appinfo: profile b2g
+	python build/make-appinfo.py --gecko addon/data/$(B2G_PLATFORM)/b2g/ --gaia gaia/ --data addon/data/
```
mykmelez: This fails on Mac, where the B2G installation directory is B2G.app/Contents/MacOS. Instead of hardcoding "b2g" in the path here, define a B2G_BIN_DIR variable in the "platform-specific defines" section at the top of the file, set it to the platform-specific value, and reference it here.

Makefile
```diff
@@ -249,11 +166,10 @@ test:
 help:
 	@echo 'Targets:'
 	@echo "  build: [default] build, download, install everything;\n"\
-	"         combines the profile, b2g, and adb make targets"
+	"         combines the profile and b2g make targets"
```
mykmelez: Nit: _build_ also does _appinfo_.

Makefile
```diff
@@ -249,11 +166,10 @@ test:
...
 	@echo '  clean: remove files created during the build process'
 	@echo '  profile: make the Gaia profile and its prosthesis addon'
 	@echo '  b2g: download and install B2G'
```
mykmelez, line 179: Nit: the _appinfo_ target should be documented here between the _profile_ and _b2g_ targets.

mykmelez:
It looks like tests are failing because you didn't remove some tests:

```
09-13 15:29 > make test
cd addon-sdk && . bin/activate && cd ../addon && cfx test --verbose --templatedir template/   
Welcome to the Add-on SDK. Run 'cfx docs' for assistance.
Using binary at '/Applications/Firefox.app/Contents/MacOS/firefox-bin'.
Using profile at '/var/folders/lp/8t_7y24119720_hjp5wc_4cr0000gn/T/tmpvstnkL.mozrunner'.
Running tests on Firefox 23.0.1/Gecko 23.0.1 ({ec8030f7-c20a-464f-9b0e-13a3a9e97384}) under darwin/x86.
Error: Error: Module: r2d2b2g/tests/test-evented-chrome-worker has no authority to load: adb/evented-chrome-worker 
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/tests/test-evented-chrome-worker.js", line 5, in 
    const EventedChromeWorker = require("adb/evented-chrome-worker").EventedChromeWorker;
  File "resource://gre/modules/commonjs/sdk/loader/cuddlefish.js", line 113, in resolve
    + id, requirer);
0 of 1 tests passed.
TypeError: tests.testRuns is undefined
resource://gre/modules/commonjs/sdk/test/runner.js 56
    callback.apply(null, args);
  File "resource://gre/modules/commonjs/sdk/test/runner.js", line 49, in 
    onDone: onDone
  File "resource://gre/modules/commonjs/sdk/test/harness.js", line 615, in runTests
    onDone({passed: 0, failed: 1});
  File "resource://gre/modules/commonjs/sdk/test/runner.js", line 30, in onDone
    printFailedTests(tests, stdout.write);
  File "resource://gre/modules/commonjs/sdk/test/runner.js", line 56, in printFailedTests
    let singleIteration = tests.testRuns.length == 1;
info: r2d2b2g: [JavaScript Error: "r2d2b2g: An exception occurred.
TypeError: tests.testRuns is undefined
resource://gre/modules/commonjs/sdk/test/runner.js 56
    callback.apply(null, args);
  File "resource://gre/modules/commonjs/sdk/test/runner.js", line 49, in 
    onDone: onDone
  File "resource://gre/modules/commonjs/sdk/test/harness.js", line 615, in runTests
    onDone({passed: 0, failed: 1});
  File "resource://gre/modules/commonjs/sdk/test/runner.js", line 30, in onDone
    printFailedTests(tests, stdout.write);
  File "resource://gre/modules/commonjs/sdk/test/runner.js", line 56, in printFailedTests
    let singleIteration = tests.testRuns.length == 1;
"]
```

mykmelez:
When I start up the Simulator from the App Manager, the Simulator window displays only black, and I get this message in Firefox's console:

> - Call to xpconnect wrapped JSObject produced this error:  *
>   [Exception... "'[JavaScript Error: "Argument 1 of Node.removeChild is not an object." {file: "chrome://browser/content/shell.js" line: 244}]' when calling method: [nsIDOMEventListener::handleEvent]"  nsresult: "0x80570021 (NS_ERROR_XPC_JAVASCRIPT_ERROR_WITH_DETAILS)"  location: "native frame :: <unknown filename> :: <TOP_LEVEL> :: line 0"  data: yes]

mykmelez:
I see other errors too:

> console.log: r2d2b2g: ************************************************************
> - Call to xpconnect wrapped JSObject produced this error:  *
>   [Exception... "'[JavaScript Error: "Argument 1 of Node.removeChild is not an object." {file: "chrome://browser/content/shell.js" line: 244}]' when calling method: [nsIDOMEventListener::handleEvent]"  nsresult: "0x80570021 (NS_ERROR_XPC_JAVASCRIPT_ERROR_WITH_DETAILS)"  location: "native frame :: <unknown filename> :: <TOP_LEVEL> :: line 0"  data: yes]
>   ************************************************************
>   reference to undefined property exn.stackconsole.error: r2d2b2g: 
>   JavaScript error: chrome://browser/content/shell.js, line 293: this.contentBrowser is null
>   console.log: r2d2b2g: ************************************************************
> - Call to xpconnect wrapped JSObject produced this error:  *
>   [Exception... "'[JavaScript Error: "this._slowSQLStartup.mainThread is not an object" {file: "file:///Users/myk/Profiles/generic/extensions/r2d2b2g@mozilla.org//resources/r2d2b2g/data/mac64/B2G.app/Contents/MacOS/components/TelemetryPing.js" line: 554}]' when calling method: [nsITelemetryPing::observe]"  nsresult: "0x80570021 (NS_ERROR_XPC_JAVASCRIPT_ERROR_WITH_DETAILS)"  location: "native frame :: <unknown filename> :: . <TOP_LEVEL> :: line 0"  data: yes]
>   ************************************************************

mykmelez:
The addon should sprout a different ID and version.

mykmelez:
@ochameau I fixed as many issues as I could on my local copy of your branch, then I pushed it to my fork:

> https://github.com/mykmelez/r2d2b2g/tree/ochameau-integrate-app-manager

And built test builds:
- Windows: https://ftp.mozilla.org/pub/mozilla.org/labs/r2d2b2g/r2d2b2g-fxos1.2-test-windows.xpi
- Mac: https://ftp.mozilla.org/pub/mozilla.org/labs/r2d2b2g/r2d2b2g-fxos1.2-test-mac.xpi
- Linux: https://ftp.mozilla.org/pub/mozilla.org/labs/r2d2b2g/r2d2b2g-fxos1.2-test-linux.xpi

They're very large, but I haven't investigated the cause of that yet.

I used a variant of the build process I've been using to generate Simulator Classic builds:

> https://github.com/mozilla/r2d2b2g/wiki/Build-Process

And the Simulator appears to integrate into the App Manager and mostly work, except for that black screen on startup issue, which seems Mac-specific, since I don't see it on Windows or Linux. But on Windows/Linux, while I can install an app, I don't seem to be able to start it nor debug it.

cc: @paulrouget

mykmelez:
> And the Simulator appears to integrate into the App Manager and mostly work, except for that black screen on startup issue, which seems Mac-specific, since I don't see it on Windows or Linux. But on Windows/Linux, while I can install an app, I don't seem to be able to start it nor debug it.

Erm, correction: I can start the app on Windows, although I still can't debug it.

mykmelez:
I tested again with the latest nightly build of Firefox, the latest nightly build of B2G, and the tip of Gaia's master branch, and things are even more broken. Now both Mac and Windows builds of B2G stall at a black screen on startup; and Firefox no longer connects to the Simulator on any platform, as its connection attempt times out.

Probably unrelated, but I also see a duplicate invitation to connect to a port on localhost:

screen shot 2013-09-16 at 09 37 35

mykmelez:
My branch for this pull request has the aforementioned updates, so you can merge it into your branch to see this behavior for yourself:

https://github.com/mykmelez/r2d2b2g/tree/ochameau-integrate-app-manager

ochameau:
Oh sorry myk, I should have immediatly commented I was working no these issues. I took your branch with the fixes you made on friday, addressed other review comments and figured out the daily breakages of b2g desktop...

I made one patch against m-c, but I'm not sure it is going to land today, so I added some workaround in prosthesis. But as soon as it lands and is being built in nightly, we should update the zips being downloaded and remove it.
https://bugzilla.mozilla.org/show_bug.cgi?id=916237

mykmelez:
I'm still seeing the black screen on startup on Windows and Mac, although I was able to start the Simulator and debug an app on Linux. I would push your fix for bug 916237, but all the trees are closed!

mykmelez:
```
    // Workaround until bug 917310 lands
    Cu.import("resource://gre/modules/WebappOSUtils.jsm").WebappOSUtils.getInstallPath = function(aApp) {
      return aApp.basePath + "/" + aApp.id;
    };
```

This fails, at least on Mac, because _Cu_ is not defined.

mykmelez:
```
+# Use the current last known revision that sucessfully builds on Windows.
+B2G_URL_BASE = http://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2013-09-16-04-02-05-mozilla-central/
```

We should use the https: variant of this URL.

mykmelez:
Even after fixing the _Cu_ issue, I still see a black screen on startup on Mac and I can't debug an app, although the App Manager claims to have connected to the Simulator. The following message appears in the window of the terminal in which I started Firefox:

```
console.log: r2d2b2g: ************************************************************
* Call to xpconnect wrapped JSObject produced this error:  *
[Exception... "'[JavaScript Error: "Argument 1 of Node.removeChild is not an object." {file: "chrome://browser/content/shell.js" line: 244}]' when calling method: [nsIDOMEventListener::handleEvent]"  nsresult: "0x80570021 (NS_ERROR_XPC_JAVASCRIPT_ERROR_WITH_DETAILS)"  location: "native frame :: <unknown filename> :: <TOP_LEVEL> :: line 0"  data: yes]
************************************************************

###!!! [Parent][AsyncChannel] Error: Channel error: cannot send/recv

TypeError: this.connection is undefined: UI.startSimulator/</<@chrome://browser/content/devtools/app-manager/connection-footer.js:140
@resource://gre/modules/commonjs/toolkit/loader.js -> resource:///modules/devtools/shared/event-emitter.js:63
EventEmitter_emit@resource://gre/modules/commonjs/toolkit/loader.js -> resource:///modules/devtools/shared/event-emitter.js:107
Connection.prototype._setStatus@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/devtools/client/connection-manager.js:239
Connection.prototype._onConnected@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/devtools/client/connection-manager.js:269
@resource://gre/modules/devtools/dbg-client.jsm:359
@resource://gre/modules/devtools/dbg-client.jsm:102
EV_notify@resource://gre/modules/devtools/dbg-client.jsm:159
@resource://gre/modules/devtools/dbg-client.jsm:249
@resource://gre/modules/devtools/dbg-client.jsm:693
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:118
then@resource://gre/modules/commonjs/sdk/core/promise.js:43
then@resource://gre/modules/commonjs/sdk/core/promise.js:153
DC_onPacket@resource://gre/modules/devtools/dbg-client.jsm:697
@resource://gre/modules/devtools/server/main.js -> resource://gre/modules/devtools/server/transport.js:198
@resource://gre/modules/devtools/DevToolsUtils.jsm -> resource://gre/modules/devtools/DevToolsUtils.js:72


###!!! [Parent][AsyncChannel] Error: Channel error: cannot send/recv

reference to undefined property exn.stack
```

ochameau:
After hotfixing 4 issues breaking b2g desktop... I got it to work on windows!
Can you try with the very last changeset I just (forced) pushed?

ochameau:
The exception about this.connection isn't harmfull.

ochameau:
Also, the duplicate connect button is a regression in m-c, it should be fixed in the next nightly.
And there is also something broken with debug button, that isn't simulator specific.
When clicking on DEBUG while the app isn't running, it may fail and deadlock the button,
it should work fine if you refresh the app manager and click on DEBUG again now that the app is already running.
(something is broken with auto-launch)

mykmelez:
Still a black screen on Mac. Here's the complete terminal output:

```
console.log: r2d2b2g: 1379438083464 Marionette  INFO    MarionetteComponent loaded
console.log: r2d2b2g: 1379438083466 Marionette  INFO    marionette not enabled via pref
console.log: r2d2b2g: desktop-helper: +++ loading scripts for app: resource://gre-resources/hiddenWindow.html
console.log: r2d2b2g: PermissionsTable.jsm: expandPermissions: Unknown Permission: inputmethod-manage
console.log: r2d2b2g: PermissionsInstaller.jsm: 'inputmethod-manage' is not a valid Webapps permission name.
console.log: r2d2b2g: processing -dbgport command line option
console.log: r2d2b2g: remote debugger will start on port: 64508
console.error: r2d2b2g: 
  OpenGL version detected: 210
console.error: r2d2b2g: 
  OpenGL version detected: 210
console.log: r2d2b2g: ************************************************************
* Call to xpconnect wrapped JSObject produced this error:  *
[Exception... "'[JavaScript Error: "Argument 1 of Node.removeChild is not an object." {file: "chrome://browser/content/shell.js" line: 244}]' when calling method: [nsIDOMEventListener::handleEvent]"  nsresult: "0x80570021 (NS_ERROR_XPC_JAVASCRIPT_ERROR_WITH_DETAILS)"  location: "native frame :: <unknown filename> :: <TOP_LEVEL> :: line 0"  data: yes]
************************************************************
```

ochameau:
There is some really interesting hacks on mac...
http://mxr.mozilla.org/mozilla-central/source/b2g/chrome/content/shell.html?force=1#32

I just pushed something that should fix it. I just tested it on windows again, I hope that will fix the last issue on Mac!

mykmelez:
Merged with updates to the metadata per conversation with @ochameau on IRC!

mykmelez:
For those following along, here are builds from the tip with these changes!
- Windows: https://ftp.mozilla.org/pub/mozilla.org/labs/r2d2b2g/fxos_1_2_simulator-6.0pre1-windows.xpi
- Mac: https://ftp.mozilla.org/pub/mozilla.org/labs/r2d2b2g/fxos_1_2_simulator-6.0pre1-mac.xpi
- Linux: https://ftp.mozilla.org/pub/mozilla.org/labs/r2d2b2g/fxos_1_2_simulator-6.0pre1-linux.xpi

PR #757, opened 2013-07-31
make receipts test stand alone Fixes #707

### LINKED ISSUE #707: receipts tests depend on other tests, hang if run by themselves
The receipts tests hang if you run them by themselves via `make test TEST=test-receipts`. So they seem to depend on other tests, which is dangerous (and breaks the common use case of running a subset of tests). So we should break that dependency.

@nickdesaulniers says he can take this on.

nickdesaulniers:
review? @ochameau


mykmelez:
@ochameau I asked @nickdesaulniers to ask you for review because it isn't clear to me whether `require("main")` in the test script is a real fix for this issue or just a workaround.

I _think_ it's a real fix, because the _receipts_ test is not a unit test of an individual module but rather an integration (or system) test of a user workflow. And the _main_ module has to have been loaded in order for that workflow to succeed, because it controls interaction between the Dashboard and the Simulator.

However, I'd like your expert opinion on the matter! Also, even if loading _main_ is a real fix, should we unload it after concluding the tests, so the next test script starts with a clean slate when running multiple test scripts?

nickdesaulniers:
Ooops looks like something else was the problem!

nickdesaulniers:
Sorry, closing this until it's 100%

nickdesaulniers:
Sorry for this horrible mess of a pull request.  @ochameau , this patch should work, but it appears that `require` in the test cases is not caching the main.js module.  My patch fixes the stand alone test `make test TEST=test-receipts` but breaks when run altogether `make test`.  When run altogether, it looks like the simulator is responding to events twice, either there are two instances of it, or the events were attached twice.  test-main simply requires main.js, but requiring it in test-receipts is introducing some strangeness to the test cases.  Does require not cache modules in test cases, or is there something else I'm doing wrong?

ochameau:
Each test is run in its own loader. That means that if two test require main module, two instances are going to be spawnd. I'd also imagine that each loader, for each test are going to be destroyed, so each main module is going to be unloaded. But I don't remember explicitely seing this destroy after each time.
Also, even if we do destroy loaders, we are most likely not cleaning up absolutely everything on unload, and, there is many chances that some cleanups are asynchronous and happen too late.

Having said that, I'd suggest to move all such integration tests to the end of the dictionary, in order to run them last. (Tests are run by alphabetical order) I hope that can solve these issues... (the adb test seems to already rely on that?)

mykmelez:
> Having said that, I'd suggest to move all such integration tests to the end of the dictionary, in order to run them last. (Tests are run by alphabetical order) I hope that can solve these issues... (the adb test seems to already rely on that?)

The problem is that we should be able to run tests both in isolation and together. So both of these test runs should succeed:

```
make test TEST=receipts
make test TEST='main\|receipts'
```

But currently only the second one succeeds; and with this change in this pull request, only the first one succeeds.

mykmelez:
If we can't break the dependency, I think I'd rather move the test functions from test-receipts.js into test-main.js. Then at least there will be no question about whether we can run these tests independently and whether "main" is loaded at the time the receipts tests run.

So that seems like the best short-term solution, although in the long term it should be possible for the receipts tests (and other such tests) to be truly independent, perhaps by making sure that the main module is completely unloaded after the main tests complete.

nickdesaulniers:
review? @mykmelez

PR #759, opened 2013-08-01
ADB in make test

bkase:
- [x] Tests pass on Mac and Linux if ADB is already open in a background process
- [ ] Better error message if libadb is open in different Firefox profile
- [ ] Fix crash on Windows with test-receipts (waiting on #757, hopefully it's related)

Unfortunately, most of my tests rely on `ADB.shell` which the fallback ADB module does not support so those tests are skipped for the first fix.


bkase:
I found a weird edge case. I'm going to make this an issue for now.

Issue #760, opened 2013-08-01
ADB in make test


bkase:
- [x] Tests pass on Mac and Linux if ADB is already open in a background process
- [x] Better error message if libadb is open in different Firefox profile
- [x] Fix crash on Windows with test-receipts (waiting on #757, hopefully it's related)

It seems like the tests are very interdependent. The current _test-z-adb.js_ executes `ADB._startAdbInBackground()` which exploits https://bugzilla.mozilla.org/show_bug.cgi?id=887914 so this instance of ADB is the one that's open throughout all the tests. If any other test tries to open ADB it's instance of ADB will be broken (I think this doesn't break other tests because the other tests don't actually use any ADB features).

In order to check if ADB is running or not ADB has to be required later. If it's required in an actual test in _test-z-adb_ then it will break only the tests in _test-z-adb_ (since another ADB will have opened first). If ADB is loaded at the top-level in a promise callback like what's in the simulator now (`require("adb/adb-running-checker").check()`), Firefox just crashes in the middle of one of the earlier tests (at least on Mac).

bkase:
I'm going to wait to see what happens with #757 before continuing to explore the other two test bugs. I think that fixing #757 will make these bugs easier to fix.

PR #763, opened 2013-08-01
Better error message if ADB is already running

bkase:
Simple change to give a better message if you try to run `make test` if another Simulator has started up ADB. See #760

### DIFF
--- a/addon/lib/adb/adb-running-checker.js
+++ b/addon/lib/adb/adb-running-checker.js
@@ -40,7 +40,7 @@ exports.check = function check() {
         let { length, data } = client.unpackPacket(aData);
         debug("length: ", length, "data: ", data);
         socket.close();
-        deferred.resolve(true);
+        deferred.resolve(data.indexOf("001f") != -1);
         break;
       default:
         debug("Unexpected State: " + state);
--- a/addon/lib/adb/adb.js
+++ b/addon/lib/adb/adb.js
@@ -268,7 +268,10 @@ exports._startAdbInBackground = function startAdbInBackground() {
 
   serverWorker.emit("init", { libPath: libPath }, function initack() {
     serverWorker.emit("start", { port: 5037, log_path: File.join(TmpD, "adb.log") }, function started(res) {
-      console.debug("adb server thread returned: " + res.result);
+      console.debug("adb server thread returned: " + res.ret);
+      if (res.ret == -1) {
+        Services.obs.notifyObservers(null, "adb-port-in-use", null);
+      }
     });
   });
 
--- a/addon/test/test-main.js
+++ b/addon/test/test-main.js
@@ -4,7 +4,7 @@
 
 /* Note: all tests that depend on firefox state should be put in this
  * file.  Re-requiring main in other tests causes the simulator to be
- * reinstantiated. 
+ * reinstantiated.
  */
 
 //==========================
@@ -25,6 +25,7 @@ exports = (function(exports) {
   Cu.import("resource://gre/modules/Services.jsm");
 
   let isPhonePluggedIn = null;
+  let isPortInUse = false;
 
   let observer = {
     QueryInterface: XPCOMUtils.generateQI([Ci.nsIObserver,
@@ -40,6 +41,9 @@ exports = (function(exports) {
           break;
         case "adb-device-disconnected":
           isPhonePluggedIn = false;
+          break;
+        case "adb-port-in-use":
+          isPortInUse = true;
           break;
       }
     }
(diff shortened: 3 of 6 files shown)


addon/test/test-z-adb.js
```diff
@@ -210,9 +210,17 @@ exports["test f device tracking, with phone"] = function(assert, done) {
...
-  assert.pass("Done!");
-  done();
+  try {
+    ADB.close();
+    assert.pass("Done!");
+    done();
+  } catch (e) {
+    console.error(e);
+    assert.fail("Have you opened the simulator in a different Firefox " + 
+                "profile? Make sure you close that Firefox process before " +
```
mykmelez: Nit: add \n between the two sentences so they wrap nicely.

addon/test/test-z-adb.js
```diff
@@ -210,9 +210,17 @@ exports["test f device tracking, with phone"] = function(assert, done) {
...
+                "running the tests");
```
mykmelez: Nit: add a period to the end of this sentence.

addon/test/test-z-adb.js
```diff
@@ -210,9 +210,17 @@ exports["test f device tracking, with phone"] = function(assert, done) {
...
-  ADB.close();
-  assert.pass("Done!");
-  done();
+  try {
+    ADB.close();
+    assert.pass("Done!");
+    done();
+  } catch (e) {
+    console.error(e);
+    assert.fail("Have you opened the simulator in a different Firefox " +
```
mykmelez: Nit: simulator -> Simulator

addon/test/test-z-adb.js
mykmelez: This assert should probably include a describe of the error. I know the error message reported by console.error contains it, but it isn't necessarily clear how the two are related. So I would make it say something like:

```
Exception while closing ADB.  Have you opened the Simulator in a different
Firefox profile?  Make sure you quit that process before running the tests.
```

mykmelez:
Other than the minor quibbles about the content of the message, this seems ok, but it seems like it just papers over the issue, which is deeper than the test code, given that the test just calls a public method (`ADB.close()`).

Also, I see a similar exception when running only the "main" test script, whether or not libadb is open in another process:

```
08-05 15:50 > make test TEST=main
…
info: r2d2b2g: Closing ADB
info: r2d2b2g: [JavaScript Error: "TypeError: socket is undefined" {file: "resource://gre/modules/XPIProvider.jsm -> file:///var/folders/lp/8t_7y24119720_hjp5wc_4cr0000gn/T/tmp6kbnDY.mozrunner/extensions/r2d2b2g@mozilla.org/bootstrap.js -> resource://gre/modules/commonjs/toolkit/loader.js -> resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/adb/adb-device-tracker.js" line: 115}]
info: r2d2b2g: Simulator.onMessage toggle
!!! error running onStopped callback: TypeError: callback is not a function
  File "/Users/myk/Mozilla/r2d2b2g/addon-sdk/bin/cfx", line 33, in <module>
    cuddlefish.run()
  File "/Users/myk/Mozilla/r2d2b2g/addon-sdk/python-lib/cuddlefish/__init__.py", line 928, in run
    bundle_sdk=options.bundle_sdk)
  File "/Users/myk/Mozilla/r2d2b2g/addon-sdk/python-lib/cuddlefish/runner.py", line 743, in run_app
    OUTPUT_TIMEOUT, test_name, parseable)
Exception: Test output exceeded timeout (60s).
make: *** [test] Error 1
```

But perhaps this is a different issue?

bkase:
hmm, I'll fix the other problems with the tests first and see if I can narrow the scope of this try-catch

bkase:
I think that exception that you found before was fixed with #775

mykmelez:
Indeed, I no longer get that exception, but I'm having trouble figuring out how to test this change. If I run tests with libadb already loaded in another process I get:

```
info: r2d2b2g: adb: Found ADB process running, not restarting
info: r2d2b2g: pass: Started
info: r2d2b2g: executing 'r2d2b2g/tests/test-main.test ab list devices'
…
info: r2d2b2g: adb: listDevices
info: r2d2b2g: adb: runCommand host:devices
info: r2d2b2g: adb: runCommand onopen
info: r2d2b2g: adb: len=16 30303043686f73743a64657669636573 000Chost:devices
info: r2d2b2g: adb: runCommand ondata
info: r2d2b2g: pass: Devices: ["full_unagi"]
```

Perhaps because of #765?

bkase:
Can you try `make package` on this branch and load that XPI in another profile? 
I modified some of the native code to aid in detection, so it won't actually detect any libadb compiled before this change.

mykmelez:
>  I modified some of the native code to aid in detection, so it won't actually detect any libadb compiled before this change.

Ah, indeed! I still see #784, but that isn't as much of a problem now that this test fails. In fact perhaps it's the right behavior. So I'll merge this and close that issue.

PR #764, opened 2013-08-02
Refactor old C->JS code to use new JsMessages

bkase:
Started big code refactor for moving all communication from C to JS to the new JsMessage.
See tasks
- [x] Send message responsible for restarting ADB
- [x] Send getLastError message (needed for Windows driver calls)
- [x] Send thread spawning messages

I made a minor change to MSG. Now it just returns the `void *` (that way it's easy to ignore). I still think it should be a macro just so we can avoid having to cast the argument to a `void *`.

Also made CommonMessageHandler so we don't have to copy paste handlers for messages common to all threads (which right now is ADB restarting).

### DIFF
--- a/addon/lib/adb/adb-io-thread.js
+++ b/addon/lib/adb/adb-io-thread.js
@@ -13,8 +13,16 @@ const EVENTED_CHROME_WORKER_URL = URL_PREFIX + "evented-chrome-worker.js";
 const CONSOLE_URL = URL_PREFIX + "worker-console.js";
 const IOUTILS_URL = URL_PREFIX + "io-utils.js";
 const ADB_TYPES = URL_PREFIX + "adb-types.js";
+const JS_MESSAGE = URL_PREFIX + "js-message.js";
+const COMMON_MESSAGE_HANDLER = URL_PREFIX + "common-message-handler.js";
 
-importScripts(INSTANTIATOR_URL, EVENTED_CHROME_WORKER_URL, CONSOLE_URL, IOUTILS_URL, ADB_TYPES);
+importScripts(INSTANTIATOR_URL,
+              EVENTED_CHROME_WORKER_URL,
+              CONSOLE_URL,
+              IOUTILS_URL,
+              ADB_TYPES,
+              JS_MESSAGE,
+              COMMON_MESSAGE_HANDLER);
 
 const worker = new EventedChromeWorker(null, false);
 const console = new Console(worker);
@@ -22,9 +30,14 @@ const console = new Console(worker);
 let I = null;
 let libadb = null;
 let io;
-let restartMeFn = function restart_me() {
-  worker.emitAndForget("restart-me", { });
-};
+let jsMsgFn = CommonMessageHandler(worker, console, function(channel, args) {
+  switch(channel) {
+    default:
+      console.log("Unknown message: " + channel);
+  }
+
+  return JsMessage.pack(-1, Number);
+});
 
 worker.once("init", function({ libPath }) {
   I = new Instantiator();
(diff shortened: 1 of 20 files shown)


addon/lib/adb/common-message-handler.js
```diff
@@ -0,0 +1,45 @@
...
+;(function(exports, module) {
+
+  let isModule = !!module;
+  if (isModule) {
+    const { Cu } = require("chrome");
+    Cu.import("resource://gre/modules/ctypes.jsm");
+    const { JsMessage } = require("adb/js-message");
+  } else {
+    module = {};
+  }
```
ochameau, line 23: Isn't it always used as a worker script? If yes, I'm not sure that's really worth keeping the comonjs compatibility code.

addon/lib/adb/adb-io-thread.js
```diff
@@ -22,9 +30,14 @@ const console = new Console(worker);
...
-let restartMeFn = function restart_me() {
-  worker.emitAndForget("restart-me", { });
-};
+let jsMsgFn = CommonMessageHandler(worker, console, function(channel, args) {
+  switch(channel) {
+    default:
+      console.log("Unknown message: " + channel);
+  }
+
+  return JsMessage.pack(10, Number);
```
ochameau: Here and elsewhere, it is weird to return `10`, I think we should return `0` if we don't have any particular value to send back.

ochameau:
Looks good given the nits being fixed.

addon/lib/adb/common-message-handler.js
bkase, line 23: I was thinking that it might be used from adb-blocking-native.js at some point (which runs on the UI thread with CommonJS during unload).

addon/lib/adb/adb-io-thread.js
bkase: The reason I didn't pick zero is because if there is a bug in the JS callback, the native code receives a `0`. If we return something other than `0`, then it's easy to tell if the JS code is failing when debugging.

`10` is pretty random though, is there a good non-zero number to return?

bkase:
@ochameau can you look at my replies to your nits?

addon/lib/adb/common-message-handler.js
ochameau, line 23: ok, makes sense.

addon/lib/adb/adb-io-thread.js
ochameau: I don't really understand. Do you want to distinguish the cases where the js callback throws an exception and end up returning `0` and the case where the JS callback explicitely return `0` in case of "more expected" error?

bkase: I want to distinguish between dieing inside of JavaScript and returning the default value if no message was handled. It probably should have been -1. In fact, I accidentally changed it to -1 in 5eeddd371892f96316c81248d38209c539439ad9. Should I also do it here?

addon/lib/adb/common-message-handler.js
```diff
@@ -0,0 +1,50 @@
...
+        case "restart-adb":
+          worker.emitAndForget("restart-me", { });
+          return JsMessage.pack(1, Number);
+        default:
+          try {
+            return andThen(channel, args);
+          } catch (e) {
+            console.log("JS exception: " + e);
+            return JsMessage.pack(-1, Number);
+          }
```
ochameau: I think it will be even better to encapsulate the whole function body, to catch eventual errors on worker.emitAndForget("restart-me", {});
or on channel.readString(). So from line 27 to 39.

ochameau:
Thanks again for your work!
Feel free to land this pull request when you think it is ready.

PR #765, opened 2013-08-02
More robust check to see if ADB is running

bkase:
Faster, more robust check to see if ADB is running on bootup. Opens a socket for port 5037 and asks for "host:version" (which is exactly what the ADB commandline client does). If only I knew about this earlier, I wouldn't have wasted a bunch of time getting that `ps` solution working.

@ochameau This is better than using `ps` or `taskList.exe` right?

### DIFF
--- a/addon/lib/adb/adb-running-checker.js
+++ b/addon/lib/adb/adb-running-checker.js
@@ -3,67 +3,75 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+/*
+ * Uses host:version service to detect if ADB is running
+ * Modified from adb-file-transfer from original ADB
+ */
+
 'use strict';
 
+const { Cu, Cc, Ci } = require("chrome");
+
 const Promise = require("sdk/core/promise");
-const subprocess = require("subprocess");
+const client = require("adb/adb-client");
 
-const { platform } = require("system");
-const env = require("api-utils/environment").env;
-const File = require("file");
+function debug() {
+  console.debug.apply(console, ["ADB: "].concat(Array.prototype.slice.call(arguments, 0)));
+}
 
-const psRegexNix = /.*adb .*fork\-server/;
-const psRegexWin = /adb.exe.*/;
-module.exports = {
-  check: function check() {
-    let deferred = Promise.defer();
+exports.check = function check() {
+  let deferred = Promise.defer();
+  let socket;
+  let state;
 
-    let ps, args;
-    if (platform === "winnt") {
-      ps = "C:\\windows\\system32\\tasklist.exe";
-      args = [];
-    } else {
-      args = ["aux"];
-      let psCommand = "ps";
+  debug("Asking for host:version");
 
-      let paths = env.PATH.split(':');
-      let len = paths.length;
-      for (let i = 0; i < len; i++) {
-        try {
-          let fullyQualified = File.join(paths[i], psCommand);
-          if (File.exists(fullyQualified)) {
-            ps = fullyQualified;
-            break;
-          }
-        } catch (e) {
-          // keep checking PATH if we run into NS_ERROR_FILE_UNRECOGNIZED_PATH
-        }
-      }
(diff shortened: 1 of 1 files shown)


addon/lib/adb/adb-running-checker.js
```diff
@@ -3,67 +3,93 @@
...
+
 'use strict';
 
+const { Cu, Cc, Ci } = require("chrome");
+
 const Promise = require("sdk/core/promise");
-const subprocess = require("subprocess");
+const client = require("adb/adb-client");
+let { TextEncoder, TextDecoder } = Cu.import("resource://gre/modules/Services.jsm");
+/* const OS = */ Cu.import("resource://gre/modules/osfile.jsm");
```
ochameau: TextEncoder, TextDecoder and OS aren't used in this module.

addon/lib/adb/adb-running-checker.js
```diff
@@ -3,67 +3,93 @@
...
+
+function debug() {
+  console.debug.apply(console, ["ADB: "].concat(Array.prototype.slice.call(arguments, 0)));
+}
+
+exports.check = function adb_push(aFrom, aDest) {
```
ochameau: aFrom and aDest aren't used in this function.

addon/lib/adb/adb-running-checker.js
```diff
@@ -3,67 +3,93 @@
...
-const psRegexNix = /.*adb .*fork\-server/;
-const psRegexWin = /adb.exe.*/;
-module.exports = {
-  check: function check() {
-    let deferred = Promise.defer();
+  let shutdown = function() {
+    debug("push shutdown");
+    socket.close();
+    deferred.reject("BAD_RESPONSE");
+  }
```
ochameau: shutdown isn't used in this function.

addon/lib/adb/adb-running-checker.js
```diff
@@ -3,67 +3,93 @@
...
+      case "start":
+        state = "send-version";
+        debug("Asking for host:version");
+        runFSM();
+        break;
+      case "send-version":
+        req = client.createRequest("host:version");
+        socket.send(req);
+        state = "wait-version";
+        break
```
ochameau, line 36: What about squashing `send-version` and `start` like this?

```
case "start":
  req = client.createRequest("host:version");
  socket.send(req);
  state = "wait-version";
```

addon/lib/adb/adb-running-checker.js
```diff
@@ -3,67 +3,93 @@
...
+        //       supports the commands we want to use
+        let { length, data } = client.unpackPacket(aData);
+        debug("length: ", length, "data: ", data);
+        state = "end";
+        runFSM();
+        break
+      case "end":
+        socket.close();
+        deferred.resolve(true);
+        break;
```
ochameau, line 44: Can we also squash `wait-version` and `end`?

addon/lib/adb/adb-running-checker.js
```diff
@@ -3,67 +3,93 @@
...
+    debug("runFSM " + state);
+    let req;
+    switch(state) {
+      case "start":
+        state = "send-version";
+        debug("Asking for host:version");
+        runFSM();
+        break;
+      case "send-version":
+        req = client.createRequest("host:version");
```
ochameau: `req` is only used in this `case`, we can move `let req` to this line.

ochameau:
I'm wondering if we should do something report error, or try to launch adb, on promise rejection.
The previous code wasn't using deferred.reject, but the new code does
although the calling code from simulator.js doesn't do anything on promise rejection:
https://github.com/mozilla/r2d2b2g/blob/master/addon/lib/simulator.js#L122-L128

bkase:
I made the promise resolve false in any error rather than rejecting the promise. That way the simulator will just start ADB itself on any error.

ochameau:
Looks good. Please merge if you think that's ready.

PR #770, opened 2013-08-02
Install hosted apps from querystring Fixes #743

### LINKED ISSUE #743: Make Simulator callable from a website
I'd like to call the Simulator with the link to the manifest.
That would make a great testing tool for jsFiddle users (manifest is currently generated - add /webapp.manifest to the URL, like in http://jsfiddle.net/zalun/AbMUa/manifest.webapp)

nickdesaulniers:
review? @mykmelez

Myk, this allows you to install a hosted app by appending the querystring ?install=[encoded manifest path] where [encoded manifest path] is a url to a hosted manifest that has been base64 encoded, then URI encoded.  For example:

`encodeURIComponent(btoa('http://www.mykzilla.org/app/manifest.webapp')); // -> "aHR0cDovL3d3dy5teWt6aWxsYS5vcmcvYXBwL21hbmlmZXN0LndlYmFwcA%3D%3D"`

Try appending `?install=aHR0cDovL3d3dy5teWt6aWxsYS5vcmcvYXBwL21hbmlmZXN0LndlYmFwcA%3D%3D` to the URL of the dashboard, once it's open.

The reason for the base 64 encoding is that only doing URI encoding doesn't work for resource:// URL's; FF reports the URL as broken and substitutes some of the characters.

### DIFF
--- a/addon/data/content/js/main.js
+++ b/addon/data/content/js/main.js
@@ -30,16 +30,8 @@ var Simulator = {
 
     $('#form-add-app').on('submit', function(evt) {
       evt.preventDefault();
-
-      var input = $('#add-app-url');
-      var valid = input[0].checkValidity();
-      // hosted and generated apps
-      window.postMessage({
-        name: "addAppByTab",
-        url: input.val().trim()
-      }, "*");
-      $("#form-add-app").get(0).reset();
-    });
+      this.submit();
+    }.bind(this));
 
     document.documentElement.addEventListener(
       "addon-message",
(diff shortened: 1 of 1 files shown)


mykmelez:
I can see why it's useful to enable sites to add apps to the Simulator, but this approach isn't going to work, because web sites can't load a resource: URL. For example, when I try to load resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/index.html from the Mykzilla app running in the browser, the browser complains:

> [12:57:40.711] Security Error: Content at http://www.mykzilla.org/app/ may not load or link to resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/index.html.
> [12:57:40.711] Error: Access to 'resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/data/content/index.html' from script denied @ http://www.mykzilla.org/app/app.js:19

And if it was possible, then it would very dangerous, since any web site could then automatically add a privileged app to the user's system which would immediately run (because the Simulator automatically starts and loads an app you add to it) and could then abuse its privileges to cause all sorts of trouble.

(It would also break the mental model of the Simulator Dashboard as a special-purpose page that can only ever be open in a single tab, like about:addons.)

So we should find another way to do this, like what @ochameau recently implemented for Marketplace to enable reviewers to install apps they're reviewing. Take a look at commit 1011908cdf53c9009c246ba8a5450a03907901e0, which added it, for details.

mykmelez:
@nickdesaulniers I'm going to close this pull request, as the changes have rotted, and the request is dormant. But please reopen if you pick this back up!

PR #771, opened 2013-08-02
No-socket ADB devices

bkase:
Listing devices re-uses information from the device-tracker. It's still enabled in the native code via sockets because some of that native code needs to be there for cleanup things, let me know if I should disable the socket entry-point.

Depends on #755, I will rebase to keep this up-to-date with that pull-request.


addon/lib/adb/adb-device-tracker.js
```diff
@@ -21,98 +23,79 @@ function debug() {
...
-    socket.close();
-    hasDevice = false;
+  listDevices: function() {
+    let deferred = Promise.defer();
+    let buf = [];
+    for (let dev in devices) {
+      buf.push([dev, devices[dev]]);
+    }
+    deferred.resolve(buf);
+    return deferred.promise;
```
mykmelez: To create a resolved promise, you don't need to call _defer_ and then _resolve_ the deferred object. Simply call _Promise.resolve_ with the value, and it'll return a resolved promise with that value, i.e.:

``` javascript
  listDevices: function() {
    let buf = [];
    for (let dev in devices) {
      buf.push([dev, devices[dev]]);
    }
    return Promise.resolve(buf);
  },
```

mykmelez:
Other than the aforementioned simplification, this looks fine, although it makes me think we should simplify the interface of _listDevices_ and just return the _devices_ object instead of reflecting it into an array. But I understand that would be an API change that would require touching a lot more code, and it isn't a big deal.

bkase:
Tests crash sometimes on OSX on this changeset. This needs to be looked into more closely.

mykmelez:
A recent commit to this branch re-adds test-z-adb.js, whose contents we recently moved to test-main.js. Is that intentional?

bkase:
I think having _test-z-adb.js_ was the reason that there were crashes. I'm not seeing any crashes anymore.

mykmelez:
> I think having test-z-adb.js was the reason that there were crashes. I'm not seeing any crashes anymore.

Me neither!

PR #774, opened 2013-08-05
Improve Windows performance immensely

bkase:
Busy-waiting for async reads (assuming system has 2 cores) improves the
performance due to numerous acks in ADB's protocol. Before we were waiting
100ms before every ack! (and there are a ton of acks)

If we need better performance on systems with one core, we should change
this to Sleep(1) to prevent resource starvation (the writer thread goes super slowly) but even 1ms sleeping slows down `adb push` considerably.

### DIFF
--- a/android-tools/adb-bin/usb_windows.cpp
+++ b/android-tools/adb-bin/usb_windows.cpp
@@ -446,6 +446,7 @@ int usb_read(usb_handle *handle, void* data, int len) {
           // the input thread will notify_should_kill
         }
 
+        Sleep(1);
       } while(!hasCompleted);
       D("Post hasComplated\n");
 
--- a/android-tools/adb-bin/usb_windows.cpp
+++ b/android-tools/adb-bin/usb_windows.cpp
@@ -446,7 +446,6 @@ int usb_read(usb_handle *handle, void* data, int len) {
           // the input thread will notify_should_kill
         }
 
-        Sleep(100);
       } while(!hasCompleted);
       D("Post hasComplated\n");
 


mykmelez:
This does indeed speed up performance on my two-core Windows 7 VM tremendously, although Firefox crashed the first time I tried to push CubeVid to the device, and it still paused for 10-15 seconds after sending the last segment:

```
debug: r2d2b2g: ADB:  Sending 1827 bytes
debug: r2d2b2g: ADB:  len=4 44415441 DATA
debug: r2d2b2g: ADB:  len=4 23070000 #...
debug: r2d2b2g: ADB:  len=1827 29e7593106f2d8c044ec9ec847a6d5af75d0cff9 ).Y1....D...G...u...
debug: r2d2b2g: ADB:  len=4 444f4e45 DONE
debug: r2d2b2g: ADB:  len=4 453efc51 E>.Q
(10-15 second pause)
debug: r2d2b2g: ADB:  push ondata
debug: r2d2b2g: ADB:  runFSM wait-done
debug: r2d2b2g: ADB:  view[0] = 1497451343
debug: r2d2b2g: ADB:  DONE: OK
debug: r2d2b2g: ADB:  runFSM end
info: r2d2b2g: ADB.push success: SUCCESS
debug: r2d2b2g: ADB:  push onclose
TCPSocket: close called
```

mykmelez:
Here's the crash report, although I'm not sure how helpful it can be without symbols:

https://crash-stats.mozilla.com/report/index/9615fdc4-d009-4117-bac8-fcb5f2130806

I eventually also started crashing repeatedly a few seconds after opening the Dashboard on startup with a device connected:

https://crash-stats.mozilla.com/report/index/f2e27110-8a85-4b0c-8a2a-3f9b92130806

mykmelez:
I tried to test for performance on my single-core Windows XP VM, but I keep crashing on startup, so I haven't been able to do so:

https://crash-stats.mozilla.com/report/index/34e14634-f828-45a5-a1d3-fbe622130806

bkase:
I can't reproduce any crashes assuming the phone plugged into my Windows machine is actually a FirefoxOS device (i.e. not Android). 

I am seeing the 10-15 second wait after we send the last segment. The reason this happens is that although we wait for an ack before sending the next packet of data when pushing, ADB acks immediately after recieving a packet of data (not waiting until the remote device gets the full data) as far as I can tell (but I haven't looked into this super closely). Anyway, it always hangs there in the logs when sending big files.

Is 10-15 seconds too slow for CubeVid?

mykmelez:
> Is 10-15 seconds too slow for CubeVid?

Overall, this change makes things much, much faster, so we should take it, provided we can resolve any crashers, even though there is this additional 10-15 second delay. We can work on resolving that delay separately.

bkase:
I just checked on OSX, it looks like it takes around 7 seconds to push CubeVid so that's only around half Windows.

mykmelez:
I'm still seeing crashes after merging the tip of master and then rebuilding on this branch. Here's the top of the stack for one that happened right after I pressed the Push button:

```
    mozjs.dll!61cd5a1c()    
>   libadb.dll!usb_read(usb_handle * handle=0x08c7a020, void * data=0x08c7c87c, int len=24)  Line 408 + 0x23 bytes  C++
    libadb.dll!remote_read(apacket * p=0x08c7c870, atransport * t=0x08c7a110)  Line 55 + 0x15 bytes C++
    libadb.dll!output_thread(void * _t=0x08c7a110, dll_io_bridge * _io_bridge=0x0edffc40)  Line 369 + 0x12 bytes    C++
    libadb.dll!device_output_thread(atransport * t=0x08c7a110, dll_io_bridge * io_bridge=0x0edffc40)  Line 149 + 0xd bytes  C++
```

(Another one happened after the push successfully completed, but I didn't capture a stack for it.)

mykmelez:
Here's one that happened a few seconds after a push successfully completed, right when I clicked on another tab in Firefox:

```
    ntdll.dll!777715de()    
    [Frames below may be incorrect and/or missing, no symbols loaded for ntdll.dll] 
    ntdll.dll!777715de()    
    ntdll.dll!7776014e()    
    nss3.dll!61ec4a25()     
    nss3.dll!61ecce34()     
    nss3.dll!61ec48b2()     
    xul.dll!60536889()  
    xul.dll!60c000c8()  
    xul.dll!60a325d0()  
    mozjs.dll!61d1cd31()    
    mozjs.dll!61d1beaa()    
    mozjs.dll!61dc81dc()    
    mozjs.dll!61cb3844()    
    mozjs.dll!61cb3844()    
    mozjs.dll!61cb3844()    
    mozjs.dll!61cb5a1c()    
>   libadb.dll!usb_read(usb_handle * handle=0x07189dd0, void * data=0x0718df14, int len=24)  Line 408 + 0x23 bytes  C++
    libadb.dll!remote_read(apacket * p=0x0718df08, atransport * t=0x07189ec0)  Line 55 + 0x15 bytes C++
    libadb.dll!output_thread(void * _t=0x07189ec0, dll_io_bridge * _io_bridge=0x0eeb05a0)  Line 369 + 0x12 bytes    C++
    libadb.dll!device_output_thread(atransport * t=0x07189ec0, dll_io_bridge * io_bridge=0x0eeb05a0)  Line 149 + 0xd bytes  C++
```

bkase:
I rebased this on top of master just to see if it would fix the crashes, and it looks like the windows compilation broke somewhere, so it won't compile now until #789 is merged.

It looks like I wasn't experiencing the crashes because apps weren't actually being installed properly on the Geeksphone Peak! (the only thing I have drivers for on Windows)

I think I have to abandon this bug now since I don't have access to a FirefoxOS device anymore :(

mykmelez:
@brendandahl, until we get @bkase another Firefox OS device (working on it!), any chance you can pick this up and figure out why the change is causing crashes on Windows?

brendandahl:
@bkase What was the motivation behind changing from sync read to an async read here https://github.com/mozilla/r2d2b2g/blob/master/android-tools/adb-bin/usb_windows.cpp#L423 ? Wondering, since we're basically doing a sync read now.

brendandahl:
I think I see now, it's so you can kill it while reading.

bkase:
@brendandahl exactly. On Windows, native code has to return back to JS before the worker is killed or else Firefox crashes. Luckily on the other platforms, this is not the case (since they don't have async implementations of the USB I/O)

mykmelez:
@bkase , @brendandahl : are either of you still looking at this?

bkase:
@mykmelez I just spent a bunch of time trying to figure out what I thought was a different bug, so I could attempt to reproduce this crash. The issue I was having was that the simulator was crashing after about 5-10 seconds of being open.
I don't know why you were experiencing something different, but it actually turned out to be the _lack_ of a sleep statement in the async read loop! 

Unfortunately we don't get as much of a speed-up, but I imagine this change will also fix those other crashes you were seeing (I can't reproduce them without this fix since the simulator dies too fast, and with it I'm not seeing any crashes).

mykmelez:
I would be more comfortable with a fix like this if we had a better understanding of why we need to sleep in this loop and how the crash occurs. Nevertheless, since we already have a sleep statement in it, and the latest change speeds up pushes significantly without appearing to introduce instability, I'm comfortable enough to take it. However, we should keep working to get back to reasonable parity with the performance of the executable implementation.

brendandahl:
Sorry missed the ping here (need to start looking at notifications from this repo too).  I am able to reproduce the crash, though it seems to happen at different places depending on the run.  The most common seems to be in somewhere in the jsctypes glue when js_msg is called.  This makes me wonder if calling into js ctypes is not thread safe. I was going to look into this more tomorrow.

PR #775, opened 2013-08-06
ADB `make test` fixes. Fixes #760

### LINKED ISSUE #760: ADB in make test

bkase:
`make test` should now work on all platforms with or without a phone plugged in with or without ADB running in the background before starting the tests.

Unfortunately, I had to move the ADB tests into _test-main.js_ in order to get everything to work (similar to #757). ADB was crashing in 8 different places non-deterministically on OSX (I just tested on OSX) when the tests were in a separate file (despite moving things around).

I grouped the tests in self-executing functions to make things private.


mykmelez:
This is very unfortunate, but I'd rather have working tests than perfectly structured test files.

PR #783, opened 2013-08-09
Geeksphone Peak on Windows Fix

bkase:
AdbWinUsbApi.dll is required for the Peak to connect to ADB. This DLL was never loaded properly, now we pass the proper path to the dll all the way into AdbWinApi.dll so it can load the DLL properly.

I'm assuming that this patch will also make a lot of other untested devices work as well.

Rarely, on exit, the device input thread will not terminate for some reason if a Peak is plugged into a Windows machine. This causes the UI thread to hang indefinitely.

### DIFF
--- a/addon/lib/adb/adb-device-poll-thread.js
+++ b/addon/lib/adb/adb-device-poll-thread.js
@@ -30,11 +30,15 @@ const console = new Console(worker);
 
 const I = new Instantiator;
 let libadb = null;
+let winusbPath_ = null;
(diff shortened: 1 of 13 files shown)


mykmelez:
I don't have a Peak at hand to test with, but this looks good and works as expected with my Unagi. And mostly-working Peak (and other such devices) is better than Peak not working at all. Although we'll want to track down the source of that intermittent failure.

PR #785, opened 2013-08-09
Fix message callback index.

brendandahl:
The wrong index was being stored for the callbacks since push() returns the new length. Callbacks were still getting called since it loops over all the callbacks, but they weren't be freed correctly and once() callbacks could possibly be called multiple times.


bkase:
I'm not sure if I have the authority to accept patches, but lgtm

mykmelez:
> I'm not sure if I have the authority to accept patches, but lgtm

You have it!

PR #792, opened 2013-08-11
Makefile: replace explicit make command with MAKE variable

frasertweedale:
Use $(MAKE) rather than explicitly naming `make' so that recursion
works when invoking`make' by a different name (e.g. `gmake').

See also GNU `make' manual [1].

[1] https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html


mykmelez:
Thanks, @frasertweedale, for the fix! (And also the reference; I didn't know about that variable!). Note: I also fixed some explicit references to _make_ in android-tools/Makefile, which is invoked when you set LIBADB_LOCATION to "local".

Issue #804, opened 2013-08-21
Expand fake geolocation to set more properties


jryans:
Right now, we only simulate the `latitude` and `longitude`, but there are other geolocation [properties](https://developer.mozilla.org/en-US/docs/WebAPI/Using_geolocation#Describing_a_position) we should help with too:
- altitude
- accuracy
- altitudeAccuracy
- heading
- speed

PR #805, opened 2013-08-21
Refactor geolocation to prepare actor for uplifting

jryans:
I am working towards moving the geolocation simulating functionality up to m-c, so that it will already be on-device in the future.

In that vein, I've extracted several actors from the general purpose `dbg-simulator-actors` into `dbg-geolocation-actors` and `dbg-geolocation-ui-actors`.

`dbg-geolocation-actor` is what I would want to get on-device. `dbg-geolocation-ui-actors` is very much dependent on the simulator's UI flow for controlling the geolocation settings, and since that may change significantly with the App Manager, I will leave this here in the simulator add-on for now.

I will likely extend `dbg-geolocation-actor` before moving it on device (as part of issue #804 and perhaps others).

### DIFF
--- a/prosthesis/content/dbg-geolocation-ui-actors.js
+++ b/prosthesis/content/dbg-geolocation-ui-actors.js
@@ -1,0 +1,88 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+  * License, v. 2.0. If a copy of the MPL was not distributed with this
+  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+Cu.import("resource://gre/modules/Services.jsm");
+
+/**
+  * Creates a GeolocationUIActor.
+  * Allows a Firefox OS device to accept fake geolocation data for use during
+  * debugging.
+  */
+let GeolocationUIActor = function GeolocationUIActor(aConnection) {
+  this.debug("Geolocation UI actor created for a new connection");
+
+  this._connection = aConnection;
+  this._listeners = {};
+
+  Services.obs.addObserver(this, "r2d2b2g:enable-real-geolocation", false);
+  Services.obs.addObserver(this, "r2d2b2g:disable-real-geolocation", false);
+};
+
+GeolocationUIActor.prototype = {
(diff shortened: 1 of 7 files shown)


prosthesis/content/dbg-geolocation-ui-actors.js
```diff
@@ -0,0 +1,82 @@
...
+   */
+  debug: function debug() {
+    dump(Array.slice(arguments).join(" ") + "\n");
+  },
+
+  onAttach: function(aRequest) {
+    this.debug("Geolocation UI actor received an 'attach' command");
+
+    return {};
+  }
```
mykmelez, line 77: Nit: it'd be worth noting that the purpose of this function is simply to trigger evaluation of this actor script. Earlier in the diff, I noticed the comment "Actors are initialized lazily, so tell them to attach", but by the time I got here, I had forgotten that and was confused by the "attach" request type, which doesn't seem to do anything, so I had to go back and read through the patch until I reread that comment and realized what was going on here. Future developers (including future versions of us, if we haven't looked at the code in a while!) may be similarly confused.

mykmelez:
I just have that one minor nit. Otherwise, this looks great!

jryans:
Good call, I have added comments to reflect that.

Should I squash the patches, or do you typically do that during merging?

ochameau:
I think it is easier if you do the squash and then we can just click the big green github button ;)

Otherwise, I think we should iterate on this in order to get rid of the always registered `FakeGeolocationProvider.js` xpcom component. I don't think it is reasonable to always load this xpcom when we are going to ship this actor on m-c. We would like to register this xpcom dynamically.
That may be handy to do that in the addon before uplifting the patch to m-c, but feel free to do that during the uplift.

jryans:
Okay, squashed!

@ochameau: I agree, I was already thinking that! The provider component does need to be loaded dynamically. I'll probably handle that in the addon with a separate PR to minimize differences with m-c after the uplift.

mykmelez:
Looks good, thanks @jryans!

PR #816, opened 2013-08-30
add 'make package' libadb target to make libadb packages

mykmelez:
@ochameau, can you take a look at this? It should be straightforward: a `make package` target for libadb (so we don't have to construct these packages by hand).

### DIFF
--- a/android-tools/adb-bin/Makefile
+++ b/android-tools/adb-bin/Makefile
@@ -135,4 +135,4 @@ winlibtest:
 	echo "cl.exe test-js-message.c /link /DLL /OUT:..\\win-out\\libtest.dll" | cmd $(VS_DEVELOPER_CMD_PATH)
 
 package:
-	zip libadb.zip $(PACKAGE_FILES)
+	zip -j libadb.zip $(PACKAGE_FILES)
--- a/android-tools/adb-bin/Makefile
+++ b/android-tools/adb-bin/Makefile
@@ -3,15 +3,24 @@
 
 ADB_AUTH ?= off
 
+PACKAGE_FILES = LICENSE README
+
 ifeq ($(OS), Windows_NT)
   VS_DEVELOPER_CMD_PATH = vcvars32.bat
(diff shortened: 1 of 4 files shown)


android-tools/adb-bin/Makefile
```diff
@@ -3,15 +3,24 @@
...
 ifeq ($(OS), Windows_NT)
   VS_DEVELOPER_CMD_PATH = vcvars32.bat
   ADB_TARGETS = winlib
   WINDOWS_ARTIFACTS = ../win-out vc110.idb vc110.pdb ../adb-win-api/api/build*_w*_x86.* \
       ../adb-win-api/api/obj* ../adb-win-api/api/*.idb ../adb-win-api/api/*.pdb \
       ../adb-win-api/winusb/obj* ../adb-win-api/winusb/build*_w*_x86.*
+  PACKAGE_FILES += \
+    ../win-out/libadb.dll \
+    ../adb-win-api/api/objfre_wxp_x86/i386/AdbWinApi.dll \
+    ../adb-win-api/winusb/objfre_wxp_x86/i386/AdbWinUsbApi.dll
```
ochameau: Wouldn't it be better to use `obj*` instead of `objfre_wxp_x86`, like what is being done for `WINDOWS_ARTIFACTS`?

ochameau:
Ship it!

mykmelez:
> Wouldn't it be better to use obj\* instead of objfre_wxp_x86, like what is being done for WINDOWS_ARTIFACTS?

Yes, it would! Fixed in ec74058c6050cf98d74009d262dde74fd66910b1.

I also fixed a couple other issues:
- Use the `-j` flag to _zip_ to trim directory paths from the files being packaged (Windows-specific, as only Windows has multiple source dirs and builds into separate object dirs).
- Build the same thing (libadb) by default on all three platforms.

mykmelez:
> Ship it!

À votre demande!

PR #826, opened 2013-09-05
Basic 'new project' support.

potch:
Added a "New Project" button to the UI, that asks the user for an app name, then creates a copy of an empty app template at the location of the user's choosing, placing basic files needed to have a working application. Next to come is multiple template support, and the ability to modify more properties of the manifest itself. New in this commit is the DevKit module, which is the home for this functionality in the event we fragment the Simulator in future releases.


addon/lib/devkit.js
```diff
@@ -0,0 +1,74 @@
...
+
+const templatePath = Self.data.url("resources");
+
+function slugify(s) {
+    if (!s) return "";
+    let re = new RegExp("[^\\w" + unicodeLetters + "\\s-]+","g");
+    s = s.replace(re, ' ').trim();
+    s = s.replace(/[-\s]+/g, '-').toLowerCase();
+    return s;
+}
```
mykmelez: This is sophisticated, and I like the idea of supporting Unicode characters outside good ole ASCII; but we have an existing (cruder) function for doing this in WebappOSUtils.jsm (https://github.com/mozilla/mozilla-central/blob/3cfdb57ce92b9305b412b365e0341d6086e5f0a1/toolkit/webapps/WebappOSUtils.jsm#L262-L268), and I suspect we're better off reusing that function (and then making it more liberal, although I'm worried that we were conservative there for good reason).

mykmelez:
In testing, after I set a location for the project, the directory gets created, and a manifest.webapp file gets created within it, but no additional files are created, and the app never appears in the Dashboard. I don't see any errors in the console, however.

addon/lib/devkit.js
potch: Works for me! This was existing code I borrowed from AMO, but happy to switch to something available elsewhere.

potch:
@mykmelez what platform are you testing on?

mykmelez:
I'm testing on Mac OS X 10.8.4!

mykmelez:
Ah, I should have also looked at the terminal. It displays the following error:

```
info: r2d2b2g: Simulator.onMessage createNewApp
info: r2d2b2g: DevKit.createNewApp
info: r2d2b2g: DevKit: Selected /Users/myk/Desktop
info: r2d2b2g: DevKit: cloning from empty-project
NS_ERROR_FILE_NOT_FOUND: Component returned failure code: 0x80520012 (NS_ERROR_FILE_NOT_FOUND) [nsIChannel.open]
    callback.apply(null, args);
  File "resource://gre/modules/commonjs/sdk/content/worker.js", line 111, in onContentEvent/<
    self._emit.apply(self, JSON.parse(args));
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 123, in _emit
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 153, in _emitOnObject
  File "resource://gre/modules/commonjs/sdk/content/worker.js", line 224, in postMessage
    self._addonWorker._emit('message', data);
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 123, in _emit
  File "resource://gre/modules/commonjs/sdk/deprecated/events.js", line 153, in _emitOnObject
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/simulator.js", line 1494, in onMessage
    var manifestPath = DevKit.createNewApp(message.manifestData);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/devkit.js", line 53, in module.exports.createNewApp
    this.cloneDir('empty-project', projectPath);
  File "resource://r2d2b2g-at-mozilla-dot-org/r2d2b2g/lib/devkit.js", line 63, in module.exports.cloneDir
    let templateFileList = Self.data.load(templatePath + 'filelist.json');
  File "resource://gre/modules/commonjs/sdk/self.js", line 37, in read
    return readURISync(uri(path));
  File "resource://gre/modules/commonjs/sdk/net/url.js", line 109, in readURISync
    let stream = channel.open();
```

potch:
The deep shame of forgetting to include a file in your commit. Should be good now.

addon/lib/devkit.js
mykmelez: It should be as simple as `Cu.import("resource://gre/modules/WebappOSUtils.jsm");` and then `WebappOSUtils.sanitizeStringForFilename()` instead of `slugify()`.

mykmelez:
> The deep shame of forgetting to include a file in your commit. Should be good now.

Indeed! And the patch looks good and works well overall. My only other nit is that the validator warns that a new project is "Missing 'icons' in Manifest." It isn't a tragedy, but it's a bit wonky for a new project to generate warnings before the developer has even had a chance to mess it up.

potch:
I agree. I'll add a default blank icon (like the one you have in the other empty-project directory)

mykmelez:
D'oh! How did I miss the commit you added four days ago? I've been awaiting it eagerly!

In any case, I've spotted it now, and it looks really great, so I merged your changes! But note that I merged them into the release-5.0 branch, since we've now branched for the 5.0 release, which will be the last major release of the "classic" version of the Simulator.

Thanks for this awesome feature!

PR #836, opened 2013-09-23
Updates for 1.2/26

ochameau:
He is a bunch of updates for 1.2 version.
First a quick update to the latest aurora builds, that allows to drop some workrounds. (I requested uplifts for the others)
I added yet another workaround (I still need to craft a patch against b2g desktop, for the contenBrowser/shell.start being accessed or called too early) in order to prevent a race condition where we would end up with the reg logo or a black screen on startup.
Also an update of the label being displayed in the app manager, so that we end up displaying "Firefox OS 1.2". (I think that what we agreed on irc?)
Finally, some cleanups in prosthesis to remove non-working UI elements and duplicated actors.


mykmelez:
Overall, this looks fine, but do you have a plan to reenable that UI? It would be a shame to lose its functionality.

mykmelez:
I see that geolocation still partly works via different UI: when a webapp asks for the geolocation, and after you confirm such access via a Gaia overlay, an B2G/OS dialog window appears asking you to share geolocation with "B2G"; and if you confirm the access, the webapp gets your geolocation.

So it looks like you can test apps with your actual geolocation, although there doesn't appear to be a way to set a specific geolocation.

mykmelez:
App > Refresh is also a pretty key feature of the current version of the Simulator. Even though you can do the same thing by going back to Firefox and pressing Update, it's still a significantly better experience to be able to do that within the Simulator itself.

mykmelez:
I'm going to merge this, since the functionality being hidden is already broken, so hiding it is reasonable. But we need a plan to reenable, replace, or intentionally remove it. We shouldn't simply give up on it because it broke.

ochameau:
Geolocation should be implemented when we are going to offer API mock/controls in the app manager.
Ryan already started cleaning the actor, so that we can land it on m-c and ship it on device and b2g desktop and have a common UI to control it. But Paul is having different plan in mind for APIs mock'n control where we wouldn't use actors... So it may not be worth making that work again if we implement it differently.
We haven't restated our goals for Q4, but API should be one, and geolocation is definitely one of the most important one.

Regarding app refresh, I want to get this back ASAP, but I'm focusing to release a new simulator version ASAP with only working feature. I would like to upload it to AMO ASAP to ensure that everyone installing the addon will be updated. Unfortunately, that won't be the case for all people installing the ftp addon :(

mykmelez:
Pushed a followup fix; don't forget to update the version string!

https://github.com/mozilla/r2d2b2g/commit/a3350593119f7438fff9353f9f56550d92be05d2

mykmelez:
> Regarding app refresh, I want to get this back ASAP, but I'm focusing to release a new simulator version ASAP with only working feature. I would like to upload it to AMO ASAP to ensure that everyone installing the addon will be updated. Unfortunately, that won't be the case for all people installing the ftp addon :(

Note that we can add automated updates for ftpmo-hosted packages, so if it turns out to be hard to get regular updates onto AMO, that's an option. It does require the user to allow ftpmo to install addons, but once they do that for the initial install, they don't have to re-approve it for updates.

mykmelez:
> Note that we can add automated updates for ftpmo-hosted packages, so if it turns out to be hard to get regular updates onto AMO, that's an option. It does require the user to allow ftpmo to install addons, but once they do that for the initial install, they don't have to re-approve it for updates.

Alternately, if another site links to the package on ftpmo, then that site is the one the user is asked to allow to install addons. For example, if devmo links to the package, and the user clicks on the link, then Firefox asks the user if devmo should be allowed to install the addon.

PR #839, opened 2013-09-25
remove Firefox permissions added by previous versions of Simulator

mykmelez:
@ochameau, here's an updated pull request for the change I originally submitted in #731, to remove the hidden feature that added permissions to Firefox for hosted apps. This is a request to pull the change into the release-5.0 branch.

I've tested this, although testing is manual and complicated, since it does a version check on install, so you can't use `cfx run`. Here's what I did:
1. create a new profile and start a supported Firefox release (23.0.1, 24.0) with it;
2. install the Simulator stable release (4.0);
3. use the Dashboard to add the Shotclock app via its manifest http://wfwalker.github.io/opensun/online.webapp;
4. use about:config to set extensions.r2d2b2g@mozilla.org.sdk.console.logLevel to the integer value 0;
5. apply the code in this changeset to a local clone and use `make package` to create a package from it;
6. install the package.

You should see something like this in the console:

> info: r2d2b2g: purgePermissions
> info: r2d2b2g: removing permission: wfwalker.github.io geolocation

Note: it isn't clear to me how such a permission is actually stored in Firefox's permissions database and what effect it has on Firefox. Going to http://wfwalker.github.io/ and then selecting Tools > Page Info > Permissions still shows "Share Location" as "Always ask", so maybe the feature hasn't actually had an effect. But since we added the permissions in the past, we should remove them now.


ochameau:
Verified, works great, the permissions sqlite file gets correctly cleaned up.

mykmelez:
Ok, thanks @ochameau! I'll merge this per your review.

PR #840, opened 2013-09-26
Ftp hosting

ochameau:
Here is some tweaks to host simulator xpis to a ftp.
I'm not sure the `release` target has its place in this Makefile, but that ends up being very handy and ensure that we upload xpis to the precise right place. But may be we can move it to ftp folder... or keep it private.

I tested by uploading a first pre3 version with its update.rdf.
No update requested when installing it.
Then uploaded pre4 based on the second revision, and also uploaded its related update.rdf,
the pre3 immediatly request an update and the update works. (I checked addon version, but also b2g version in the settings app)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ endif
 
 B2G_VERSION=1.2
 ADDON_NAME=fxos_1_2_simulator
-ADDON_VERSION=6.0pre3
+ADDON_VERSION=6.0pre4
 XPI_NAME=$(ADDON_NAME)-$(ADDON_VERSION)-$(B2G_PLATFORM).xpi
 
 UPDATE_PATH=$(B2G_VERSION)/$(B2G_PLATFORM)
@@ -55,7 +55,7 @@ B2G_TYPE ?= specific
 # B2G_ID
 
 # Use the current last known revision that sucessfully builds on Windows.
-B2G_URL_BASE = https://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2013-09-22-00-40-01-mozilla-aurora/
+B2G_URL_BASE = https://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2013-09-26-00-40-01-mozilla-aurora/
 
 # Currently, all B2G builds are custom so we can optimize for code size and fix
 # bugs in B2G or its nightly build environments (like 844047 and 815805).
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,15 @@ ifndef B2G_PLATFORM
   endif
 endif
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -35,6 +35,15 @@ ifndef B2G_PLATFORM
   endif
 endif
 
+B2G_VERSION=1.2
+ADDON_NAME=fxos_1_2_simulator
+ADDON_VERSION=6.0pre4
+XPI_NAME=$(ADDON_NAME)-$(ADDON_VERSION)-$(B2G_PLATFORM).xpi
```
ochameau, line 41: I duplicated the addon name in the Makefile in order to ease xpi filename naming. We should ideally pull it from package.json or automagically update package.json from the makefile value. Same applies to addon version.

Makefile
```diff
@@ -164,7 +173,17 @@ run:
...
+
+production: PRODUCTION_ARG=--update-link $(UPDATE_LINK) --update-url $(UPDATE_URL)
+production: package
+	mkdir -p ftp/$(UPDATE_PATH)/
+	cp addon/$(ADDON_NAME).update.rdf ftp/$(UPDATE_PATH)/update.rdf
+	cp addon/$(ADDON_NAME).xpi ftp/$(UPDATE_PATH)/$(XPI_NAME)
+
+release: ftp/$(UPDATE_PATH)/$(XPI_NAME) ftp/$(UPDATE_PATH)/update.rdf
+	cd ftp/ && lftp sftp://stage.mozilla.org -u $(FTP_USER) -e "mput -O /pub/mozilla.org/labs/r2d2b2g/ -d $(UPDATE_PATH)/$(XPI_NAME); quit;"
+	cd ftp/ && lftp sftp://stage.mozilla.org -u $(FTP_USER) -e "mput -O /pub/mozilla.org/labs/r2d2b2g/ -d $(UPDATE_PATH)/update.rdf; quit;"
```
ochameau, line 186: Now that I look at https://github.com/mozilla/r2d2b2g/wiki/Build-Process, lftp sounds like a very complex tool compared to scp :o

mykmelez, line 186: Indeed, lftp looks a lot more complex! And isn't part of default OS installations, unlike scp, which exists everywhere, including in MozillaBuild.

Makefile
mykmelez, line 41: Yup, that'd be a useful enhancement!

mykmelez:
Also useful would be to create "latest" links that point to the latest build, so other sites can link to them instead of having to link directly to the latest build and then be updated every time it changes. But that can happen in a separate changeset.

I haven't tested this, but it looks right by visual inspection.

PR #842, opened 2013-09-30
Ftp tweaks

ochameau:
Some followups from the previous pull request.
Note that, I pushed a new tag (https://github.com/mozilla/r2d2b2g/releases/tag/6.0pre4) associated to the set of xpis I released on ftp (https://ftp.mozilla.org/pub/mozilla.org/labs/fxos-simulator/1.2/)

Am I doing it right?!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ B2G_TYPE ?= specific
 # B2G_ID
 
 # Use the current last known revision that sucessfully builds on Windows.
-B2G_URL_BASE = https://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2013-09-26-00-40-01-mozilla-aurora/
+B2G_URL_BASE = https://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2013-09-29-00-40-04-mozilla-aurora/
 
 # Currently, all B2G builds are custom so we can optimize for code size and fix
 # bugs in B2G or its nightly build environments (like 844047 and 815805).
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ B2G_VERSION=1.2
 ADDON_NAME=fxos_1_2_simulator
 # compute addon version out of package.json
 # matches xx.yy[pre,a,b]zz version patterns
-ADDON_VERSION=$(shell cat addon/package.json | grep version | grep -oE '[0-9]+\.[0-9]+(pre|a|b)?[0-9]+')
+ADDON_VERSION=$(shell grep version addon/package.json | perl -p -e 's/.*([0-9]+\.[0-9]+(pre|a|b)?[0-9]*(dev)?).*/\1/')
 
 XPI_NAME=$(ADDON_NAME)-$(ADDON_VERSION)-$(B2G_PLATFORM).xpi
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -37,12 +37,16 @@ endif
 
 B2G_VERSION=1.2
 ADDON_NAME=fxos_1_2_simulator
-ADDON_VERSION=6.0pre4
+# compute addon version out of package.json
+# matches xx.yy[pre,a,b]zz version patterns
+ADDON_VERSION=$(shell cat addon/package.json | grep version | grep -oE '[0-9]+\.[0-9]+(pre|a|b)?[0-9]+')
```
ochameau: `grep -o` doesn't exists on mozillabuild... I'll have to find something else for windows.

ochameau:
I'll update https://github.com/mozilla/r2d2b2g/wiki/Build-Process accordingly when this patch lands.

Makefile
mykmelez: sed or perl? We seem to use sed quite a bit in the Gaia Makefile.

Makefile
```diff
@@ -176,14 +179,18 @@ package:
...
-	cd ftp/ && lftp sftp://stage.mozilla.org -u $(FTP_USER) -e "mput -O /pub/mozilla.org/labs/r2d2b2g/ -d $(UPDATE_PATH)/$(XPI_NAME); quit;"
-	cd ftp/ && lftp sftp://stage.mozilla.org -u $(FTP_USER) -e "mput -O /pub/mozilla.org/labs/r2d2b2g/ -d $(UPDATE_PATH)/update.rdf; quit;"
+production: locales build package
+
+release: addon/$(ADDON_NAME).xpi addon/$(ADDON_NAME).update.rdf
+	@if [ -z $(SSH_USER) ]; then \
+	  echo "release target requires SSH_USER env variable to be defined."; \
+	  exit 1; \
+	fi
+	ssh $(SSH_USER)@stage.mozilla.org 'mkdir -m 755 -p $(FTP_ROOT_PATH)/$(UPDATE_PATH)'
```
mykmelez: Consider making this 775 so other members of the "labs" group can make changes in these directories if necessary while you're away.

Makefile
```diff
@@ -176,14 +179,18 @@ package:
...
+	chmod 766 addon/$(ADDON_NAME).xpi addon/$(ADDON_NAME).update.rdf
```
mykmelez: This is a weird permission set for a file. Typically such files are 644, although this should perhaps be 664 so other members of the "labs" group can make changes.

mykmelez:
> Note that, I pushed a new tag (https://github.com/mozilla/r2d2b2g/releases/tag/6.0pre4) associated to the set of xpis I released on ftp (https://ftp.mozilla.org/pub/mozilla.org/labs/fxos-simulator/1.2/)
> 
> Am I doing it right?!

The tag looks good to me. The idea is to set it to the revision that was checked out when you produced the builds, so it's possible to reproduce them by checking out the tag and rebuilding.

Makefile
mykmelez: In fact the Build Process doc also uses _sed_:

```
sed -i '' s/${OLD_DEV_VERSION}/${BUILD_VERSION}/g addon/package.json
```

But you need `sed -E` to use extended regular expressions with plus signs, and MozillaBuild's version of _sed_ is too old for that.

So I think you're going to have to fall back to something like this line of Perl, which works on all three platforms:

```
grep version addon/package.json | perl -p -e 's/.*([0-9]+\.[0-9]+(pre|a|b)?[0-9]+).*/\1/'
```

ochameau:
I fixed the file permissions and used perl (it works fine, thanks for the tip!)

PR #854, opened 2013-10-21
Remove unused submodules

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=920198
We now have a bunch of submodules that we no longer use, it may help building the simulator on build farm if we clean this up.


mykmelez:
Looks good!

PR #859, opened 2013-10-28
Random fixes

ochameau:
I have bunch of small modifications against the simulator to make it work again on last versions:
- We have a regression on touch event that introduce unexpected contextmenu event in the homescreen when running an app that ends up introducing the drag'ndrop mode for apps when you get back to the homescreen.
- Fix the crash on MacOS by using last subprocess version.

And some random minor improvements:
- Enable the home button back thanks to the preliminary work being done on the platform in https://bugzilla.mozilla.org/show_bug.cgi?id=926671 It happens that the software button isn't displayed on 1.2. And my platform path for ui.physicalHomeButton hasn't reached 1.2 branch. So we always show the simulator home button.
- Cleanup some code by removing very old workarounds that have been fixed in the platform.
- A quite important fix where we ensure closing the simulator when firefox exit or the addon is disabled/uinstalled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ B2G_TYPE ?= specific
 # B2G_ID
 
 # Use the current last known revision that sucessfully builds on Windows.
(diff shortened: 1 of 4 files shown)


jryans:
These all look good to me.

PR #862, opened 2013-10-31
Bug 929061 - Allows to run the simulator on a custom gaia and/or gecko

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=929061

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ B2G_TYPE ?= specific
 # B2G_ID
 
 # Use the current last known revision that sucessfully builds on Windows.
-B2G_URL_BASE = https://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2013-11-25-00-40-01-mozilla-b2g26_v1_2/
+B2G_URL_BASE = https://ftp.mozilla.org/pub/mozilla.org/b2g/nightly/2014-01-06-00-40-01-mozilla-b2g26_v1_2/
 
 # Currently, all B2G builds are custom so we can optimize for code size and fix
 # bugs in B2G or its nightly build environments (like 844047 and 815805).
@@ -142,16 +142,25 @@ clean:
 profile:
 	cp build/override-prefs.js gaia/build/custom-prefs.js
 	cp build/override-settings.json gaia/build/custom-settings.json
+	# We can't set DESKTOP=1 until bug 930104 reach all branches we are targeting
+	# In the meantime, we set DESKTOP=1 for `make preferences` to make prefs
+	# in a debug profile and then copy them to the regular profile below.
 	NOFTU=1 GAIA_APP_TARGET=production $(MAKE) -C gaia
 	DESKTOP=1 NOFTU=1 GAIA_APP_TARGET=production $(MAKE) -C gaia preferences
 	python build/override-webapps.py
+	# As we do not use DESKTOP=1 for extensions target, helper addons aren't installed into the profile
 	cd gaia/tools/extensions/desktop-helper/ && zip -r ../../../profile/extensions/desktop-helper\@gaiamobile.org.xpi *
 	cd gaia/tools/extensions/activities/ && zip -r ../../../profile/extensions/activities\@gaiamobile.org.xpi *
+	# Remove junks gaia build system creates
 	rm -rf gaia/profile/startupCache gaia/profile/places.* gaia/profile/permissions.sqlite gaia/profile/defaults
+	# Strip dictionaries to reduce xpi size
 	zip -d gaia/profile/webapps/keyboard.gaiamobile.org/application.zip js/imes/latin/dictionaries/*
+	# Remove firefox addons that mess up with the simulator
+	rm -rf gaia/profile/extensions/{browser-helper@gaiamobile.org,gaia-build@gaiamobile.org,httpd@gaiamobile.org,httpd,screen-reader-simulator@gaiamobile.org}
 	rm -rf addon/template
 	mkdir -p addon/template
 	mv gaia/profile addon/template/
+	# As preferences target is executed with DESKTOP=1, it is installed in another debug profile
 	cp gaia/profile-debug/user.js addon/template/profile/
 	cp addon-sdk/app-extension/bootstrap.js addon/template/
 	cp addon-sdk/app-extension/install.rdf addon/template/
--- a/addon/lib/simulator-process.js
+++ b/addon/lib/simulator-process.js
@@ -8,6 +8,8 @@
 const { Cc, Ci, Cu, ChromeWorker } = require("chrome");
 
 Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/osfile.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
 
 const { EventTarget } = require("sdk/event/target");
 const { emit, off } = require("sdk/event/core");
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -142,16 +142,23 @@ clean:
 profile:
 	cp build/override-prefs.js gaia/build/custom-prefs.js
 	cp build/override-settings.json gaia/build/custom-settings.json
+	# We can't set DESKTOP=1 until bug 930104 reach all branches we are targeting
 	NOFTU=1 GAIA_APP_TARGET=production $(MAKE) -C gaia
```
mykmelez, line 148: Nit: it would probably make sense to add something to the comment like:

```
# In the meantime, we set DESKTOP=1 for `make preferences` to make prefs
# in a debug profile and then copy them to the regular profile below.
```

addon/lib/simulator-process.js
```diff
@@ -57,6 +59,29 @@ exports.SimulatorProcess = Class({
...
+                           URL.toFilename(PROFILE_URL), "extensions");
+                         return OS.File.copy(
+                           OS.Path.join(simulatorAddons, name),
+                           profileXpiPath);
+                       }
+                    })
+               .then(this._doRun.bind(this));
+    }
+
+    return this._doRun();
```
mykmelez: Nit: this promise chain would be easier to read as a task!

mykmelez: In fact you could make the whole _run_ function be a task, which would not only make the code more readable, it would mean you wouldn't have to break it into two separate _run_ and _doRun_ functions.

addon/lib/simulator-process.js
```diff
@@ -136,7 +161,11 @@ exports.SimulatorProcess = Class({
   // compute current b2g file handle
   get b2gExecutable() {
     if (this._executable) return this._executable;
-
+    if (Prefs.customRuntime) {
+      this._executable = Prefs.customRuntime;
+      this._executableFilename = "Custom runtime";
+      return this._executable;
+    }
```
mykmelez, line 164: Nit: I would keep the line of whitespace (and put another one after this conditional block) to make it easier to read the multi-part early-returning logic at the top of this getter.

mykmelez, line 164: (I would also take a close look at these magic getters and see if there's a way to simplify them; but that's a separate issue.)

addon/lib/simulator-process.js
```diff
@@ -180,7 +209,11 @@ exports.SimulatorProcess = Class({
   get b2gArguments() {
     let args = [];
 
-    let profile = URL.toFilename(PROFILE_URL);
+    let profile;
+    if (Prefs.gaiaProfile)
+      profile = Prefs.gaiaProfile;
+    else
+      profile= URL.toFilename(PROFILE_URL);
```
mykmelez: Nit: add a space between "profile" and the equals sign.

Also, nit: I hear plans are afoot to land this code in mozilla-central, so it's worth adopting more of its formatting conventions, even if we prefer alternatives. I'm thinking here specifically of the Mozilla convention to surround all conditional blocks with braces, even one-line one-statement blocks like these:

```
    let profile;
    if (Prefs.gaiaProfile) {
      profile = Prefs.gaiaProfile;
    } else {
      profile = URL.toFilename(PROFILE_URL);
    }
```

However, in this case you can probably reduce this logic to a single statement using the tertiary operator:

```
    let profile = Prefs.gaiaProfile ? Prefs.gaiaProfile : 
                                      URL.toFilename(PROFILE_URL);
```

Or perhaps even to a single line:

```
    let profile = Prefs.gaiaProfile || URL.toFilename(PROFILE_URL);
```

addon/package.json
```diff
@@ -31,6 +31,15 @@
   "permissions": {
     "private-browsing": true
   },
+  "preferences": [{
+    "type": "directory",
+    "name": "gaiaProfile",
+    "title": "Select a custom gaia profile directory"
```
mykmelez: Nit: gaia -> Gaia

addon/lib/simulator-process.js
```diff
@@ -48,13 +51,27 @@ exports.SimulatorProcess = Class({
   /**
    * Start the process and connect the debugger client.
    */
-  run: function() {
+  run: function () Task.spawn((function () {
```
mykmelez: Currently, the _launch_ function in main.js is the only code that calls `SimulatorProcess.run()`, and it ignores its return value. But a task returns a promise, and it's good practice to explicitly resolve that promise, so it's obvious from reading the task what the return value will be.

Rejecting the promise within a task is simple: just throw an exception anywhere in the task, and Task will reject the promise with the exception's error message. Resolving the promise, on the other hand, depends on how you define the generator function.

If you define it via the non-standard, SpiderMonkey-specific syntax, i.e. a regular function with at least one yield operator inside it (`function() { … yield … }`), then you must resolve the promise by throwing an instance of _Task.Result_, which is a special exception type that Task catches and uses to resolve the promise:

```
throw new Task.Result("Value");
```

If, on the other hand, you use the new, standard "function star" syntax, i.e. `function*() { … }`, then you can resolve the promise simply by returning a value:

```
return "Value";
```

That's a lot easier to read than `throw new Task.Result("Value")`! So I would make this generator function a function*:

```
  run: function () Task.spawn((function*() {
```

And then at the end of the function I would return something, like _this.process_:

```
    // spawn a b2g instance
    this.process = Subprocess.call({
      …
    });

    return this.process;
  }).bind(this)),
```

addon/lib/simulator-process.js
```diff
@@ -48,13 +51,27 @@ exports.SimulatorProcess = Class({
...
     // kill before start if already running
     if (this.process != null) {
-      this.process
-          .kill()
-          .then(this.run.bind(this));
-      return;
+      yield this.process
+                .kill();
```
mykmelez: Not only is this easier to read, it's also more robust! Before, run() would call kill() and then call itself again. Which means that if kill() had a bug and started resolving its promise without setting _this.process_ to `null`, then the code would enter an infinite loop!

We could have prevented that by passing an argument to run(), like _didAlreadyTryToKill_, and then checking that before trying to kill() again. Which would have prevented the infinite loop, but at the cost of making the code even uglier and harder to read.

Now, however, there is no chance of an infinite loop. Whether kill() succeeds or fails, we don't try calling it again. And if it fails with an exception, then we don't even need to do anything special to handle it, because the exception will cause the run() task to reject its promise.

We might choose to catch that exception, of course; and if kill() rejects its promise, then this yield operation would return a value, and we'd want to check it and react accordingly (`let rv = yield this.process.kill()`). But as it currently stands, we no longer risk looping, which is awesome!

Just one nit: you should be able to put this all on one line:

```
    yield this.process.kill();
```

addon/lib/simulator-process.js
```diff
@@ -90,24 +107,24 @@ exports.SimulatorProcess = Class({
       environment: environment,
 
       // emit stdout event
-      stdout: (function(data) {
+      stdout: (data) => {
         emit(this, "stdout", data);
-      }).bind(this),
+      },
```
mykmelez, line 111: Nice use of fat arrow functions to simplify these function declarations! You probably already know this, but fat arrow functions with a single parameter don't need to enclose that parameter in parentheses, so this could be:

```
  stdout: data => {
    emit(this, "stdout", data);
  },
```

You can even omit the brackets and make this a one-liner:

```
  stdout: data => emit(this, "stdout", data),
```

Although I'm not sure either simplification is worth it. The second one in particular seems misleading, since it returns the return value of emit(), which suggests that the return value has significance, even though it doesn't.

addon/lib/simulator-process.js
ochameau: It feels weird to resolve to process as you can access it from SimulatorProcess instance.
So I explicitely resolved it to null.

PR #863, opened 2013-10-31
Bug 925765 - Enable rotation button back

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=925765


jryans:
I tried this out on OS X 10.8, but it doesn't appear to work. When I click the rotation button, nothing happens.

I remember there being Mac-specific complexities when this was converted from XUL to HTML, perhaps that is related?

addon/lib/simulator-process.js
```diff
@@ -188,7 +188,7 @@ exports.SimulatorProcess = Class({
     
     // Ignore eventual zombie instances of b2g that are left over
     args.push("-no-remote");
-
+args.push("-jsconsole");
```
mykmelez: Make sure you remove this before merging!

prosthesis/components/SimulatorScreen.manifest
```diff
@@ -1,3 +1,3 @@
 component {c83c02c0-5d43-4e3e-987f-9173b313e880} SimulatorScreen.js
 contract @mozilla.org/simulator-screen;1 {c83c02c0-5d43-4e3e-987f-9173b313e880}
-category JavaScript-global-property screen @mozilla.org/simulator-screen;1
+category profile-after-change SimulatorScreen @mozilla.org/simulator-screen;1
```
mykmelez: Nit: this file is missing a trailing newline.

mykmelez:
The code looks reasonable, but there are still a couple of issues with the experience.

First, one can only rotate the view 90º in one direction, which was fine when only orientable apps were experiencing rotation. But now that all apps experience rotation, and non-orientable apps like Homescreen turn sideways, it seems likely that users will split between wanting to see those sideways apps pointing in one direction vs. the other. So we should enable that (perhaps by making each successive click of the button cause the view to rotate another 90º in the same direction).

Second, if you're in an orientable app (like Marketplace), and you rotate the screen to landscape, and then you return to the Homescreen, the rotation changes back to portrait. I would expect it to stay rotated.

I'd prefer to address these experience issues before landing the change, but in any case this is a big improvement, so I'm ok with landing it and then iterating on the details.

mykmelez:
Looks good, r=myk!

PR #871, opened 2013-12-02
Replace bars xpcom component by a pref (before uplifting)

ochameau:
The xpcom component trick felt too hacky to be uplifted, so I searched for other ways to make various bars (menubar, toolbar, locationbar, ...) appear to be hidden via the API, while our xul menubar to be still visible.
I found that toolkit.defaultChromeFeatures controls bars APIs values but doesn't prevent displaying our xul menu, handy!

I'm commit that first to the addon, before uplifting it to b2g, to ensure that it works fine.


ochameau:
Uplift tracked over here:
https://bugzilla.mozilla.org/show_bug.cgi?id=945235

mykmelez:
Aha! Great fix. :-)

Issue #875, opened 2014-01-08
Syntax error in simulator-process.js: missing ( before formal parameters


panzi:
I guess the problem is that ECMAScript 6 features are used (generator functions) but the JavaScript interpreter runs in an older language mode?

Firefox 26.0 Linux x86_64
Firefox OS 1.3 Simulator 7.0pre6.20140107 (also happens in Firefox OS 1.2 Simulator 6.0pre7.20140106)

Error line:

``` javascript
  /**
   * Start the process and connect the debugger client.
   */
  run: function () Task.spawn((function *() { // SyntaxError is in this line
    // kill before start if already running
    if (this.process != null) {
      yield this.process.kill();
    }
```

Log on the Linux console:

```
console.error: fxos_1_3_simulator: 
  Message: SyntaxError: missing ( before formal parameters
  Stack:
    @resource://fxos_1_3_simulator-at-mozilla-dot-org/fxos_1_3_simulator/lib/simulator-process.js:54
CuddlefishLoader/options<.load@resource://gre/modules/commonjs/sdk/loader/cuddlefish.js:129
@resource://fxos_1_3_simulator-at-mozilla-dot-org/fxos_1_3_simulator/lib/main.js:3
CuddlefishLoader/options<.load@resource://gre/modules/commonjs/sdk/loader/cuddlefish.js:129
run@resource://gre/modules/commonjs/sdk/addon/runner.js:145
startup/<@resource://gre/modules/commonjs/sdk/addon/runner.js:113
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:118
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:118
then@resource://gre/modules/commonjs/sdk/core/promise.js:43
then@resource://gre/modules/commonjs/sdk/core/promise.js:153
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:185
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:118
then@resource://gre/modules/commonjs/sdk/core/promise.js:43
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:185
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:118
resolve@resource://gre/modules/commonjs/sdk/core/promise.js:118
then@resource://gre/modules/commonjs/sdk/core/promise.js:43
then@resource://gre/modules/commonjs/sdk/core/promise.js:153
[...]

PR #876, opened 2014-01-08
Fix FF26 compatibility.

ochameau:
I removed the usage of star function in order to fix Firefox 26 support
https://github.com/mozilla/r2d2b2g/issues/875


ochameau:
@mykmelez could you confirm that's the right way to resolve the Task?

mykmelez:
Yup, that's the right way to do it!

PR #879, opened 2014-01-21
Update adb - deprecate

ochameau:
- Update adb to prevent crashes on maverick
- Display a deprecation message if we are on firefox 26+

### DIFF
--- a/addon/data/content/index.html
+++ b/addon/data/content/index.html
@@ -30,6 +30,11 @@
                 <h5 id="device-status" class="device-dependent">
                     <img src="device.svg" alt="Device"> Device connected.
                 </h5>
+                <div id="deprecation-message" class="deprecated" hidden="true">
+                    This add-on is incompatible with Firefox 26 and greater.
+                    Please use the App Manager with the new Simulator add-ons.<br/>
+                    <a href="https://developer.mozilla.org/docs/Mozilla/Firefox_OS/Using_the_App_Manager">More information</a>
+                </div>
             </header>
 
             <section id="dashboard">
--- a/addon/data/content/css/main.css
+++ b/addon/data/content/css/main.css
@@ -585,6 +585,13 @@ section header {
     pointer-events: all;
 }
 
+#sidebar .deprecated {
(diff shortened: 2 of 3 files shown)


addon/data/content/index.html
```diff
@@ -30,6 +30,11 @@
                 <h5 id="device-status" class="device-dependent">
                     <img src="device.svg" alt="Device"> Device connected.
                 </h5>
+                <div id="deprecation-message" class="deprecated" hidden="true">
+                    This addon has compatibility issues with Firefox 26 and greater.
+                    Please use the App manager with the new Simulator addons.<br/>
```
mykmelez: Nits: addon -> add-on, App manager -> App Manager, addons -> add-ons

Also, I would be more emphatic here and say that this addon is _not compatible_ with Firefox 26 and greater.

mykmelez:
This looks reasonable, but we haven't shipped the 5.0 release yet, and it isn't yet ready for release. So I think we should make these changes to the 4.0 codebase and ship a 4.0.1 with them.

(We should probably also decide that we're not going to ship 5.0, assuming that there are few developers on older versions of Firefox, so it isn't worth completing that work, and we should focus instead on shipping libadb in the ADB Helper addon for the App Manager than in an updated version of the classic addon for older Firefoxes.)

ochameau:
You're right, the (bad) reason I based on 5.0 is that, we only have a 5.0 branch that I can pull request to.
I'll rebase that patch against https://github.com/mozilla/r2d2b2g/releases/tag/4.0
and create a 4.0 branch. Is it the right revision I should target?

mykmelez:
> You're right, the (bad) reason I based on 5.0 is that, we only have a 5.0 branch that I can pull request to.
> I'll rebase that patch against https://github.com/mozilla/r2d2b2g/releases/tag/4.0
> and create a 4.0 branch. Is it the right revision I should target?

Yes, that's the right revision!

ochameau:
Closing in favor of https://github.com/mozilla/r2d2b2g/pull/881

PR #881, opened 2014-01-22
Update adb deprecate

ochameau:


ochameau:
@mykmelez I tweaked the message and rebased against 4.0.

mykmelez:
Looks great, thanks @ochameau! I'll roll a 4.0.1 update with these changes.

